PR #21, opened 2023-11-20
Add error message and self-help instructions for kubernetes permissions

Obliviour:
## Fixes / Features
- Adds instructions on how to add kubernetes permissions
- Adds error message with cli instructions for adding kubernetes permissions

## Testing / Documentation
Viewed error message:

```
[XPK] This likely means you're missing Kubernetes Permissions, you can validate this by checking if the error references permission problems such as `requires one of ["container.*"] permission(s)`. Users fix this permission issue by applying the role  `Kubernetes Engine Admin` in https://console.cloud.google.com/iam-admin/ Run the following command to add this permission to your current user:

 `PROJECT_ID=my-project-id; CURRENT_GKE_USER=$(gcloud config get account); gcloud projects add-iam-policy-binding $PROJECT_ID --member user:$CURRENT_GKE_USER --role=roles/container.admin`

If you see other missing permissions, please add the correct role to your user. Then remember to `gcloud auth login` and make sure you have the right permissions by running:

 `PROJECT_ID=my-project-id; CURRENT_GKE_USER=$(gcloud config get account); gcloud projects get-iam-policy $PROJECT_ID --filter="bindings.members:$CURRENT_GKE_USER" --flatten="bindings[].members"`.
```

- [ y] Tests pass

### DIFF
--- a/xpk.py
+++ b/xpk.py
@@ -1112,7 +1112,7 @@ def set_jobset_on_cluster(args) -> int:
       ' `PROJECT_ID=my-project-id; CURRENT_GKE_USER=$(gcloud config get account);'
       ' gcloud projects get-iam-policy $PROJECT_ID'
       ' --filter="bindings.members:$CURRENT_GKE_USER"'
-      ' --flatten="bindings[].members"`\n.'
+      ' --flatten="bindings[].members"`.\n'
     )
     return 1
   return 0
--- a/xpk.py
+++ b/xpk.py
@@ -1096,6 +1096,24 @@ def set_jobset_on_cluster(args) -> int:
         'jobset command on server side returned with ERROR returncode'
         f' {return_code}.\n'
     )
+    xpk_print(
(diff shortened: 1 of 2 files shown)


xpk.py
```diff
@@ -1020,6 +1020,24 @@ def set_jobset_on_cluster(args) -> int:
         'jobset command on server side returned with ERROR returncode'
         f' {return_code}.\n'
     )
+    xpk_print(
+      'This likely means you\'re missing Kubernetes Permissions, you can'
```
rwitten, line 1024: Seems super duplicative. Sure you don't want to link to the README and have a deeplink in the README?

rwitten (approved):
This is fine but seems unmaintainable, might be worth unifying.

xpk.py
Obliviour: Agreed and added a link to the url weblink to the readme in the error message.

PR #25, opened 2023-11-21
Add support for ondemand/spot/reservation for cluster creation

Obliviour:
## Fixes / Features
- Adds direct xpk arguments for on-demand spot and reservations
-

## Testing / Documentation
Tested that we can create clusters with reservations
Tested that multiple cluster capacity flags will fail with the right error
Tested that not setting these arguments will fail and list the reservations available.
- [ y ] Tests pass

### DIFF
--- a/xpk.py
+++ b/xpk.py
@@ -775,6 +775,76 @@ def get_all_nodepools_programmatic(args) -> tuple[list[str], int]:
   return all_nodepools, 0
 
 
+def print_reservations(args) -> int:
+  """Print the reservations in the project.
+
+  Args:
+    args: user provided arguments for running the command.
+
+  Returns:
+    0 if successful and 1 otherwise.
+  """
+  command = (
+        f'gcloud compute reservations list --project={args.project}'
+    )
+  return_code = (
+      run_command_with_updates(
+          command, 'Get all reservations in the project', args)
+  )
+  if return_code != 0:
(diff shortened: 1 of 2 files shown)


xpk.py
```diff
@@ -775,6 +775,76 @@ def get_all_nodepools_programmatic(args) -> tuple[list[str], int]:
...
+  Returns:
+    Tuple with string with the capacity argument to use and
+    int of 0 if successful and 1 otherwise.
+  """
+  capacity_args = ""
+  num_types = 0
+  return_code = 0
+
+  # Determine the capacity argument.
+  if args.on_demand:
```
rwitten, line 815: I don't love having three separate flags that conflict and aren't linked. It seems a lot of elegant to have a `--capacity-pool` that then takes the three arguments. (That helps users discover how to switch and expresses the idea that you can't have both a reservation be spot.) However, I understand we're constrained by upstream.

rwitten (approved):
Conflicted by one detail but I think we're overconstrained.

xpk.py
Obliviour: Discussed offline, since combining the arguments would fight against GKE/GCE API we won't change this behavior. The other option here was 

```
we contain all three options in one arg and then also require a reservation id (probably would have to be an additional argument like --capacity-pool=reservation --reservation-id=apple-pie)
```

PR #36, opened 2023-11-30
Configuration to deploy a sidecar container to display stack traces

SurbhiJainUSC:
## Fixes / Features
- Add configuration to deploy a sidecar container in a workload that will read the stack traces collected in `/tmp/debugging` directory on kubernetes pod

There are two main things to periodically collect stack traces and view it on Google Cloud:
1. Adding stack traces configuration in your training script so that program could periodically dump traces in a temp file inside `/tmp/debugging` directory on TPU. MaxText has this configuration already added: https://github.com/google/maxtext/pull/53. Non-MaxText users should follow this PR to enable stack trace collection for their script.
2. Reading the traces from temp directory and uploading it to Google Cloud. For TPU workloads running on GCE, we have a runtime agent running on VMs that automatically upload the traces from temp directory to Google Cloud. However, for GKE workloads there is no such kind of agent. As a workaround, we can deploy a sidecar container along with our main container (training workload) to read the traces from temp file and upload it to Google Cloud to be viewable for a separate container (to make sure logs for main container are clean and separated from stack traces). This PR is basically adding the configuration to deploy the sidecar container for GKE workloads.

## Testing / Documentation
Tested with MaxText.

- [ y ] Tests pass

### DIFF
--- a/README.md
+++ b/README.md
@@ -321,3 +321,46 @@ gcloud compute machine-types list --zones=$ZONE_LIST
 # Adjust default cpu machine type.
 python3 xpk.py cluster create --cluster-cpu-machine-type=CPU_TYPE ...
 ```
+
+## Permission Issues: `requires one of ["permission_name"] permission(s)`.
+
+1) Determine the role needed based on the permission error:
+
+    ```shell
+    # For example: `requires one of ["container.*"] permission(s)`
+    # Add [Kubernetes Engine Admin](https://cloud.google.com/iam/docs/understanding-roles#kubernetes-engine-roles) to your user.
+    ```
+
+2) Add the role to the user in your project.
+
+    Go to [iam-admin](https://console.cloud.google.com/iam-admin/) or use gcloud cli:
+    ```shell
+    PROJECT_ID=my-project-id
+    CURRENT_GKE_USER=$(gcloud config get account)
+    ROLE=roles/container.admin  # container.admin is the role needed for Kubernetes Engine Admin
+    gcloud projects add-iam-policy-binding $PROJECT_ID --member user:$CURRENT_GKE_USER --role=$ROLE
+    ```
+
+3) Check the permissions are correct for the users.
+
+    Go to [iam-admin](https://console.cloud.google.com/iam-admin/) or use gcloud cli:
+
+    ```shell
+    PROJECT_ID=my-project-id
+    CURRENT_GKE_USER=$(gcloud config get account)
+    gcloud projects get-iam-policy $PROJECT_ID --filter="bindings.members:$CURRENT_GKE_USER" --flatten="bindings[].members"
+    ```
+
+4) Confirm you have logged in locally with the correct user.
+
+    ```shell
+    gcloud auth login
+    ```
+
+
+
+### Roles needed based on permission errors:
+
+* `requires one of ["container.*"] permission(s)`
+
+  Add [Kubernetes Engine Admin](https://cloud.google.com/iam/docs/understanding-roles#kubernetes-engine-roles) to your user.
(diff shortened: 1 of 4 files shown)


xpk.py
```diff
@@ -2215,6 +2286,16 @@ def directory_path_type(value):
     ),
 )
 
+workload_create_parser_optional_arguments.add_argument(
```
Obliviour, line 2289: if you use `action='store_false'` then it will be easier to verify that the user sets this argument correctly. The user would then only need to say `--deploy-stacktrack-sidecar` instead of setting a string which they could accidently type wrong.

https://docs.python.org/3/library/argparse.html#action

README.md
```diff
@@ -185,6 +185,38 @@ all zones.
   xpk-test --tpu-type=v5litepod-16 --priority=medium
   ```
 
+### Workload Debugging
+#### Collect Stack Traces
+`cloud-tpu-diagnostics` PyPI package can be used to generate stack traces for workloads running in GKE. This package dumps the Python traces when a fault such as segmentation fault, floating-point exception, or illegal operation exception occurs in the program. Additionally, it will also periodically collect stack traces to help you debug situations when the program is unresponsive. You must make the following changes in the docker image running in a Kubernetes main container.
```
Obliviour, line 190: add a link to the PyPI package or github package readme for more details?

xpk.py
```diff
@@ -1517,6 +1502,85 @@ def setup_docker_image(args) -> tuple[int, str]:
...
+      yaml for main and sidecar container
+  """
+  yaml = """- name: stacktrace-explorer
+                image: busybox:1.28
+                args: [/bin/sh, -c, "while [ ! -d /tmp/debugging ]; do sleep 60; done; while [ ! -e /tmp/debugging/* ]; do sleep 60; done; tail -n+1 -f /tmp/debugging/*"]
+                volumeMounts:
+                - name: tpu-stack-trace
+                  readOnly: true
+                  mountPath: /tmp/debugging
+              - name: {args.docker_name}
```
Obliviour, line 1524: We want to avoid duplicating the yaml file (main container aspect) in two functions if possible, otherwise it can be harder to maintain and refactor changes to the main container in the future.

In your get_main_and_sidecar_container function, I recommend it directly calls the get_main_container function instead of replicating the main container details. LMK if you see any issues with this!

README.md
Obliviour: "You must make the following changes in the docker image running in a Kubernetes main container."

I am slightly confused by the wording here, could you help me understand what the Kubernetes main container is and how the python code below relates? Thank you!

Obliviour (commented):
Overall Looks Good! Have some structural recommendations to help to maintain and make it easier to use!

README.md
SurbhiJainUSC: This change must be done in the main training workload that we are running on Kubernetes.

xpk.py
SurbhiJainUSC: That's nice, I will make this change.

Obliviour (approved):

xpk.py
```diff
@@ -2594,6 +2649,15 @@ def directory_path_type(value):
     ),
 )
 
+workload_create_parser_optional_arguments.add_argument(
+    '--deploy-stacktrace-sidecar',
+    action='store_true',
+    help=(
+        'Add this argument to deploy a sidecar container that will '
+        'read the stack traces collected in /tmp/debugging directory.'
```
bvandermoon, line 2657: Should this also say "and forward them to Cloud Logging"?

SurbhiJainUSC: Sorry, I missed your comment before merging. Will update in next PR.

PR #37, opened 2023-12-04
Verify reservation exists and print describe details if it does

Obliviour:
## Fixes / Features
- Verifies that the reservation exists and prints out details on the utilization / size of the reservation. 
- AI to check size of the reservation within xpk.

## Testing / Documentation
Tested bad reservation or zone and the command failed. Tested correct reservation and the command worked.

- [ y ] Tests pass


rwitten (approved):
neat change. would be much sweeter if we could parse the reservation in json, not sure if that is possible.

xpk.py
```diff
@@ -820,6 +844,9 @@ def get_capacity_arguments(args) -> tuple[str, int]:
     capacity_args = ""
     num_types+=1
   if args.reservation:
+    return_code = verify_reservation_exists(args)
+    if return_code > 0:
```
rwitten, line 848: should we warn if we're using on-demand?

rwitten (approved):

xpk.py
Obliviour: Hmm I think we don't need to since the user would have to specify `--on-demand` in their command.

rwitten: Fair.

rwitten (approved):

PR #39, opened 2023-12-07
Outlier

SurbhiJainUSC:
## Fixes / Features
- This PR outputs the link of the monitoring dashboard deployed in Google Cloud project using Terraform resource in https://github.com/google/cloud-tpu-monitoring-debugging. The monitoring dashboard will display the statistics and outlier mode of GKE metrics. The dashboard can be filtered for a particular GKE Custer, job name or a GKE Pod.

## Testing / Documentation
Tested for internal projects.

- [ y ] Tests pass
    - No change required in README

### DIFF
--- a/xpk.py
+++ b/xpk.py
@@ -1838,6 +1838,42 @@ def setup_docker_image(args) -> tuple[int, str]:
 
   return 0, docker_image
 
+def get_gke_outlier_dashboard(args):
+  outlier_dashboard_filter = "displayName:'GKE - TPU Monitoring Dashboard'"
+  command = (
+      'gcloud monitoring dashboards list'
+      f' --project={args.project} --filter="{outlier_dashboard_filter}" --format="value(name)"'
+  )
+  
+  return_code, return_value = run_command_for_value(command, 'GKE Dashboard List', args)
+  
+  if return_code != 0:
+    xpk_print(f'GKE Dashboard List request returned ERROR {return_code}')
+    return None
+  
+  if not return_value:
+    xpk_print(
+        f'No dashboard with {outlier_dashboard_filter} found in the'
+        f' project:{args.project}.'
+    )
+    xpk_print(
+        'Follow'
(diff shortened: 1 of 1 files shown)


xpk.py
```diff
@@ -1889,11 +1932,22 @@ def workload_create(args) -> int:
...
       'Follow your workload here:'
       # pylint: disable=line-too-long
       f' https://console.cloud.google.com/kubernetes/service/{zone_to_region(args.zone)}/{args.cluster}/default/{args.workload}/details?project={args.project}'
   )
+
+  if outlier_dashboard_id is not None:
+    xpk_print(
+        'Check statistics and outlier mode of GKE metrics for your workload here:'
+        # pylint: disable=line-too-long
+        f'http://console.cloud.google.com/monitoring/dashboards/builder/{outlier_dashboard_id}?project={args.project}'
```
bvandermoon, line 1948: Is it possible to attach the clusterName, jobName, and podName filters to the link here too?

xpk.py
```diff
@@ -1838,6 +1838,49 @@ def setup_docker_image(args) -> tuple[int, str]:
...
+        f' project:{args.project}.'
+    )
+    xpk_print(
+        'Follow https://github.com/google/cloud-tpu-monitoring-debugging'
+        ' to deploy monitoring dashboard to view statistics and outlier mode of GKE metrics.'
+    )
+
+  dashboards = return_value.strip().split('\n')
+  if len(dashboards) > 1:
+    xpk_print(f'Multiple dashboards with same {outlier_dashboard_filter} exist in the project:{args.project}.')
```
Obliviour, line 1876: Interesting edge case, how would someone fix this?

bvandermoon: Would Cloud Monitoring actually let this happen? Or would they block it at creation time?

xpk.py
SurbhiJainUSC: Yeah, great idea. I can add the cluster name to the link. However, it is not ideal to add a pod name filter as there can be multiple pods in a cluster.

bvandermoon (approved):

PR #48, opened 2023-12-12
Add debugging dashboard url to workload creation output

SurbhiJainUSC:
## Fixes / Features
- This PR outputs the link of the debugging dashboard deployed in Google Cloud project using Terraform resource in https://github.com/google/cloud-tpu-monitoring-debugging. The debugging dashboard will display the stack traces collected for the workload using `cloud-tpu-diagnostics` package.
- Refactor the code for both outlier and debugging dashboard

## Testing / Documentation
Testing details.

- [ y ] Tests pass

### DIFF
--- a/xpk.py
+++ b/xpk.py
@@ -1895,7 +1895,7 @@ def get_gke_dashboard(args, dashboard_filter):
 
   Returns:
     str:
-      identifier of dashbord if deployed in project,
+      identifier of dashboard if deployed in project,
       None otherwise.
   """
   command = (
@@ -1906,7 +1906,10 @@ def get_gke_dashboard(args, dashboard_filter):
   return_code, return_value = run_command_for_value(command, 'GKE Dashboard List', args)
 
   if return_code != 0:
-    xpk_print(f'GKE Dashboard List request returned ERROR {return_code}')
+    xpk_print(f'GKE Dashboard List request returned ERROR {return_code}. '
+              'If there is a permissions error, please check '
+              'https://github.com/google/xpk/blob/main/README.md#roles-needed-based-on-permission-errors '
+              'for possible solutions.')
     return None
 
   if not return_value:
@@ -1936,7 +1939,7 @@ def get_gke_outlier_dashboard(args):
 
   Returns:
     str:
-      identifier of outlier dashbord if deployed in project,
+      identifier of outlier dashboard if deployed in project,
       None otherwise.
   """
   outlier_dashboard_filter = "displayName:'GKE - TPU Monitoring Dashboard'"
@@ -1961,7 +1964,7 @@ def get_gke_debugging_dashboard(args):
 
   Returns:
     str:
-      identifier of debugging dashbord if deployed in project,
+      identifier of debugging dashboard if deployed in project,
       None otherwise.
   """
   debugging_dashboard_filter = "displayName:'GKE - TPU Logging Dashboard'"
--- a/xpk.py
+++ b/xpk.py
@@ -1887,6 +1887,47 @@ def get_main_container(args, system, docker_image, command) -> str:
                    docker_image=docker_image,
                    command=command)
 
+def get_gke_dashboard(args, dashboard_filter):
+  """Get the identifier of GKE dashboard deployed in the project.
+
+  Args:
+    args: user provided arguments for running the command.
+
+  Returns:
+    str:
+      identifier of dashbord if deployed in project,
+      None otherwise.
+  """
+  command = (
+      'gcloud monitoring dashboards list'
+      f' --project={args.project} --filter="{dashboard_filter}" --format="value(name)" --verbosity=error'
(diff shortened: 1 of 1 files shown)


xpk.py
```diff
@@ -1887,21 +1887,20 @@ def get_main_container(args, system, docker_image, command) -> str:
...
       None otherwise.
   """
-  outlier_dashboard_filter = "displayName:'GKE - TPU Monitoring Dashboard'"
   command = (
       'gcloud monitoring dashboards list'
-      f' --project={args.project} --filter="{outlier_dashboard_filter}" --format="value(name)" --verbosity=error'
+      f' --project={args.project} --filter="{dashboard_filter}" --format="value(name)" --verbosity=error'
   )
 
   return_code, return_value = run_command_for_value(command, 'GKE Dashboard List', args)
```
Obliviour, line 1906: Sorry for an aside, could you also add to the below error message:

```
    xpk_print(f'GKE Dashboard List request returned ERROR {return_code}')

```

to be

```
    xpk_print(f'GKE Dashboard List request returned ERROR {return_code}. If you see a permissions error, please see xpk readme for possible solutions. ')

```

xpk.py
```diff
@@ -1887,21 +1887,20 @@ def get_main_container(args, system, docker_image, command) -> str:
...
+def get_gke_dashboard(args, dashboard_filter):
+  """Get the identifier of GKE dashboard deployed in the project.
 
   Args:
     args: user provided arguments for running the command.
 
   Returns:
     str:
-      identifier of outlier dashbord if deployed in project,
+      identifier of dashbord if deployed in project,
```
Obliviour, line 1898: nit dashbord -> dashboard

Obliviour: might be misspelled in the other function descriptions as well

xpk.py
SurbhiJainUSC: Yeah, that's a great idea.

xpk.py
```diff
@@ -2008,6 +2059,14 @@ def workload_create(args) -> int:
         f' To view the metric data for your workload, select {args.workload} from the JobName filter on the dashboard.'
     )
 
+  if debugging_dashboard_id is not None:
```
bvandermoon, line 2062: Should this be behind the `--deploy-stacktrace-sidecar` flag as well?

SurbhiJainUSC: It won't be needed, because if the flag is not set then `debugging_dashboard_id` will be `None`. In that case, we won't see any logs.

bvandermoon: I might be misunderstanding - but isn't debugging_dashboard_id tied to whether the user has separately run the terraform script? If so, wouldn't it be possible for users to have that setup in their project, but to not have stack traces available because they weren't forwarded with the `--deploy-stacktrace-sidecar` flag?

SurbhiJainUSC: For GKE workloads, debugging dashboard won't display any stack traces for the workload unless user opt to forward the logs using `deploy-stacktrace-sidecar`. So, it doesn't make any sense to output the link to debugging dashboard when this flag is not enabled because there won't be anything to check.

bvandermoon: Yes, 100% agreed on not showing the link if the flag isn't enabled. I missed seeing where debugging_dashboard_id was set to None in the code. I see it now before making the call to get_gke_debugging_dashboard. Thanks Surbhi

bvandermoon (approved):

xpk.py
```diff
@@ -1933,6 +1931,56 @@ def get_gke_outlier_dashboard(args):
...
+  Args:
+    args: user provided arguments for running the command.
+
+  Returns:
+    str:
+      identifier of outlier dashboard if deployed in project,
+      None otherwise.
+  """
+  outlier_dashboard_filter = "displayName:'GKE - TPU Monitoring Dashboard'"
+  dashboard_id = get_gke_dashboard(args, outlier_dashboard_filter)
```
Obliviour, line 1946: Overall looks great!

I see that we check the dashboard_id in two ways before erroring:
1) if it is None which behind the scenes means an error in the command
2) If not dashboard_id which means the dashboard id is empty and there are no dashboards available.

I think we can make this easier to read by having get_gke_dashboard return a tuple with (return_error, return_val) syntax? what do you think about this?

xpk.py
```diff
@@ -1887,51 +1887,113 @@ def get_main_container(args, system, docker_image, command) -> str:
...
+  return True, None
+
+def get_gke_outlier_dashboard(args):
+  """Get the identifier of GKE outlier dashboard deployed in the project.
+
+  Args:
+    args: user provided arguments for running the command.
+
+  Returns:
+    bool:
```
Obliviour, line 1945: I think `get_gke_outlier_dashboard` only returns a str

Obliviour (approved):

PR #50, opened 2023-12-20
Early failure of workload creation based on cluster resources

SurbhiJainUSC:
## Fixes / Features
- Workload create verifies that the cluster resources exist for the tpu type to be created otherwise fail early with appropriate error message.

## Testing / Documentation
Testing details.

- [ y ] Tests pass

### DIFF
--- a/xpk.py
+++ b/xpk.py
@@ -229,6 +229,14 @@ spec:
         command: [ "sleep", "inf" ]
 """
 
+cluster_configmap_yaml = """kind: ConfigMap
+apiVersion: v1 
+metadata:
+  name: {args.cluster}-configmap 
+data:
+  {data}
+"""
+
 @dataclass
 class SystemCharacteristics:
   topology: str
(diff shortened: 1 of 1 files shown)


xpk.py
```diff
@@ -1763,6 +1829,43 @@ def check_if_workload_exists(args) -> bool:
...
+    args: user provided arguments for running the command.
+    system: system characteristics
+
+  Returns:
+    returns true if workload can schedule, otherwise returns false.
+  """
+  cluster_config_map = get_cluster_configmap(args)
+
+  # Prevents workload creation failure for existing clusters with no ConfigMap
+  if cluster_config_map is None:
```
Obliviour, line 1845: This is smart to allow it to be backward compatible with clusters created without this change. +1.

xpk.py
```diff
@@ -229,6 +229,14 @@
         command: [ "sleep", "inf" ]
 """
 
+cluster_configmap_yaml = """kind: ConfigMap
+apiVersion: v1
```
Obliviour, line 233: The config map is great. I am thinking we will use the config map to store additional data. How do you recommend we do this? Like for example, we want to store a XPK version that the cluster was created with.

One way I think we could do this is we set a custom name for each config map?

xpk.py
```diff
@@ -999,6 +1007,59 @@ def run_gke_cluster_create_command(args) -> int:
...
+  """Run the Create GKE Cluster ConfigMap request.
+
+  Args:
+    args: user provided arguments for running the command.
+
+  Returns:
+    0 if successful and 1 otherwise.
+  """
+  system = UserFacingNameToSystemCharacteristics[args.tpu_type]
+  data = f'{args.tpu_type}: "{int(args.num_slices) * system.vms_per_slice}"'
```
Obliviour, line 1020: Just to note that @michelle-yooh is adding gpu support which will need to modify this a bit. https://github.com/google/xpk/pull/49. So we just have to be aware depending on the order that things get merged!

xpk.py
SurbhiJainUSC: I was looking into some nesting structure but I don't think ConfigMap support that. We would have to create separate config map to store metadata like XPK version.

xpk.py
```diff
@@ -1763,6 +1829,43 @@ def check_if_workload_exists(args) -> bool:
   return False
 
 
+def check_if_workload_can_schedule(args, system):
+  """Check if workload can schedule based on the cluster resources (tpu_type and maximu VM in cluster).
+
```
Obliviour, line 1834: Nit maximum

xpk.py
SurbhiJainUSC: Thanks for the heads up.

xpk.py
Obliviour: Thinking a bit more, I think we can change the config map name here to align with its purpose (to store the resource types of the cluster). This will also, allow us to create additional config maps easier

xpk.py
```diff
@@ -1763,6 +1829,43 @@ def check_if_workload_exists(args) -> bool:
...
+      'XPK will not create this workload.'
+    )
+    return False
+
+  max_vm_in_cluster = cluster_config_map[args.tpu_type]
+  vm_required_by_workload = int(args.num_slices) * system.vms_per_slice
+  if vm_required_by_workload > max_vm_in_cluster:
+    xpk_print(
+        f'{args.workload} is requesting {vm_required_by_workload} VMs of {args.tpu_type} '
+        f'but the cluster only contains {max_vm_in_cluster} VMs of {args.tpu_type}. '
```
Obliviour, line 1861: Is it possible to add the number of slices of the tpu type being requested as well? I think it is easier to understand the workload based on the same args the user provides which is in num-slices and tpu type. The vm count is also great since kueue using vm count so having both will be super clear!

xpk.py
SurbhiJainUSC: yes, will change that.

Obliviour (commented):
This is a great feature and looks really good!! Thank you for making this change!

xpk.py
```diff
@@ -229,6 +229,14 @@
         command: [ "sleep", "inf" ]
 """
 
+cluster_configmap_yaml = """kind: ConfigMap
+apiVersion: v1 
+metadata:
+  name: {args.cluster}-resources-configmap
```
Obliviour, line 235: Not needed now but when we add the cluster version in a new config map, we can change this line to
`name: {args.cluster}-{config_map_name}-configmap`, then we can reuse this config map. Up to you if you want to add that now or later!

xpk.py
```diff
@@ -999,6 +1007,60 @@ def run_gke_cluster_create_command(args) -> int:
...
+
+  Returns:
+    0 if successful and 1 otherwise.
+  """
+  # TODO: Update when GPU support is enabled
+  data = f'{args.tpu_type}: "{int(args.num_slices) * system.vms_per_slice}"'
+  yml_string = cluster_configmap_yaml.format(args=args,
+                                           data=data)
+  tmp = write_temporary_file(yml_string)
+  command = f'kubectl apply -f {str(tmp.file.name)}'
```
Obliviour, line 1025: What happens if the cluster has a config map already with the same name? Can it apply it or does it error? We would want to be to update config maps on new calls of cluster create.

Obliviour (approved):
Two thoughts and looks great!

xpk.py
SurbhiJainUSC: If the ConfigMap already exist with the same map, it will update the data in the map.

PR #54, opened 2024-01-10
Fix device-type=none bug when using tpu-type

michelle-yooh:
## Fixes / Features
- A bug where the commands are requesting `None` when using `--tpu-type` instead of `--device-type`

## Testing / Documentation
Testing details.
Created a cluster with 2 slices of v4-8 and a workload using `--tpu-type`

- [ y ] Tests pass

### DIFF
--- a/xpk.py
+++ b/xpk.py
@@ -1069,7 +1069,8 @@ def create_cluster_configmap(args, system):
   Returns:
     0 if successful and 1 otherwise.
   """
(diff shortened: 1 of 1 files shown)


Obliviour (approved):
LGTM, I think xpk code needs a refactor so that we can have a good place to only set the device_type once instead of multiple times but thats outside the scope of this PR. Thanks Michelle for the quick fix!

PR #56, opened 2024-01-16
Adding CPU support in XPK

RoshaniN:
## Features
- Adding single slice CPU support for XPK, works with n2-standard-32 machines.

Major changes:
- allow for CPU device-type in the format <"n2-standard-32">-<# of n2-standard-32 VMs>
- to avoid confusion with cluster's default pool CPU type, deprecated the argument "--cluster-cpu-machine-type" in favor of  "--default-pool-cpu-machine-type" and added error check.
- added nodeAffinity to workload create spec, which is useful for CPUs to ensure that workload pods don't get scheduled onto default pool machines.
- introduced CPU cluster env to inject JAX variables into GKE 

## Testing / Documentation
- Tested cluster creation, CPU nodepool creation with args and env properly set.
- Ran Maxtext workloads on CPUs using these changes.
- Documented changes in README as well.
- Verified that TPU training (with Maxtext workloads) run as expected.

- [ y ] Tests pass

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,8 +18,8 @@
 
 xpk (Accelerated Processing Kit, pronounced x-p-k,) is a software tool to help
 Cloud developers to orchestrate training jobs on accelerators such as TPUs and
-GPUs on GKE. xpk handles the "multihost pods" of TPUs and GPUs (HGX H100) as
-first class citizens.
+GPUs on GKE. xpk handles the "multihost pods" of TPUs, GPUs (HGX H100) and CPUs
+(n2-standard-32) as first class citizens.
 
 xpk decouples provisioning capacity from running jobs. There are two structures:
 clusters (provisioned VMs) and workloads (training jobs). Clusters represent the
@@ -44,6 +44,9 @@ xpk supports the following TPU types:
 and the following GPU types:
 * a100
 * h100
+
+and the following CPU types:
+* n2-standard-32
 
 # Installation
 To install xpk, run the following command:
@@ -314,6 +317,36 @@ In order to use XPK for GPU, you can do so by using `device-type` flag.
     --command="echo hello world"
     ```
 
+## CPU usage
+
+In order to use XPK for CPU, you can do so by using `device-type` flag.
+
+*   Cluster Create (provision on-demand capacity):
+
+    ```shell
+    # Run cluster create with on demand capacity.
+    python3 xpk.py cluster create \
+    --cluster xpk-test \
+    --device-type=n2-standard-32-256 \
+    --num-slices=1 \
+    --default-pool-cpu-machine-type=n2-standard-32 \
+    --on-demand
+    ```
+    Note that `device-type` for CPUs is of the format <cpu-machine-type>-<number of VMs>, thus in the above example, user requests for 256 VMs of type n2-standard-32.
+    Currently workloads using < 1000 VMs are supported.
+
+*   Run a workload:
+
+    ```shell
+    # Submit a workload
+    python3 xpk.py workload create \
+    --cluster xpk-test \
+    --num-slices=1 \
+    --device-type=n2-standard-32-256 \
+    --workload xpk-test-workload \
+    --command="echo hello world"
+    ```
+
 
 # How to add docker images to a xpk workload
 
@@ -403,7 +436,7 @@ Please select a CPU type that exists in all zones in the region.
 # Find CPU Types supported in zones.
 gcloud compute machine-types list --zones=$ZONE_LIST
 # Adjust default cpu machine type.
-python3 xpk.py cluster create --cluster-cpu-machine-type=CPU_TYPE ...
+python3 xpk.py cluster create --default-pool-cpu-machine-type=CPU_TYPE ...
 ```
 
 ## Permission Issues: `requires one of ["permission_name"] permission(s)`.
--- a/xpk.py
+++ b/xpk.py
@@ -91,6 +91,7 @@ spec:
             spec:
               schedulerName: {args.scheduler}
               restartPolicy: Never
+              {affinity}
               nodeSelector:
                 {accelerator_label}
                 {machine_label}
@@ -100,6 +101,7 @@ spec:
               terminationGracePeriodSeconds: {args.termination_grace_period_seconds}
               containers:
               {container}
+                {env}
                 volumeMounts:
                 - mountPath: /dev/shm
                   name: dshm-2
@@ -247,7 +249,8 @@ data:
 
 AcceleratorType = {
   'TPU': 1,
-  'GPU': 2
+  'GPU': 2,
+  'CPU': 3
 }
 
 @dataclass
@@ -264,6 +267,10 @@ AcceleratorTypeToAcceleratorCharacteristics = {
      # GPU
       AcceleratorType['GPU']: AcceleratorCharacteristics(
       'nvidia.com/gpu', 'cloud.google.com/gke-accelerator', 'cloud.google.com/gce-machine-type'
+      ),
+     # CPU
+      AcceleratorType['CPU']: AcceleratorCharacteristics(
+      'cpu', '', 'cloud.google.com/gke-nodepool' 
       )
 }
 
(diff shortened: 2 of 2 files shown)


xpk.py
```diff
@@ -642,6 +656,25 @@ class SystemCharacteristics:
...
+
+    # CPU system characteristics
+    # n2-standard-32-$VMs
+    'n2-standard-32-1': SystemCharacteristics(
+      'N/A', 1,'N/A', 'n2-standard-32', 1, AcceleratorType['CPU'], 'n2-standard-32'
+    ),
+    'n2-standard-32-2': SystemCharacteristics(
+      'N/A', 2,'N/A', 'n2-standard-32', 1, AcceleratorType['CPU'], 'n2-standard-32-2'
+    ),
+    'n2-standard-32-4': SystemCharacteristics(
```
RoshaniN, line 668: Hi @Obliviour 
I have used device type similar to those specified in the GPUs section "n2-standard-32-4" is "machine_type-$VMs" . Does that make sense or would you have any inputs here?

Obliviour: I think this makes sense. The -$VM syntax is consistent with our other usage which is great.

Let's rename -$VM to -$vCPUs to align with the syntax in cloud.google.com:

https://cloud.google.com/compute/docs/general-purpose-machines

xpk.py
```diff
@@ -91,6 +91,15 @@
             spec:
               schedulerName: {args.scheduler}
               restartPolicy: Never
+              affinity:
+                nodeAffinity:
+                  requiredDuringSchedulingIgnoredDuringExecution:
+                    nodeSelectorTerms:
+                    - matchExpressions:
+                      - key: 'cloud.google.com/gke-nodepool'
```
danielvegamyhre, line 99: Remove quotes

xpk.py
```diff
@@ -91,6 +91,15 @@
...
+                        operator: NotIn
+                        value:
+                          default-pool
```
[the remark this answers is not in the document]
RoshaniN: Got it, thanks.

xpk.py
RoshaniN: There are 32vCPUs in an n2-standard-32 but it is considered 1 VM, hence the annotation. 

I am using "n2-standard-32-4" to denote 4VMs of configuration n2-standard-32. Using vCPUs instead of VMs here would be wrong actually as there are 32 vCPUs per machine and there would be 4 such machines.

Would you prefer 4-n2-standard-32 to say 4 VMs of n2-standard-32?

README.md
xuefgu, line 1: Shall we also document how to emulate multiple slices using a single CPU nodepool?

xpk.py
```diff
@@ -2629,14 +2752,24 @@ def directory_path_type(value):
...
     type=str,
     default='e2-standard-16',
     help=(
-      'Set the machine tpu within the default cpu node pool. For'
+      'Set the machine type within the default cpu node pool. For'
       ' regional clusters, all zones must support the machine type.'
     )
 )
+cluster_create_optional_arguments.add_argument(
+  '--cluster-cpu-machine-type',
```
xuefgu, line 2764: Do we really want to deprecate this field?

RoshaniN: Yes, I hope to minimize confusion by having better args names, especially since I am introducing CPUs. This field is deprecated, to be replaced by "--default-pool-cpu-machine-type"

README.md
RoshaniN: Slices are emulated using Maxtext parallelism types, hence I think it would belong in Maxtext documentation.

xpk.py
Obliviour: I agree it can be confusing for the --device-type=CPU_TYPE user. 

Since we added deprecated messages, the impact would be that any current user would have to change their argument. How about we set this to a 'warning' level for now so the old argument still works. We want to warn users still with a print statement during xpk execution, and also in the help argument. In the future we can delete it completely after users are more aware that it is being deprecated.

xpk.py
```diff
@@ -91,6 +91,15 @@
...
               restartPolicy: Never
+              affinity:
+                nodeAffinity:
+                  requiredDuringSchedulingIgnoredDuringExecution:
+                    nodeSelectorTerms:
+                    - matchExpressions:
+                      - key: cloud.google.com/gke-nodepool
+                        operator: NotIn
+                        values:
+                        - default-pool
```
Obliviour, line 102: What do you think of setting this only for CPUs?

Also would appreciate a comment saying that this "The scheduler can't schedule the Pod in the default-pool.".

xpk.py
```diff
@@ -1036,6 +1090,12 @@ def run_gke_cluster_create_command(args) -> int:
   Returns:
     0 if successful and 1 otherwise.
   """
+  # cluster_cpu_machine_type is being deprecated!
+  if args.cluster_cpu_machine_type != '':
+    xpk_print('Note that cluster-cpu-machine-type is deprecated,',
+              'please use default-pool-cpu-machine-type instead!',
```
Obliviour, line 1096: I like the message you used in the argument description:

Can we rephrase to:
```
'Note that `--cluster-cpu-machine-type` is deprecated,',
' Please use `--default-pool-cpu-machine-type={args.cluster_cpu_machine_type}` instead,'
' to denote the machine type of the default cpu node pool. Set'
' the machine type of other cpu nodepools using `--device-type`.'
```

xpk.py
```diff
@@ -1316,11 +1376,13 @@ def run_gke_node_pool_create_command(args, system) -> int:
         f' {capacity_args}'
         ' --scopes=storage-full,gke-default'
         ' --enable-gvnic --max-pods-per-node 15'
-        f' {args.custom_tpu_nodepool_arguments}'
     )
     if system.accelerator_type == AcceleratorType['TPU']:
+      command += (' --placement-type=COMPACT ')
       command += (f' --tpu-topology={system.topology}')
+      command += (f' {args.custom_tpu_nodepool_arguments}')
```
Obliviour, line 1383: Nice catch on `custom_tpu_nodepool_arguments`.

We will need a custom_device_nodepool_arguments to support the same feature across all device types. I will create a bug for this.

Obliviour (commented):
Looks great! Very clean set of changes. Excited to see multiple node pool support for CPUs

xpk.py
RoshaniN: Sure, sounds good, will update the print message to have a warning about deprecation.

README.md
```diff
@@ -19,7 +19,7 @@
 xpk (Accelerated Processing Kit, pronounced x-p-k,) is a software tool to help
 Cloud developers to orchestrate training jobs on accelerators such as TPUs and
 GPUs on GKE. xpk handles the "multihost pods" of TPUs and GPUs (HGX H100) as
-first class citizens.
+first class citizens. xpk now also supports n2-standard-32 CPUs.
```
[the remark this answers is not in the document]
RoshaniN: Sure.

xpk.py
RoshaniN: I left it here because it is a no-op for TPUs and GPUs, considering that we don't have the workload pods run on CPUs of default pool anyway. I could move this to be CPU-specific though and will add a comment (thanks for noticing that.)

README.md
```diff
@@ -314,6 +317,36 @@ In order to use XPK for GPU, you can do so by using `device-type` flag.
...
+    ```shell
+    # Submit a workload
+    python3 xpk.py workload create \
+    --cluster xpk-test \
+    --num-slices=1 \
+    --device-type=n2-standard-32-256 \
+    --workload xpk-test-workload \
+    --command="echo hello world"
+    ```
+
```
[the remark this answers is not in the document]
RoshaniN: I plan to have the documentation in Maxtext emulator for storage at the moment, but can definitely build a guide in Maxtext soon.

xpk.py
RoshaniN: Thank you!

Obliviour (approved):

xpk.py
```diff
@@ -2198,6 +2272,72 @@ def create_machine_label(accelerator_type, system) -> str:
     return f"{AcceleratorTypeToAcceleratorCharacteristics[accelerator_type].machine_label}: {system.topology}"
   return ""
 
+def calculate_process_count(num_slices, vms_per_slice) -> str:
```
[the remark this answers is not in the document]
RoshaniN: Added now, thanks for noticing!

PR #57, opened 2024-01-18
Add --tpu-topology flag for specifying custom topology types

Obliviour:
## Fixes / Features
- Supports tpu-topology flag for specifying custom topologys for TPUs
-

## Testing / Documentation
Added a check to make sure the format of topologies fits with the argument expected and the command fails if the format is unexpected.

- [ y ] Tests pass

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,7 +178,8 @@ all zones.
 
     ```shell
     python3 xpk.py cluster cacheimage \
-    --cluster xpk-test --docker-image gcr.io/your_docker_image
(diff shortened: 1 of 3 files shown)


rwitten (approved):

Obliviour:
The workload creation needs to know the topology as well so CURRENTLY the user needs to set the same topology for workload create and cluster create. This only adds the flag to cluster create.

In this case, users would have to manage the topology.

PR #60, opened 2024-01-25
Refactor xpk command argument for Disruption handling.

abhinavclemson:
## Fixes
- Enable disruption handling

## Testing
Testing details.

- [ y ] Tests pass

### DIFF
--- a/xpk.py
+++ b/xpk.py
@@ -2123,7 +2123,7 @@ def get_main_container(args, system, docker_image, command, resource_type) -> st
                 - bash
                 - -c
                 - |
-                  echo XPK Start: $(date) ; {command} ; EXIT_CODE=$? ; echo XPK End: $(date); echo EXIT_CODE=$EXIT_CODE ; sleep 5; exit $EXIT_CODE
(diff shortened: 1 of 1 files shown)


xpk.py
```diff
@@ -2057,7 +2057,7 @@ def get_main_container(args, system, docker_image, command, resource_type) -> st
                 - bash
                 - -c
                 - |
-                  echo XPK Start: $(date) ; {command} ; EXIT_CODE=$? ; echo XPK End: $(date); echo EXIT_CODE=$EXIT_CODE ; sleep 5; exit $EXIT_CODE
+                  echo XPK Start: $(date) ; {command}
```
Obliviour, line 2060: Sorry for the back and forth here and slow response but I think we need to understand the fix for keeping / bringing back `EXIT_CODE=$? ; echo XPK End: $(date); echo EXIT_CODE=$EXIT_CODE ; sleep 5; exit $EXIT_CODE`.

Though these logs do not directly impact our user workloads, I don't want to unexpectedly break our user workloads by removing this. I observe that users view XPK End commands as a sign of their workload completing. Without it, we have no consistent way to mark this.

Can we talk with the GKE team to understand why bash solutions that should work are not working here  for disruption handling? I don't think we gain anything much by this CL before solving this?

Obliviour: @danielvegamyhre From what @abhinavclemson was describing, we are having troubles with disruption handling in the jobset api when keeping the statements after the {command}. I believe @abhinavclemson had a solution that worked in bash but didnt work with jobset api. This seems strange. @abhinavclemson if that is correct, can you investigate this more with @danielvegamyhre ?

Obliviour (approved):

PR #62, opened 2024-02-02
Support multiple workload deletion from single workload delete command

SurbhiJainUSC:
## Fixes / Features
- Support deletion of all workloads in the cluster
- Support deletion of workloads filtering by job name in the cluster

## Testing / Documentation
Testing details.

- [ y ] Tests pass

### DIFF
--- a/README.md
+++ b/README.md
@@ -235,7 +235,7 @@ checkpointing so the job restarts near where it was interrupted.
     --workload xpk-test-workload --cluster xpk-test
     ```
 
-    This will only delete `xpk-test-workload` workload in `xpk-test` cluster. Deletion will only begin if you type `Y` at the prompt.
+    This will only delete `xpk-test-workload` workload in `xpk-test` cluster.
 
 *   Workload Delete (delete all training jobs in the cluster):
 
@@ -244,7 +244,7 @@ checkpointing so the job restarts near where it was interrupted.
     --cluster xpk-test
     ```
 
-    This will delete all the workloads in `xpk-test` cluster. Deletion will only begin if you type `Y` at the prompt for every workload.
+    This will delete all the workloads in `xpk-test` cluster. Deletion will only begin if you type `y` or `yes` at the prompt.
 
 *   Workload Delete supports filtering. Delete a portion of jobs that match user criteria.
     * Filter by Job: `filter-by-job`
@@ -254,7 +254,7 @@ checkpointing so the job restarts near where it was interrupted.
     --cluster xpk-test --filter-by-job=$USER
     ```
 
-    This will delete all the workloads in `xpk-test` cluster whose names start with `$USER`. Deletion will only begin if you type `Y` at the prompt for every workload.
+    This will delete all the workloads in `xpk-test` cluster whose names start with `$USER`. Deletion will only begin if you type `y` or `yes` at the prompt.
 
 ## Workload List
 *   Workload List (see training jobs):
--- a/README.md
+++ b/README.md
@@ -466,7 +466,7 @@ gcloud beta compute reservations list --project=$PROJECT_ID
 gcloud beta compute reservations describe $RESERVATION --project=$PROJECT_ID --zone=$ZONE
 ```
 
-# Workload Debugging
+# TPU Workload Debugging
 
 ## Collect Stack Traces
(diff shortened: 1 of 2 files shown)


xpk.py
```diff
@@ -2481,14 +2481,37 @@ def workload_delete(args) -> int:
...
+    # Skip the header
+    workloads = return_value.strip().split("\n")[1:]
+  else:
+    workloads = [args.workload]
+
+  if not workloads:
+    xpk_print("There are no workloads to delete matching the filter in the cluster.")
+
+  for workload in workloads:
+    result = input(f'Do you want to delete the workload: {workload} (Y/n)? ')
```
Obliviour, line 2504: I like the yes / no user feedback logic here! two suggestions:

1. Can the user enter yes/no/y/n once per workload delete command, instead of once per workload. We can print out all the workloads that are planned to be deleted. (Same as https://github.com/google/xpk/blob/main/xpk.py#L1396-L1401)

I think it would be hard to ask the user to type yes several times. 

2. Can you refactor the user input y/yes/n/no logic we use in node pool delete into a function and use that here as well. I think we will also need it in cluster deletion as well for a new feature request. (https://github.com/google/xpk/blob/main/xpk.py#L1396-L1401)

xpk.py
```diff
@@ -3187,18 +3224,25 @@ def directory_path_type(value):
...
+    help='The name of the cluster to delete the job on.',
+    required=True,
+)
+
+### Optional arguments
+workload_delete_parser_optional_arguments.add_argument(
     '--workload',
     type=workload_name_type,
     default=None,
     help='The name of the workload to delete.',
```
Obliviour, line 3239: Can you add that if the workload is not specified, the default behavior is that all workloads will be deleted from the cluster?

README.md
```diff
@@ -235,6 +235,27 @@ checkpointing so the job restarts near where it was interrupted.
...
+    This will only delete `xpk-test-workload` workload in `xpk-test` cluster. Deletion will only begin if you type `Y` at the prompt.
+
+*   Workload Delete (delete all training jobs in the cluster):
+
+    ```shell
+    python3 xpk.py workload delete \
+    --cluster xpk-test
+    ```
+
+    This will delete all the workloads in `xpk-test` cluster. Deletion will only begin if you type `Y` at the prompt for every workload.
```
Obliviour, line 247: nit: to adjust the readme that users only need to type yes/y once at the prompt.

xpk.py
```diff
@@ -3187,18 +3224,25 @@ def directory_path_type(value):
...
 ### Required arguments
 workload_delete_parser_required_arguments.add_argument(
+    '--cluster',
+    type=str,
+    default=None,
+    help='The name of the cluster to delete the job on.',
+    required=True,
+)
+
+### Optional arguments
```
Obliviour, line 3234: Can you add a `--force` argument that skips the user prompt? This is good for automated tests for example. Or if a user really believes in themselves. You can build from https://github.com/google/xpk/blob/main/xpk.py#L2823-L2830

xpk.py
```diff
@@ -2481,14 +2481,37 @@ def workload_delete(args) -> int:
...
+    xpk_print("Get the name of the workloads in the cluster.")
+    return_code, return_value = get_workload_list(args, columns)
+
+    if return_code != 0:
+      xpk_print(f'List Job request returned ERROR {return_code}')
+      xpk_exit(return_code)
+    # Skip the header
+    workloads = return_value.strip().split("\n")[1:]
+  else:
+    workloads = [args.workload]
```
Obliviour, line 2498: Hmm I think we should skip the prompt if they use args.workload since that keeps the current behavior the same. Users are currently happy with the delete single workload case and I don't think we need to doublecheck users there.

Obliviour (commented):
Looks great! I already heard of a customer interested in using this feature so this will be eagerly used!

xpk.py
SurbhiJainUSC: Updated!

xpk.py
SurbhiJainUSC: Added!

README.md
SurbhiJainUSC: Done!

xpk.py
SurbhiJainUSC: Sure, added!

xpk.py
SurbhiJainUSC: Gotcha, updated!

xpk.py
```diff
@@ -2560,6 +2597,28 @@ def determine_workload_list_filter_by_job(args) -> str:
     return workload_list_awk_command(f'{job_name_arg} ~ \"{args.filter_by_job}\"')
 
 
+def get_workload_list(args, columns) -> None:
+  """Function to get the list of the workloads in the cluster.
```
Obliviour, line 2601: nit: maybe add a comment that this function is used by other internal xpk functions that need to list workloads. Not used directly by `workload list` command

Obliviour: actually I see better how this is used. I think it is good as is!

Obliviour (approved):
LGTM, great set of changes! Thank you

PR #74, opened 2024-02-22
Pathways integration with XPK

RoshaniN:
## Fixes / Features
- Enable Pathways cluster creation using "--enable-pathways". Additional CPU nodepools for RM, Proxy and user workloads are created, along with the requested TPU resources.
- Accept a Pathways workload using "--use-pathways" and a user-provided docker image (currently Maxtext).
- Making Cluster Queue and Resource Flavor changes to allow management of Pathways resources.
- Pathways enabled cluster can also accept regular multislice-TPU workloads!
- Bumping Jobset version to 0.4.0 and using exclusive topology only for TPU workers.
- Allow Pathways workloads to run in parallel as long as TPU resources are available, CPU resources are autoscaled.

## Testing / Documentation
Following tests were completed:
- Cluster creation with --enable-pathways creates CPU nodepools.
- Cluster queue handles workloads in parallel as long as TPUs are available, CPUs must autoscale.
-- Pathways + regular workloads in parallel
-- Pathways + Pathways workloads in parallel
- Workloads with --use-pathways schedule and execute as expected.
- Resources are returned to the queue once user workload succeeds/fails.
- Workload delete works as expected.
- XPK cluster can handle both pathways and non-pathways workloads.
- Graceful exit message when Pathways workload is created without enabling pathways cluster.

- [ y ] Tests pass


xpk.py
```diff
@@ -114,6 +114,183 @@
                 name: dshm-2
 """
 
+pw_workload_create_yaml = """apiVersion: jobset.x-k8s.io/v1alpha2
+kind: JobSet
+metadata:
+  annotations:
+    alpha.jobset.sigs.k8s.io/exclusive-topology: cloud.google.com/gke-nodepool
+  name: {args.workload}-pathways
```
shaurya89, line 122: Any reason to add the "-pathways" prefix?

xpk.py
```diff
@@ -114,6 +114,183 @@
...
+        template:
+          spec:
+            containers:
+            - args:
+              {pathways_worker_args}
+              env:
+              - name: TPU_STDERR_LOG_LEVEL
+                value: "0"
+              - name: TPU_MIN_LOG_LEVEL
+                value: "0"
```
shaurya89, line 153: The default here is already 0, oso fine to drop variable

xpk.py
```diff
@@ -114,6 +114,183 @@
...
+              - name: TF_CPP_MIN_LOG_LEVEL
```
shaurya89, line 154: I don't think we need this variable either

xpk.py
```diff
@@ -114,6 +114,183 @@
...
+                value: "0"
+              - name: XCLOUD_ENVIRONMENT
```
shaurya89, line 156: This is only relevant with python code, so shouldn't be needed here

xpk.py
```diff
@@ -114,6 +114,183 @@
...
+                    fieldPath: metadata.annotations['jobset.sigs.k8s.io/replicatedjob-name']
+              - name: JOBSET_NAME
+                valueFrom:
+                  fieldRef:
+                    fieldPath: metadata.annotations['jobset.sigs.k8s.io/jobset-name']
+              - name: HOST_ADDRESS
+                value: $(JOBSET_NAME)-$(REPLICATED_JOB_NAME)-0-0.$(JOBSET_NAME)
+              - name: TPU_SKIP_MDS_QUERY
+                value: "true"
+              - name: XCLOUD_ENVIRONMENT
```
shaurya89, line 214: Drop this as well

xpk.py
```diff
@@ -114,6 +114,183 @@
...
+                limits:
+                  cpu: "4"
+                  memory: 8G
+              securityContext:
+                privileged: true
+              volumeMounts:
+              - mountPath: /tmp
+                name: shared-tmp
+            nodeSelector:
+              cloud.google.com/gke-nodepool: cpu-rm-np
```
shaurya89, line 231: Maybe we need to address this in a separate PR, but GKE now supports running multiple CPU containers (RM, Proxy and USer job) on a single nodepool. Please check with Yash on that

xpk.py
```diff
@@ -114,6 +114,183 @@
...
+              name: pathways-proxy
+              ports:
+              - containerPort: 38676
+              resources:
+                limits:
+                  cpu: "24"
+                  memory: 100G
+            nodeSelector:
+              cloud.google.com/gke-nodepool: cpu-proxy-np
+  - name: user-deploy
```
shaurya89, line 263: Maybe rename this to "main" to be explicit that this is the starting point of the job.

xpk.py
```diff
@@ -158,6 +359,26 @@
...
       - name: "{resource_type}"
         nominalQuota: {total_chips}  # Number of slices * number of chips in each slice
+  - coveredResources: ["cpu", "memory"]    
+    flavors:
+    - name: cpu-rm
+      resources:
+      - name: "cpu"
+        nominalQuota: 6
+      - name: "memory"
+        nominalQuota: 10G
```
shaurya89, line 369: What is the diff between the resources specified here and in the yaml above?

xpk.py
```diff
@@ -2112,44 +2390,121 @@ def get_main_container(args, system, docker_image, resource_type) -> str:
...
+    str:
+      YAML with the env config for the main container, as a YAML string.
+  """
+  env_yaml="""
+                - name: XCLOUD_ENVIRONMENT
+                  value: GCP
+                - name: JAX_PLATFORMS
+                  value: proxy
+                - name: JAX_BACKEND_TARGET
+                  value: grpc://{args.workload}-pathways-proxy-0-0.{args.workload}-pathways:38676"""
```
shaurya89, line 2452: Remove "-pathways" if you remove that prefix from the jobset name

xpk.py
```diff
@@ -2363,11 +2718,60 @@ def get_cpu_affinity(accelerator_type) -> str:
...
 
+def get_pathways_rm_args(args) -> str:
+  """ Arguments for the Pathways resource manager."""
+  yaml="""- --alsologtostderr
+              - --pathways_server_port=38677
+              - --pathways_server_provides_devices=false
+              - --pathways_device_type=NONE
+              - --pathways_persistent_compilation_cache=false
+              - --pathways_compilation_mode=compile_at_worker
+              - --pathways_tmp_dir_pattern=gs://cloud-pathways-staging/tmp
```
shaurya89, line 2734: Can we parameterize this location for the user to set?

shaurya89 (changes requested):
Thanks. Just some minor comments.

xpk.py
RoshaniN: The Jobset was initially called "pathways", I left it in there to differentiate regular workloads from pathways workloads.

xpk.py
RoshaniN: Thanks for catching that, will do!

xpk.py
RoshaniN: Thanks for catching that, will do!

xpk.py
RoshaniN: Great to know that! Will check how to imbibe it in this/  an upcoming PR

xpk.py
```diff
@@ -114,6 +114,183 @@
...
+    xpk.google.com/workload: {args.workload}
+spec:
+  failurePolicy:
+    maxRestarts: {args.max_restarts}
+  successPolicy:
+    operator: "All"
+    targetReplicatedJobs: 
+    - "user-deploy"
+  replicatedJobs:
+  - name: worker
```
guptaaka, line 134: Is there an acronym for pathways that we can add to the replicated jobs? Something like pw-worker, pw-rm, pw-proxy. The only reason I suggest that is, it would be easier to find all the pathways-related jobs in a cluster, especially if a cluster is running many sorts of services. "Proxy" and "worker" are very generic names and it might get challenging to find the right job for pathways.

wdyt? @shaurya89 @RoshaniN

xpk.py
RoshaniN: ResourceFlavors are mostly the nodeLabels, in case of CPUs, they are just the nodepool names. 
Here we specify what kind of resources the cluster queue should expect, which in case of Pathways is TPUs + CPUs and memory.

These are the resources I learned from -
1. https://cloud.google.com/kubernetes-engine/docs/tutorials/kueue-intro#create_the_resourceflavor - more on resource flavor
2. https://kueue.sigs.k8s.io/docs/tasks/administer_cluster_quotas/#1-create-a-clusterqueuedocsconceptscluster_queue  - resources in cluster queue.

LMK if you see an issue!

xpk.py
RoshaniN: Thanks, I missed it, will do!

xpk.py
RoshaniN: Got it.

xpk.py
```diff
@@ -114,6 +114,183 @@
...
+                privileged: true
+              volumeMounts:
+              - mountPath: /tmp
+                name: shared-tmp
+            nodeSelector:
+              {accelerator_label}
+              {machine_label}
+            priorityClassName: {args.priority}
+            # hostNetwork: true
+            # dnsPolicy: ClusterFirstWithHostNet
```
guptaaka, line 178: Did you mean to remove these? I see some other comments too in the file.

xpk.py
guptaaka: That sounds good to me. If we end up removing "pathways" in the other comment, we might want to reconsider here. Thanks Roshani!

xpk.py
RoshaniN: Yes, cleaning up of comments remains, will work on that later today. Thanks for noticing, Akanksha!

xpk.py
```diff
@@ -1107,6 +1328,8 @@ def run_gke_cluster_create_command(args) -> int:
       f' --cluster-version={args.gke_version} --location-policy=BALANCED'
       f' --machine-type={machine_type}'
       ' --scopes=storage-full,gke-default'
+      ' --enable-ip-alias '
+      f' --create-subnetwork name={args.cluster}-subnetwork '
```
guptaaka, line 1332: Which network will this subnetwork be created in? A project can have multiple networks.

guptaaka, line 1332: Some customers use shared VPC. Does XPK allow a user to select an option to use shared VPC? If yes, would this subnetwork also be created in the shared VPC? cc @Obliviour

RoshaniN: I intend to use this https://github.com/google/xpk/blob/main/xpk.py#L2877 variable and test. I have not tested shared VPC so far. Thanks for the resources provided offline!

RoshaniN: I will check that, currently I believe it uses the default network.

xpk.py
Obliviour: You can also use labels:

and define a new label if you want to parse jobs and jobsets. See xpk.google.com/workload label in the jobset and pods level as two examples.

xpk.py
Obliviour: In theory this should remove the need for more than one cpu node pool -- aka just have the default cpu node pool. Would be a helpful simplification in the cluster creation for pathways. It might make it easier to do resource management for kueuing of jobs.

xpk.py
Obliviour: What's the purpose of creating the subnetwork?

xpk.py
```diff
@@ -158,6 +337,26 @@
       resources:
       - name: "{resource_type}"
         nominalQuota: {total_chips}  # Number of slices * number of chips in each slice
+  - coveredResources: ["cpu", "memory"]    
+    flavors:
+    - name: cpu-rm
+      resources:
+      - name: "cpu"
+        nominalQuota: 6
```
Obliviour, line 345: How is nominalQuota for cpu/memory calculated? Should this be dynamic since we have autoscaling enabled for the pathways cpu node pools?

xpk.py
```diff
@@ -2129,10 +2386,10 @@ def get_main_container(args, system, docker_image, resource_type) -> str:
 
   yaml = """- name: {args.docker_name}
                 image: {docker_image}
-                env: {args.env}
+                imagePullPolicy: Always
```
Obliviour, line 2389: Hmm I am not sure if this setting is benefitical for all TPU cases with non-pathways. I remember it slowed down workload start up at large scale (like 5k vms) workloads. 

We can make this pathways specific for now to avoid changing non-pathways flows for now

xpk.py
```diff
@@ -2368,11 +2698,60 @@ def get_cpu_affinity(accelerator_type) -> str:
...
                         values:
                         - default-pool
-  """
+"""
   if accelerator_type == AcceleratorType['CPU']:
     return yaml
   return ""
 
+def get_pathways_rm_args(args) -> str:
+  """ Arguments for the Pathways resource manager."""
```
Obliviour, line 2707: nit to add doc string for input and output args for the three new functions

xpk.py
```diff
@@ -3091,7 +3511,7 @@ def directory_path_type(value):
         'but if your docker container is missing the dependencies, it might '
         'look more like "--command=\'bash setup.sh && python3 train.py\'".'
     ),
-    required=True,
+    required=False,
```
Obliviour, line 3514: When is --command not needed?

Small edge case:
We might need a if statement check in xpk to check that command is set if --use-pathways is not provided.

xpk.py
RoshaniN: Removed "pathways"

xpk.py
RoshaniN: Thanks, that makes sense!

xpk.py
RoshaniN: Hoping to figure this out soon. At the moment, I am having issues having multiple workloads share the CPU nodepools. Issue is observed when one node is occupied with a workload but other nodes are available. This happens for RM, Proxy and User CPU nodepools.

xpk.py
RoshaniN: This was added to Pathways cluster creation to mitigate this error, but is widely useful for all GKE cluster creation -
ERROR: (gcloud.container.clusters.create) ResponseError: code=400, message=This operation will exceed max secondary ranges per subnetwork (30) for subnet "default", consider reusing existing secondary ranges or use a different subnetwork.

xpk.py
RoshaniN: SG, makes sense. Thanks for flagging that!

xpk.py
RoshaniN: Thanks for the suggestion. For now, I have set it to max-nodes (With autoscaling) * resources needed for a workload. Exploring the best set of values here. Not sure if they can be dynamic, will check.

xpk.py
RoshaniN: Removed Pathways prefixes, user can add them if they wish, as we discussed.

xpk.py
RoshaniN: Keeping the resource flavors to have the maximum resource limits with autoscaling enabled.

xpk.py
RoshaniN: Moved this to be Pathways specific for now, but I think this might be widely useful.

RoshaniN: Will test shared VPC once requirements are clearer, perhaps part of the next PR.

xpk.py
```diff
@@ -3475,7 +3910,37 @@ def directory_path_type(value):
...
+workload_pathways_workload_arguments.add_argument(
+    '--proxy-server-image', 
+    type=str,
+    default='gcr.io/cloud-tpu-multipod-dev/pathways-prototype/gke/pathways-demo:proxy_server',
+    help=(
+        'Please provide the proxy server image for Pathways.'
+    ),
+)
+workload_pathways_workload_arguments.add_argument(
+    '--server-image',
```
RoshaniN, line 3929: @shaurya89 Any thoughts on whether we should provide default images or have users provide the same (i.e mark this argument as "required")?

guptaaka:
> * Pathways enabled cluster can also accept regular workloads!

Minor: frm the commit message, by regular, do you mean "Pathways accepts non-TPU workloads" or "the cluster accepts workloads outside of Pathways"?

RoshaniN:
> > * Pathways enabled cluster can also accept regular workloads!
> 
> Minor: from the commit message, by "regular", do you mean "Pathways accepts non-TPU workloads" or "the cluster accepts workloads outside of Pathways"?



> :

I meant other multislice-TPU workloads, that are not Pathways. Thanks for raising this, I will clarify.

xpk.py
```diff
@@ -1419,6 +1623,28 @@ def run_gke_node_pool_create_command(args, system) -> int:
     commands.append(command)
     task_names.append(task)
 
+  desired_pw_cpu_node_pools = ['cpu-user-np', 'cpu-rm-np', 'cpu-proxy-np']
```
RoshaniN, line 1626: This snippet should "update" an existing cluster to add Pathways CPU nodepools as well, without explicitly deleting and recreating the cluster (as long as tpu-type remains the same). @shaurya89

xpk.py
```diff
@@ -1611,7 +1838,7 @@ def set_jobset_on_cluster(args) -> int:
   """
   command = (
       'kubectl apply --server-side -f'
-      ' https://github.com/kubernetes-sigs/jobset/releases/download/v0.3.2/manifests.yaml'
+      ' https://github.com/kubernetes-sigs/jobset/releases/download/v0.4.0/manifests.yaml'
```
shaurya89, line 1841: I would recommend moving this version bump into a separate PR unless pathways is requiring this somehow

shaurya89 (commented):
Changes look good. I would recommend taking the pathways helper methods out of this file (it is already too long) and moving them into a separate file for better readability

xpk.py
RoshaniN: This is actually needed for Pathways. This version enabled exclusive-topology at replicatedJobs level, which is helping us scale up cpu nodepools.

RoshaniN:
> 

Great suggestion! I will do that!

shaurya89:
> Changes look good. I would recommend taking the pathways helper methods out of this file (it is already too long) and moving them into a separate file for better readability

I am not sure what design you used to separate pathways logic into the helper class but I still see a bunch of pathways specific things in `xpk.py`. Can you create a TODO somewhere to revisit this design?

shaurya89 (approved):

README.md
```diff
@@ -125,6 +125,16 @@ all zones.
     --num-slices=4 --spot
     ```
 
+* Cluster Create for Pathways:
```
Obliviour, line 128: Can we move the Pathways details to a specific section of the readme? I think it would keep the readme cleaner.

We can also give a brief description to what Pathways is as well.

RoshaniN:
> > Changes look good. I would recommend taking the pathways helper methods out of this file (it is already too long) and moving them into a separate file for better readability
> 
> I am not sure what design you used to separate pathways logic into the helper class but I still see a bunch of pathways specific things in `xpk.py`. Can you create a TODO somewhere to revisit this design?

I moved all methods exclusive to Pathways to the helper file. Some of the other methods are shared between non-pathways and pathways workloads. I will make a ToDo and revisit this in my next PR. Thanks!

xpk.py
```diff
@@ -120,6 +121,161 @@
                 name: dshm-2
 """
 
+pw_workload_create_yaml = """apiVersion: jobset.x-k8s.io/v1alpha2
```
Obliviour, line 124: This yaml would fit nicely in the pathways_xpk_helpers.py module since it is pathways specific.

xpk.py
```diff
@@ -149,6 +305,30 @@
     {machine_label}
 ---
 apiVersion: kueue.x-k8s.io/v1beta1
+kind: ResourceFlavor
```
Obliviour, line 308: These resource flavors are PW specific. We shouldn't apply them when the cluster is not using pathways right?

xpk.py
Obliviour: Makes sense to bump this. The con of bumping it as part of this PR is that we now multiple points of potential breakage from one PR. 

Having it in a separate PR allows for integ tests and for us to check for issues against the version change specifically.

xpk.py
```diff
@@ -2192,18 +2420,83 @@ def get_main_container(args, system, docker_image, resource_type) -> str:
...
                 resources:
                   limits:
-                    {resource_type}: {system.chips_per_vm}
+                    {resources}
+                volumeMounts:
+                {pw_volume_mounts}
   """
   return yaml.format(args=args,
                    system=system,
+                   image_pull_policy=pw.add_image_pull_policy_for_pw(args),
```
Obliviour, line 2429: Why are some of these functions pw specific and some not?

I like the idea of pathways_xpk_helpers.py but we should include get_env_container that does the same.

---

I think we should use pathways_xpk_helper.py to store the yaml configurations that are pathways specific. Keep the functions that are shared between xpk and pathways in xpk.

Yaml configurations like the kueue settings, workload creation command, pathways yaml file are unique to pathways and can go in the pathways_helpers.py fille. 

The functions like get_env_container/ add_image_pull_policy might fork more in the future for other applications or new xpk features so we should keep them in xpk.py or some separate file that is xpk_shared_helpers since it could be used outside of pathways. Happy to explain this more! Let me know if it makes sense

xpk.py
```diff
@@ -2490,17 +2783,49 @@ def workload_create(args) -> int:
...
-  tmp = write_temporary_file(yml_string)
-  command = f'kubectl apply -f {str(tmp.file.name)}'
-  return_code = run_command_with_updates(command, 'Creating Workload', args)
+  if args.use_pathways:
+    # Ensure the cluster and CPU nodepools were created with --enable-pathways
+    all_node_pools = get_all_nodepools_programmatic(args)
+    desired_pw_cpu_node_pools = {'cpu-user-np', 'cpu-rm-np', 'cpu-proxy-np'}
+    if not desired_pw_cpu_node_pools.issubset(set(all_node_pools[0])):
+      xpk_print(
+          'Cluster needs to be created with --enable-pathways to run Pathways workloads.'
```
Obliviour, line 2792: nit Could add that the reason this errored is because there are missing cpu node pools, and print the missing ones?

xpk.py
```diff
@@ -2490,17 +2783,49 @@ def workload_create(args) -> int:
...
-                                           container=container,
-                                           affinity=get_cpu_affinity(system.accelerator_type),
-                                           env=get_cpu_env(args.num_slices,system),
-                                           accelerator_label=create_accelerator_label(system.accelerator_type, system),
-                                           machine_label=create_machine_label(system.accelerator_type, system),
-                                           local_queue_name=_LOCAL_QUEUE_NAME)
-  tmp = write_temporary_file(yml_string)
-  command = f'kubectl apply -f {str(tmp.file.name)}'
-  return_code = run_command_with_updates(command, 'Creating Workload', args)
+  if args.use_pathways:
```
Obliviour, line 2786: within the `if args.use_pathways` code would be really good place for a helper function that exists in pathways_xpk_helpers.py.

README.md
RoshaniN: Okay, will create a Pathways section.

xpk.py
Obliviour: How would moving the config file create a circular dependencies?

xpk.py
RoshaniN: get_env_container is also used by non-pw workloads - it returns args.env in that case. (It's a shared function, hence lives in xpk.)

Ohh yes, I am already in the process of moving Pathways YAML extensions to the helper file!

This makes sense! xpk is currently a huge file with lots of helpers, I am happy to address readability improvements in the next PR!

xpk.py
RoshaniN: Sure, makes sense

Obliviour (approved):
Talked offline that we will do the splitting of pathways specific helpers in another PR. For now pathway changes can go directly into xpk.py. Thank you for this change!

xpk.py
RoshaniN: Agreed, I already moved to a helper in the new changes! Thank you

xpk.py
RoshaniN: Discussed offline :)

xpk.py
RoshaniN: For now defaults contain images in cloud-tpu-v2-images

PR #79, opened 2024-02-27
check gsutil installation when debug_dump_gcs is passed

ssusie:
##  Feature
- Checks whether gsutil is installed in docker image when --debug-dump-gcs is passed

## Testing 
Tested offline with and without installation.

### DIFF
--- a/xpk.py
+++ b/xpk.py
@@ -2112,8 +2112,13 @@ def get_main_container(args, system, docker_image, resource_type) -> str:
       yaml for main container
   """
 
-  xpk_internal_commands = ""
+  xpk_internal_commands = ''
+  gsutil_test_command = ''
   if args.debug_dump_gcs:
+    gsutil_test_command = (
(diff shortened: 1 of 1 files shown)


xpk.py
```diff
@@ -2112,8 +2112,13 @@ def get_main_container(args, system, docker_image, resource_type) -> str:
       yaml for main container
   """
 
-  xpk_internal_commands = ""
+  xpk_internal_commands = ''
+  gsutil_test_command = 'echo'
```
Obliviour, line 2116: nit: Curious if we need this to run echo in the default case? Could it be empty instead, probably need to delete the ; from `{gsutil_test_command}**;**` and move the ; at the end of the gsutil_test_command ` is required but not installed. Aborting"; exit 24;}**;**`

Obliviour (approved):

xpk.py
ssusie: Thanks for the suggestion. Done!

PR #86, opened 2024-03-15
Create service account in cluster create

SurbhiJainUSC:
## Fixes / Features
- Create a project-specific service account in cluster create
- All the required permissions for the workload will be assigned to this new service account
- This service account will replace Compute Engine default service account for cluster & node-pools
- This functionality will be done behind a flag `_USE_SERVICE_ACCOUNT` to prevent breakage for current xpk users

## Testing / Documentation
Testing details.

- [ y ] Tests pass
- [ N/A ] Appropriate changes to documentation are included in the PR

### DIFF
--- a/xpk.py
+++ b/xpk.py
@@ -1450,7 +1450,7 @@ def run_gke_cluster_create_command(args) -> int:
       ' --scopes=storage-full,gke-default'
       f' {args.custom_cluster_arguments}'
   )
-  
+
   if _SERVICE_ACCOUNT_FEATURE_FLAG:
     service_account_name = get_service_account_name(args)
     service_account_exists = check_if_service_account_exists(args)
--- a/xpk.py
+++ b/xpk.py
@@ -71,6 +71,9 @@ _CLUSTER_QUEUE_NAME='cluster-queue'
 _LOCAL_QUEUE_NAME='multislice-queue'
 _CLUSTER_RESOURCES_CONFIGMAP = 'resources-configmap'
 _CLUSTER_METADATA_CONFIGMAP = 'metadata-configmap'
+_XPK_SERVICE_ACCOUNT = 'xpk-sa'
+# Set to True to attach a service account to cluster & node pools
+_USE_SERVICE_ACCOUNT = False
 
 
 workload_create_yaml = """apiVersion: jobset.x-k8s.io/v1alpha2
@@ -1112,8 +1115,18 @@ def zone_to_region(zone) -> str:
   return zone_terms[0] + '-' + zone_terms[1]
 
 
-def check_if_service_account_exist(args):
-  """Check if a service account with the given name exist in the project.
+def get_service_account_name(args) -> str:
+  """Get the name for the service account.
+  Args:
+    args: user provided arguments.
+
+  Returns:
+    the name of the service account.
+  """
+  return f'{args.project}-{_XPK_SERVICE_ACCOUNT}@{args.project}.iam.gserviceaccount.com'
+
+def check_if_service_account_exists(args) -> bool:
+  """Check if a service account with the given name exists in the project.
 
   Args:
     args: user provided arguments for running the command.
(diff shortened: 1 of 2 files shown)


xpk.py
```diff
@@ -1097,6 +1109,74 @@ def zone_to_region(zone) -> str:
...
+
+  command = (
+    f'gcloud iam service-accounts create {args.project}-xpk-sa \
+    --description="Service Account for XPK" \
+    --display-name="{args.project}-xpk-sa"'
+  )
+  return_code = run_command_with_updates(command, 'Service Account Create', args)
+  if return_code != 0:
+    xpk_print(f'Service Account Create request returned ERROR {return_code}')
+    xpk_print('Make sure you have Service Account Admin Role attached to your user.')
```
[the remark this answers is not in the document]
SurbhiJainUSC: To avoid breaking users who already have cluster, I have added this SA creation in `run_gke_cluster_create_command()`, which is only called when user is creating a new cluster. . Also, this PR doesn't add SA to node pools yet.

xpk.py
```diff
@@ -1135,6 +1215,8 @@ def run_gke_cluster_create_command(args) -> int:
     xpk_print(f'GKE Cluster Create request returned ERROR {return_code}')
     return 1
 
+  create_sevice_account(args)
```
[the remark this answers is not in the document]
SurbhiJainUSC: Yes, we need to attach service account to the node pools at the time of cluster creation.

`if the cluster is already created, then the service account creation step is skipped.` - yes correct.

xpk.py
```diff
@@ -1127,6 +1142,120 @@ def zone_to_region(zone) -> str:
...
+  roles = []
+  service_account_name = get_service_account_name(args)
+  command = (
+    f'gcloud projects get-iam-policy {args.project}'
+    f' --filter="bindings.members:{service_account_name}"'
+    ' --flatten="bindings[].members" --format="table(bindings.role)"'
+  )
+  return_code, return_value = run_command_for_value(command, 'Get IAM Roles For Service Account', args)
+  if return_code != 0:
+    xpk_print(f'Get IAM Roles For Service Account request returned ERROR {return_code}')
```
Obliviour, line 1212: If return_code is !=0, we hit this print statement but the function will return an empty list of existing roles and continue running. LMK if that is the expected behavior you want or do you want xpk to exit early?

xpk.py
```diff
@@ -1127,6 +1142,120 @@ def zone_to_region(zone) -> str:
...
+  """
+  command = (
+    f'gcloud iam service-accounts create {args.project}-{_XPK_SERVICE_ACCOUNT} \
+    --description="Service Account for XPK" \
+    --display-name="{args.project}-{_XPK_SERVICE_ACCOUNT}"'
+  )
+  return_code = run_command_with_updates(command, 'Service Account Create', args)
+  if return_code != 0:
+    xpk_print(f'Service Account Create request returned ERROR {return_code}')
+    xpk_print('Make sure you have Service Account Admin Role attached to your user.')
```
Obliviour, line 1191: Can you add a print statement that says: xpk_print('Please look at https://github.com/google/xpk/blob/main/README.md#permission-issues-requires-one-of-permission_name-permissions for how to add permissions to your user.')

xpk.py
```diff
@@ -1127,6 +1142,120 @@ def zone_to_region(zone) -> str:
...
+    command = (
+      f'gcloud projects add-iam-policy-binding {args.project} \
+      --member="serviceAccount:{service_account_name}" \
+      --role="{role}" \
+      --condition=None'
+    )
+    return_code = run_command_with_updates(command, 'Add IAM Role to Service Account', args, verbose=False)
+    if return_code != 0:
+      xpk_print(f'Add IAM Role to Service Account request returned ERROR {return_code}')
+      xpk_print('Make sure you have Project IAM Admin Role attached to your user.')
```
Obliviour, line 1254: Can you add a print statement that says: xpk_print('Please look at https://github.com/google/xpk/blob/main/README.md#permission-issues-requires-one-of-permission_name-permissions for how to add permissions to your user.')

xpk.py
```diff
@@ -1097,6 +1109,74 @@ def zone_to_region(zone) -> str:
   return zone_terms[0] + '-' + zone_terms[1]
 
 
+def check_if_service_account_exist(args):
```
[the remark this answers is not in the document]
SurbhiJainUSC: Done!

xpk.py
```diff
@@ -1097,6 +1109,74 @@ def zone_to_region(zone) -> str:
...
+  """Check if a service account with the given name exist in the project.
```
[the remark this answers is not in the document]
SurbhiJainUSC: Updated!

xpk.py
```diff
@@ -1097,6 +1109,74 @@ def zone_to_region(zone) -> str:
...
+
+  Returns:
+    True if service account exist, False otherwise.
+  """
+  command = (
+    f'gcloud iam service-accounts describe {args.project}-xpk-sa@{args.project}.iam.gserviceaccount.com'
+  )
+  return_code = run_command_with_updates(command, 'Service Account Describe', args, verbose=False)
+  if return_code != 0:
+    xpk_print(f'Service Account Describe request returned ERROR {return_code}')
```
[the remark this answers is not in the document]
SurbhiJainUSC: Updated!

xpk.py
```diff
@@ -1097,6 +1109,74 @@ def zone_to_region(zone) -> str:
...
+  """Check if a service account with the given name exist in the project.
+
+  Args:
+    args: user provided arguments for running the command.
+
+  Returns:
+    True if service account exist, False otherwise.
+  """
+  command = (
+    f'gcloud iam service-accounts describe {args.project}-xpk-sa@{args.project}.iam.gserviceaccount.com'
```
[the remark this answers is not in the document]
SurbhiJainUSC: Done!

xpk.py
```diff
@@ -1097,6 +1109,74 @@ def zone_to_region(zone) -> str:
...
+  )
+  return_code = run_command_with_updates(command, 'Service Account Describe', args, verbose=False)
+  if return_code != 0:
+    xpk_print(f'Service Account Describe request returned ERROR {return_code}')
+    return False
+  return True
+
+def create_sevice_account(args):
```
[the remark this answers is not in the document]
SurbhiJainUSC: Done!

xpk.py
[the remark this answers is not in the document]
SurbhiJainUSC: Attached the newly created service account to cluster as well as node pools.

xpk.py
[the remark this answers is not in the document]
SurbhiJainUSC: Moved the service account creation code behind a flag which is be default turned off.

xpk.py
```diff
@@ -1160,6 +1289,14 @@ def run_gke_cluster_create_command(args) -> int:
       ' --scopes=storage-full,gke-default'
       f' {args.custom_cluster_arguments}'
   )
+  if _USE_SERVICE_ACCOUNT:
+    service_account_name = get_service_account_name(args)
+    service_account_exists = check_if_service_account_exists(args)
+    if service_account_exists:
+      command += f' --service-account={service_account_name}'
+    else:
+      xpk_print(f'Service Account: {service_account_name} does not exist in the project.'
```
Obliviour, line 1298: nit comment here saying: we will not create the service account for previously created GKE clusters. I think this will be useful to reminder us why we have this case.

xpk.py
```diff
@@ -1467,6 +1604,14 @@ def run_gke_node_pool_create_command(args, system) -> int:
     elif system.accelerator_type == AcceleratorType['GPU']:
       command += (' --placement-type=COMPACT ')
       command += f' --accelerator type={system.gke_accelerator},count={str(system.chips_per_vm)}'
+    if _USE_SERVICE_ACCOUNT:
+      service_account_name = get_service_account_name(args)
+      service_account_exists = check_if_service_account_exists(args)
+      if service_account_exists:
+        command += f' --service-account={service_account_name}'
+      else:
+        xpk_print(f'Service Account: {service_account_name} does not exist in the project.'
```
Obliviour, line 1613: nit comment here saying: we will not create the service account for previously created GKE clusters. I think this will be useful to reminder us why we have this case.

xpk.py
```diff
@@ -71,6 +71,9 @@
 _LOCAL_QUEUE_NAME='multislice-queue'
 _CLUSTER_RESOURCES_CONFIGMAP = 'resources-configmap'
 _CLUSTER_METADATA_CONFIGMAP = 'metadata-configmap'
+_XPK_SERVICE_ACCOUNT = 'xpk-sa'
+# Set to True to attach a service account to cluster & node pools
+_USE_SERVICE_ACCOUNT = False
```
Obliviour, line 76: _SERVICE_ACCOUNT_FEATURE_FLAG = Flase

seems like a nicer name? Since this flag means we attempt to attach a SA but we don't always do it (like for previously created clusters).

Obliviour: I think this flag is great though, we can ultimately set it to = (xpk_version >= .4.0 ) or something.

xpk.py
SurbhiJainUSC: For now, I haven't added any documentation related to service account in README. So, directing users to that link won't help. I am thinking to update the README once make this functionality default for everyone. wdyt?

xpk.py
SurbhiJainUSC: Yeah, this is the expected behavior. When this method returns an empty set of roles, then `add_roles_to_service_account()` will add the list of predefined roles to the service account. However, if by chance the role is already attached to the SA, the method won't do anything. So, we can keep the xpk running here.

xpk.py
SurbhiJainUSC: Yes, this is a great suggestion. Updated!

xpk.py
Obliviour: Sounds good! We have a section in the readme already for adding new permissions to the user. We can extend that / rewrite it for service accounts

SurbhiJainUSC: yes, have an AI on me to update the README once we are ready to release this feature.

xpk.py
Obliviour: Is it possible `return_code, return_value = run_command_for_value(command, 'Get IAM Roles For Service Account', args)` will fail for other reasons besides returning an empty set of roles that we would want to abort for? 

If not, this sounds good to me!

SurbhiJainUSC: Yes, it may fail if there is no `resourcemanager.projects.getIamPolicy` permission attached to user. However, I think failing xpk here won't make sense. This is because we are maily interested in attaching the appropriate permissions to the SA. In case, user has `resourcemanager.projects.setIamPolicy` but not `resourcemanager.projects.getIamPolicy`, then they should be able to set the permissions to SA, which is all we want for the cluster to run.

Obliviour (approved):

PR #87, opened 2024-03-15
Nina xpk gpu h100

NinaCai:
## Fixes / Features
- Add cluster and workload creation for A3 cluster

## Testing / Documentation
Ran xpk cluster create successfully in integ tests

- [ y ] Tests pass

### DIFF
--- a/README.md
+++ b/README.md
@@ -447,7 +447,7 @@ In order to use XPK for GPU, you can do so by using `device-type` flag.
 
     ```shell
     python3 xpk.py workload delete \
-    --cluster xpk-test --filter-by-job=$USER --device-type=h100-80gb-8
+    --cluster xpk-test --filter-by-job=$USER
     ```
 
     This will delete all the workloads in `xpk-test` cluster whose names start with `$USER`. Deletion will only begin if you type `y` or `yes` at the prompt.
@@ -456,7 +456,7 @@ In order to use XPK for GPU, you can do so by using `device-type` flag.
 
     ```shell
     python3 xpk.py workload delete \
-    --cluster xpk-test --filter-by-status=QUEUED --device-type=h100-80gb-8
+    --cluster xpk-test --filter-by-status=QUEUED
     ```
     
     This will delete all the workloads in `xpk-test` cluster that have the status as Admitted or Evicted, and the number of running VMs is 0. Deletion will only begin if you type `y` or `yes` at the prompt. Status can be: `EVERYTHING`,`FINISHED`, `RUNNING`, `QUEUED`, `FAILED`, `SUCCESSFUL`.
--- a/README.md
+++ b/README.md
@@ -411,7 +411,7 @@ In order to use XPK for GPU, you can do so by using `device-type` flag.
     # Run cluster create with reservation.
     python3 xpk.py cluster create \
     --cluster xpk-test --device-type=h100-80gb-8 \
-    --num-slices=2 \
+    --num-nodes=2 \
     --reservation=$RESERVATION_ID
     ```
 
@@ -502,48 +502,6 @@ In order to use XPK for GPU, you can do so by using `device-type` flag.
     ```
     
     This will delete all the workloads in `xpk-test` cluster that have the status as Admitted or Evicted, and the number of running VMs is 0. Deletion will only begin if you type `y` or `yes` at the prompt. Status can be: `EVERYTHING`,`FINISHED`, `RUNNING`, `QUEUED`, `FAILED`, `SUCCESSFUL`.
-
-*   Workload List (see training jobs):
-
-    ```shell
-    python3 xpk.py workload list \
-    --cluster xpk-test
-    ```
-
-* Example Workload List Output:
-
-  The below example shows four jobs of different statuses:
-
-  * `user-first-job-failed`: **filter-status** is `FINISHED` and `FAILED`.
-  * `user-second-job-success`: **filter-status** is `FINISHED` and `SUCCESSFUL`.
-  * `user-third-job-running`: **filter-status** is `RUNNING`.
-  * `user-forth-job-in-queue`: **filter-status** is `QUEUED`.
-  * `user-fifth-job-in-queue-preempted`: **filter-status** is `QUEUED`.
-
-  ```
-  Jobset Name                     Created Time           Priority   TPU VMs Needed   TPU VMs Running/Ran   TPU VMs Done      Status     Status Message                                                  Status Time
-  user-first-job-failed           2023-1-1T1:00:00Z      medium     4                4                     <none>            Finished   JobSet failed                                                   2023-1-1T1:05:00Z
-  user-second-job-success         2023-1-1T1:10:00Z      medium     4                4                     4                 Finished   JobSet finished successfully                                    2023-1-1T1:14:00Z
-  user-third-job-running          2023-1-1T1:15:00Z      medium     4                4                     <none>            Admitted   Admitted by ClusterQueue cluster-queue                          2023-1-1T1:16:00Z
-  user-forth-job-in-queue         2023-1-1T1:16:05Z      medium     4                <none>                <none>            Admitted   couldn't assign flavors to pod set slice-job: insufficient unused quota for google.com/tpu in flavor 2xv4-8, 4 more need   2023-1-1T1:16:10Z
-  user-fifth-job-preempted        2023-1-1T1:10:05Z      low        4                <none>                <none>            Evicted    Preempted to accommodate a higher priority Workload             2023-1-1T1:10:00Z
-  ```
-
-* Workload List supports filtering. Observe a portion of jobs that match user criteria.
-
-  * Filter by Status: `filter-by-status`
-
-  Filter the workload list by the status of respective jobs.
-  Status can be: `EVERYTHING`,`FINISHED`, `RUNNING`, `QUEUED`, `FAILED`, `SUCCESSFUL`
-
-  * Filter by Job: `filter-by-job`
-
-  Filter the workload list by the name of a job.
-
-    ```shell
-    python3 xpk.py workload list \
-    --cluster xpk-test --filter-by-job=$USER
-    ```
 
 ## CPU usage
 
--- a/README.md
+++ b/README.md
@@ -326,6 +326,36 @@ In order to use XPK for GPU, you can do so by using `device-type` flag.
     --reservation=$RESERVATION_ID
     ```
 
+*   Cluster Delete (deprovision capacity):
+
+    ```shell
+    python3 xpk.py cluster delete \
+    --cluster xpk-test
+    ```
+
+*   Cluster List (see provisioned capacity):
+
+    ```shell
+    python3 xpk.py cluster list
+    ```
+
+*   Cluster Describe (see capacity):
+
+    ```shell
+    python3 xpk.py cluster describe \
+    --cluster xpk-test
+    ```
+
+
+*   Cluster Cacheimage (enables faster start times):
+
+    ```shell
+    python3 xpk.py cluster cacheimage \
+    --cluster xpk-test --docker-image gcr.io/your_docker_image \
+    --device-type=h100-80gb-8
+    ```
+
+
 *   [Install NVIDIA GPU device drivers](https://cloud.google.com/container-optimized-os/docs/how-to/run-gpus#install)
     ```shell
     # List available driver versions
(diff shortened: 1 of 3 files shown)


xpk.py
```diff
@@ -117,12 +117,160 @@
...
+                  - name: LD_LIBRARY_PATH
+                    value: /usr/local/nvidia/lib64
+                  - name: COMMAND
+                    value: "{command}"
+                  {args.env}
+                command:
+                  - "bash"
+                  - "-c"
+                  - |
+                    echo XPK Start: $(date) ; _sigterm() ( kill -SIGTERM $!;); trap _sigterm SIGTERM; (cd /deps && bash gpu_multi_process_run.sh) & PID=$!; while kill -0 $PID 2>/dev/null; do sleep 5; done; EXIT_CODE=$? ; echo XPK End: $(date); echo EXIT_CODE=$EXIT_CODE; echo Main app is done > /usr/share/maxtext/workload_terminated
```
Obliviour, line 252: "echo Main app is done > /usr/share/maxtext/workload_terminated" seems MaxText specific --> will this work on non-maxtext workloads?

xpk.py
```diff
@@ -117,12 +117,160 @@
...
+                    nvidia.com/gpu: {chips_per_vm}
+"""
+
 workload_delete_yaml = """apiVersion: jobset.x-k8s.io/v1alpha2
 kind: JobSet
 metadata:
   name: {args.workload}
-  annotations:
-    alpha.jobset.sigs.k8s.io/exclusive-topology: cloud.google.com/gke-nodepool # 1:1 job replica to node pool assignment
+  {annotation_config}
```
Obliviour, line 273: Hmm, thinking about the deletion process more, currently we only support deleting TPU jobs, sad. And in order to support GPUs, we also need to add a new required argument for the device type.

---

I am thinking if there is a way to avoid this new required argument since it will break current user flows and shouldn't be needed in theory. 

@danielvegamyhre do you have any thoughts here. We want to delete a jobset just from its name. Do we need the annotation?

xpk.py
```diff
@@ -1138,7 +1379,150 @@ def run_gke_cluster_create_command(args) -> int:
...
+  Returns:
+    0 if successful and 1 otherwise.
+  """
+  existing_network_names, return_code = get_all_networks_programmatic(args)
+  if return_code > 0:
+    xpk_print('Listing all networks failed!')
+    return return_code
+
+  network_name = f'{args.cluster}-net-{index}'
+  if network_name not in existing_network_names:
```
Obliviour, line 1421: do you mind adding an else statement:

```
else:
  xpk_print(f'Reusing existing network {existing_network_names}')
```

xpk.py
```diff
@@ -1138,7 +1379,150 @@ def run_gke_cluster_create_command(args) -> int:
...
+      f' --region={zone_to_region(args.zone)} --range=192.168.{index}.0/24'
+    )
+    return_code = run_command_with_updates(
+        command, 'Create Cluster Subnet', args, verbose=False
+    )
+
+    if return_code != 0:
+      xpk_print(f'Create Cluster Subnet request returned ERROR {return_code}')
+      return 1
+
```
Obliviour, line 1467: Similar else statement:

```
else:
  xpk_print(f'Reusing existing network {existing_network_names}')
```

xpk.py
```diff
@@ -1392,28 +1838,51 @@ def run_gke_node_pool_create_command(args, system) -> int:
...
   commands = []
   task_names = []
+
+  xpk_print(
+      f'Creating {args.num_slices} node pool or pools of {system.device_type}\n'
+      f'Underlyingly, we assume that means: {system}'
+  )
+
+  if system.device_type == h100_device_type:
+    desired_node_pool_names = [f'{args.cluster}-np-0']
```
Obliviour, line 1848: Oh this is interesting, do h100s only have one np? If each slice is in a different NP then the normal logic should work right?

Also is this gpu specific or h100 specific

xpk.py
```diff
@@ -1392,28 +1838,51 @@ def run_gke_node_pool_create_command(args, system) -> int:
...
+    elif system.device_type == h100_device_type:
+      command += (f' --num-nodes={args.num_slices}')
+      command += (f' --accelerator type={system.gke_accelerator},count={str(system.chips_per_vm)}'
+        f' --additional-node-network network={args.cluster}-net-1,subnetwork={args.cluster}-sub-1'
+        f' --additional-node-network network={args.cluster}-net-2,subnetwork={args.cluster}-sub-2'
+        f' --additional-node-network network={args.cluster}-net-3,subnetwork={args.cluster}-sub-3'
+        f' --additional-node-network network={args.cluster}-net-4,subnetwork={args.cluster}-sub-4'
+        ' --no-enable-autoupgrade  --scopes="https://www.googleapis.com/auth/cloud-platform"'
+        )
+    else: # other gpu types
```
Obliviour, line 1884: We also have CPU Types which would fall into this else statement. Can you add an `elif system.accelerator_type == AcceleratorType['GPU']` gpu case? (if needed)

Or add a comment saying other GPU and CPU Types.

xpk.py
```diff
@@ -1598,6 +2073,51 @@ def enable_kueue_crds(args, system) -> int:
   return 0
 
 
+def get_kueue_covered_resources_config(args, cluster_hardware_name, resource_type, total_chips) -> str:
```
Obliviour, line 2076: Just to note this might conflict with pathways changes (https://github.com/google/xpk/pull/74/files). Not sure which PR will go in first but I am happy to work through the merge conflict.

I like this function so hopefully we can use this in pathways PR. cc @RoshaniN

xpk.py
```diff
@@ -2453,25 +3041,41 @@ def workload_create(args) -> int:
...
   debugging_dashboard_id = None
   resource_type = AcceleratorTypeToAcceleratorCharacteristics[system.accelerator_type].resource_type
-  if system.accelerator_type == AcceleratorType['TPU'] and args.deploy_stacktrace_sidecar:
-    xpk_print('Sidecar container to display stack traces for TPU workloads will also be deployed.')
-    container = get_main_and_sidecar_container(args, system, docker_image)
-    # Get GKE debugging dashboard only when sidecar container is deployed for TPU workloads
-    debugging_dashboard_id = get_gke_debugging_dashboard(args)
+
+  device_type = args.tpu_type if args.tpu_type else args.device_type
+  if device_type == h100_device_type:
```
Obliviour, line 3053: nit you can use system.device_type == h100_device_type here

xpk.py
Obliviour: I talked with @danielvegamyhre offline and he suggested that we delete the jobset by name using cli command or python sdk. Let's use CLI command for now since that aligns with our other command usage:

From Daniel:

You can delete a jobset by namespaced name
you'll just need to do it using a different kubectl command, or the python sdk
kubectl delete jobset <name> -n <namespace>

---

So the ask here is to move to `kubectl delete jobset {NAME} -n default`. Then minimizes complexity in workload delete code by a lot, yay.

The namespace we use is default. (https://github.com/google/xpk/blob/main/xpk.py#L1548). Feel free to create a const var for the namespace if you'd like.

xpk.py
NinaCai: This is what we learnt from the GKE doc, both A3 and A3+ have this feature.  (https://docs.google.com/document/d/1ijw1hiAvgNQk6ug8BBsjA2a3opnAM1jHsqdLCAQerTM/edit?pli=1&resourcekey=0-k05v3cpWnx8r0DtZVgfdqw&tab=t.0#heading=h.6wmhdex26cnj):

When creating the A3 node-pool, you could optionally choose to create a nodepool with Kernel Crash Dump enabled, which could capture a full kernel memory crash dump in the event of a kernel crash and save it locally on the instance’s boot disk.

xpk.py
```diff
@@ -874,7 +1101,12 @@ def add_env_config(args):
                        'XLA_FLAGS.')
     env['XLA_FLAGS'] = '--xla_dump_to=/tmp/xla_dump/'
 
-  env_format = '''
+  if device_type == h100_device_type:
```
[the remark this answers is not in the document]
NinaCai: You are right, I just deleted the redundant lines.

xpk.py
```diff
@@ -851,7 +1077,8 @@ def add_env_config(args):
   Args:
     args: user provided arguments for running the command.
   """
-  env = {'JOBSET_NAME': args.workload}
+  device_type = args.tpu_type if args.tpu_type else args.device_type
+  env = {} if device_type == h100_device_type else {'JOBSET_NAME': args.workload}
```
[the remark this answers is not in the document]
NinaCai: We set the `JOBSET_NAME` under the `gpu_workload_create_yaml`, but removing this if-else won't hurt.

xpk.py
NinaCai: It works for `--command "echo goodbye"`, any other workloads you are interested to test?

xpk.py
NinaCai: Looks like Pathways PR is merged already. How do we proceed here for any potential conflict?

xpk.py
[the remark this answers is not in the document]
Obliviour: Oh thank you I see that now. Can you add a comment that explains that difference? What do the env variables get aligned to that's different?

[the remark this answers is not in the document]
NinaCai: It is the variables in `env_file`, for example, in maxtext, it is the XLA_FLAGS.

xpk.py
Obliviour: Why is there a MaxText specific command run in the general xpk command?

If it is only needed for MaxText,  `/usr/share/maxtext/workload_terminated` shouldn't be in the general xpk command right? What is the purpose of `/usr/share/maxtext/workload_terminated` is for.

xpk.py
```diff
@@ -2553,7 +3161,11 @@ def workload_delete(args) -> int:
       args.workload = workload
       yml_string = workload_delete_yaml.format(args=args)
       tmp = write_temporary_file(yml_string)
-      command = f'kubectl delete -f {str(tmp.file.name)}'
+      device_type = args.tpu_type if args.tpu_type else args.device_type
+      if device_type == h100_device_type:
+        command = f'kubectl delete jobset {workload} -n default'
```
Obliviour, line 3166: Thanks Nina, I think we don't even need to parse if it is a TPU / CPU / GPU.

Can we use `kubectl delete jobset {workload} -n default` for all delete cases and delete the --tpu-type and --device-type required arguments for the workload_delete_parser?

This allows us to avoid creating new required arguments, and delete the not needed workload_delete_yaml (big yay).

xpk.py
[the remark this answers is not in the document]
Obliviour: Sounds good, thanks Nina!

xpk.py
```diff
@@ -1115,21 +1343,30 @@ def run_gke_cluster_create_command(args) -> int:
               ' the machine type of other cpu nodepools using `--device-type`.')
     machine_type = args.cluster_cpu_machine_type
 
-  # Create the regional cluster with `num-nodes` CPU nodes in the same zone as
-  # TPUs. This has been tested with clusters of 300 VMs. Larger clusters will
-  # benefit from a larger initial `--num-nodes`. After the cluster is created,
-  # the auto-scaler can reduce/increase the nodes based on the load.
   command = (
       'gcloud beta container clusters create'
-      f' {args.cluster} --release-channel rapid --enable-autoscaling'
```
RoshaniN, line 1124: Hi Nina,
Are you sure you don't want the default nodepool to be autoscaled as per the incoming workloads?

xpk.py
```diff
@@ -1367,23 +1822,15 @@ def run_gke_node_pool_create_command(args, system) -> int:
 
   Args:
     args: user provided arguments for running the command.
-    system: System characteristics based on TPU type/topology.
+    system: System characteristics based on device type/topology.
 
   Returns:
     0 if successful and 1 otherwise.
   """
-  device_type = args.tpu_type if args.tpu_type else args.device_type
```
RoshaniN, line 1375: I am trying to understand why line is being removed.

xpk.py
```diff
@@ -1392,28 +1839,51 @@ def run_gke_node_pool_create_command(args, system) -> int:
...
       continue
     command = (
         'gcloud beta container node-pools create'
-        f' {node_pool_name} --node-version={args.gke_version}'
+        f' {node_pool_name}'
+        f' --region={zone_to_region(args.zone)}'
         f' --cluster={args.cluster}'
         f' --project={args.project} --node-locations={args.zone}'
-        f' --region={zone_to_region(args.zone)}'
-        f' --num-nodes={system.vms_per_slice}'
```
RoshaniN, line 1404: Num nodes is required for CPUs, this won't work for CPUs anymore. I would suggest to add 
"elif system.accelerator_type == AcceleratorType['CPU']:" and append num_nodes, scopes and region there as well.
Thanks

xpk.py
RoshaniN: I made a helper to add resources to kueue config for Pathways, we can work on adding the kueue config conditionally.

If --enable-pathways, {pathways resources are added}
if GPU , {GPU resources are added}

RoshaniN (changes requested):
Thank you for your changes, Nina!

Please pull in the latest changes, you may have to solve a few merge conflicts, but I am happy to help as needed.

Most of Pathways code is present under enable-pathways and use-pathways flags. I believe most of your changes are compatible.

xpk.py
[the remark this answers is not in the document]
Obliviour: Understood,  thank you Nina!

xpk.py
NinaCai: num_nodes and region are still within the command, I moved them either upper or lower in the command.

xpk.py
NinaCai: We are short of capacity, we could keep it without autoscaling this time.

xpk.py
```diff
@@ -1306,25 +1533,33 @@ def run_gke_cluster_create_command(args) -> int:
...
+  else:
+    # Create the regional cluster with `num-nodes` CPU nodes in the same zone as
+    # TPUs. This has been tested with clusters of 300 VMs. Larger clusters will
+    # benefit from a larger initial `--num-nodes`. After the cluster is created,
+    # the auto-scaler can reduce/increase the nodes based on the load.
 
-  if args.enable_pathways:
-    command += (' --enable-ip-alias ')
-    command += (f' --create-subnetwork name={args.cluster}-subnetwork')
+    command += ('--release-channel rapid --enable-autoscaling --location-policy=BALANCED'
```
Obliviour, line 1554: missing space before --release-channel which is causing the build test to fail

README.md
```diff
@@ -415,6 +415,36 @@ In order to use XPK for GPU, you can do so by using `device-type` flag.
     --reservation=$RESERVATION_ID
```
Obliviour, line 415: Change to --num-nodes=2

README.md
```diff
@@ -436,6 +466,85 @@ In order to use XPK for GPU, you can do so by using `device-type` flag.
...
+
+  * Filter by Job: `filter-by-job`
+
+  Filter the workload list by the name of a job.
+
+    ```shell
+    python3 xpk.py workload list \
+    --cluster xpk-test --filter-by-job=$USER
+    ```
+
```
Obliviour, line 547: These lines can be deleted, They are already in the readme in earlier sections and is not GPU specific

xpk.py
```diff
@@ -1344,6 +1726,15 @@ def create_cluster_configmaps(args, system):
   Returns:
     0 if successful and 1 otherwise.
   """
+  device_type = args.tpu_type if args.tpu_type else args.device_type
+  if device_type == h100_device_type:
+    data = f'{device_type}: "{1 * int(args.num_nodes)}"'
```
Obliviour, line 1731: nit "1 * int(args.num_nodes)" can be int(args.num_nodes)

xpk.py
Obliviour: Talking offline, we figured that GPUs will use one node pool.

xpk.py
```diff
@@ -3072,10 +3678,18 @@ def workload_create(args) -> int:
     container = get_main_and_sidecar_container(args, system, docker_image)
     # Get GKE debugging dashboard only when sidecar container is deployed for TPU workloads
     debugging_dashboard_id = get_gke_debugging_dashboard(args)
-  else:
+  elif system.accelerator_type == AcceleratorType['CPU']:
```
Obliviour, line 3681: We also want this to run in the TPU case.

xpk.py
```diff
@@ -3115,9 +3725,42 @@ def workload_create(args) -> int:
...
+                                        pathways_worker_args = get_pathways_worker_args(args),
+                                        pathways_proxy_args = get_pathways_proxy_args(args),
+                                        resource_type=resource_type,
+                                        local_queue_name=_LOCAL_QUEUE_NAME)
     tmp = write_temporary_file(yml_string)
     command = f'kubectl apply -f {str(tmp.file.name)}'
-    return_code = run_command_with_updates(command, 'Creating Workload', args)
+    return_code = run_command_with_updates(command, 'Creating a Pathways Workload', args)
+
+
```
Obliviour, line 3763: This Pathways code (+3732 to +3763) is duplicated. Already exists above. We can delete it here.

Obliviour (approved):
Thank you for this change, h100s in XPK is awesome

RoshaniN:
Nina was able to check that Pathways functionality still work after these changes! Thank you for the change!

RoshaniN (approved):
Thanks Nina for verifying Pathways functionality!

PR #92, opened 2024-03-20
Update xpk.py

sadikneipp:
Remove server argument from proxy server arguments.

## Fixes / Features
- Without this change, the proxy server was not starting up for me. I'd get an unknown argument error.

## Testing / Documentation
This unblocked me from running a workload.

### DIFF
--- a/xpk.py
+++ b/xpk.py
@@ -2559,7 +2559,6 @@ def get_pathways_proxy_args(args) -> str:
               - --pathways_ifrt_proxy_server_resource_manager={args.workload}-rm-0-0.{args.workload}:38677
               - --pathways_ifrt_proxy_server_port=38676
               - --pathways_tmp_dir_pattern={args.pathways_gcs_location}
-              - --pathways_xprof_trace_enable_bulk_upload=true
               - --pathways_plaque_network=gcp"""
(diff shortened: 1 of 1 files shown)


Obliviour (commented):
Thanks @sadikneipp for the fast change!

1. Nit to change the PR Name to be more specific. 

2. @RoshaniN can you confirm if this flag is needed for other aspects -- is this the right fix for the issue?

RoshaniN:
Hey Sadi,
I am curious to know if any logic around this argument was recently changed in Pathways, as I did not face this issue.

Obliviour:
Looking at the Build Test failure, you may need to create a PR from a branch instead of a fork. I provided you permissions to do this, just in case.

RoshaniN (changes requested):
Thanks for the change, Sadi!

Looks like the flag fails after adding cl/607802178, which will be merged soon. Currently this flag is not super useful either, so I see no issues removing it.

 Changes are GTG once the branch is rebased onto main .

sadikneipp:
Since it's such a small change @RoshaniN aligned with me to incorporate this PR in #100 . I'll close this PR.

PR #97, opened 2024-03-25
debug dump gcs using gsutil -m

GallagherCommaJack:
should be much faster


Obliviour (approved):
Thanks Jack, this is a good find!

PR #100, opened 2024-03-27
Ensure proxy and server images are only provided with --use-pathways.

RoshaniN:
## Fixes / Features
- Remove default values for proxy-server and server images.
- Ensure user provides both proxy-server-image and server-image when --use-pathways is set and vice-versa.
- Validate that the proxy-server-image and server-image exist on GCR.
- Remove a Pathways flag.

## Testing / Documentation
Testing details.
Additionally, I did some manual tests to see that the user gets the right error message for - 
```
[XPK] --proxy-server-image and --server-image need to be provided with --use-pathways.
``` 
Also manually tested that invalid docker images are detected and user is guided with an error message.
```
[XPK] Failed to validate your docker image, check the docker image. You should be able to navigate to the URL gcr.io/hello_world in cloud-tpu-multipod-dev
[XPK] Please check the proxy-server-image or server-image as advised!
```
- [ y ] Tests pass

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,15 +125,6 @@ all zones.
     --num-slices=4 --spot
     ```
 
-* Cluster Create for Pathways:
-    Pathways compatible cluster can be created using `--enable-pathways`
-    ```shell
-    python3 xpk.py cluster create \
-    --cluster xpk-pw-test \
-    --num-slices=4 --on-demand \
-    --tpu-type=v5litepod-16 \
-    --enable-pathways
-    ```
  
 *   Cluster Create can be called again with the same `--cluster name` to modify
     the number of slices or retry failed steps.
@@ -209,36 +200,6 @@ all zones.
     --workload xpk-test-workload --command "echo goodbye" \
     --cluster xpk-test \
     --tpu-type=v5litepod-16
-    ```
-
-*   Workload Create for Pathways:
-    Pathways workload can be submitted using `--use-pathways` on a Pathways enabled cluster (created with `--enable-pathways`)
-    
-    Pathways workload example:
-    ```shell
-    python3 xpk.py workload create \
-    --workload xpk-pw-test \
-    --num-slices=1 \
-    --tpu-type=v5litepod-16 \
-    --use-pathways \
-    --cluster xpk-pw-test \
-    --docker-name='user-workload' \
-    --docker-image=<maxtext docker image> \
-    --command='bash /usr/pathways/ifrt/maxtext_entrypoint.sh base_output_directory=<output directory> dataset_path=<dataset path> per_device_batch_size=1 enable_checkpointing=false enable_profiler=false remat_policy=full global_parameter_scale=4 steps=300 max_target_length=2048 use_iota_embed=true reuse_example_batch=1 dataset_type=synthetic attention=flash gcs_metrics=True run_name=$(USER)-pw-xpk-test-1'
-    ```
-
-    Regular workload can also be submitted on a Pathways enabled cluster (created with `--enable-pathways`)
-    
-    Pathways workload example:
-    ```shell
-    python3 xpk.py workload create \
-    --workload xpk-regular-test \
-    --num-slices=1 \
-    --tpu-type=v5litepod-16 \
-    --cluster xpk-pw-test \
-    --docker-name='user-workload' \
-    --docker-image=<maxtext docker image> \
-    --command='python3 MaxText/train.py MaxText/configs/base.yml base_output_directory=<output directory> dataset_path=<dataset path> per_device_batch_size=1 enable_checkpointing=false enable_profiler=false remat_policy=full global_parameter_scale=4 steps=300 max_target_length=2048 use_iota_embed=true reuse_example_batch=1 dataset_type=synthetic attention=flash gcs_metrics=True run_name=$(USER)-pw-xpk-test-1'
     ```
 
 ### Set `max-restarts` for production jobs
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -354,6 +315,49 @@ checkpointing so the job restarts near where it was interrupted.
...
+    --workload xpk-pw-test \
+    --num-slices=1 \
+    --tpu-type=v5litepod-16 \
+    --cluster xpk-pw-test \
+    --use-pathways \
+    --server-image=<Pathways server image> \
+    --proxy-server-image=<Pathways proxy server image> \
+    --docker-name='user-workload' \
+    --docker-image=<maxtext docker image> \
+    --command='bash /usr/pathways/ifrt/maxtext_entrypoint.sh base_output_directory=<output directory> dataset_path=<dataset path> per_device_batch_size=1 enable_checkpointing=false enable_profiler=false remat_policy=full global_parameter_scale=4 steps=300 max_target_length=2048 use_iota_embed=true reuse_example_batch=1 dataset_type=synthetic attention=flash gcs_metrics=True run_name=$(USER)-pw-xpk-test-1'
```
Obliviour, line 345: Curious if you have a simpler example. This seems like a complicated README case example to maintain.

Like in the non-pathways workload command example we just print hello world. Is that possible here too?

README.md
```diff
@@ -354,6 +315,49 @@ checkpointing so the job restarts near where it was interrupted.
...
+    python3 xpk.py workload create \
+    --workload xpk-pw-test \
+    --num-slices=1 \
+    --tpu-type=v5litepod-16 \
+    --cluster xpk-pw-test \
+    --use-pathways \
+    --server-image=<Pathways server image> \
+    --proxy-server-image=<Pathways proxy server image> \
+    --docker-name='user-workload' \
+    --docker-image=<maxtext docker image> \
```
Obliviour, line 344: --base-docker-image flow is more recommended for xpk users so that local changes can be iterated on. Is that possible for pathways?

README.md
RoshaniN: Thanks for the question, Victor! Let me check if other dependencies are up-to-date. If so that I can simplify it.

xpk.py
```diff
@@ -3200,6 +3200,37 @@ def setup_docker_image(args) -> tuple[int, str]:
...
+  return args.use_pathways == (args.proxy_server_image is not None and args.server_image is not None)
+
+def validate_pathways_docker_images(args) -> bool:
+  """Validates the existence of Pathways server and proxy docker images in the project.
+
+  Args:
+      args (_type_): _description_
+
+  Returns:
+      bool: _description_
```
Obliviour, line 3223: nit to replace type / description

xpk.py
```diff
@@ -3200,6 +3200,37 @@ def setup_docker_image(args) -> tuple[int, str]:
...
+  """
+  # pylint: disable=line-too-long
```
Obliviour, line 3225: nit I don't think the line is too long

xpk.py
```diff
@@ -3200,6 +3200,37 @@ def setup_docker_image(args) -> tuple[int, str]:
...
+  proxy_server_return_code = validate_docker_image(args.proxy_server_image, args)
+  server_return_code = validate_docker_image(args.server_image, args)
+  if proxy_server_return_code>0 or server_return_code>0:
```
Obliviour, line 3228: nit spaces ` > 0`

README.md
RoshaniN: Currently, the docker-image provided by the users is a fixed docker image, but if the others deps (mentioned) are merged, I can update docker-image to base-docker-image! I would also prefer that!

xpk.py
```diff
@@ -3200,6 +3200,37 @@ def setup_docker_image(args) -> tuple[int, str]:
...
+  Both --proxy-server-image and --server-image need to be set if --use-pathways is set. 
+  Neither can be provided if --use-pathways is not provided.
+  Args:
+    args: user provided arguments for running the command.
+
+  Returns:
+    bool:
+      yaml for main and sidecar container
+  """
+  return args.use_pathways == (args.proxy_server_image is not None and args.server_image is not None)
```
Obliviour, line 3214: very nifty!

xpk.py
RoshaniN: Thanks for catching that!

xpk.py
RoshaniN: Thanks for catching that! It was long previously!

xpk.py
```diff
@@ -4857,17 +4896,15 @@ def directory_path_type(value):
 workload_pathways_workload_arguments.add_argument(
     '--proxy-server-image',
     type=str,
-    default='gcr.io/cloud-tpu-v2-images/pathways/pathways-demo:proxy_server',
```
Obliviour, line 4860: lmk if you do want to keep the ability to have a default and we can think through how this would look!

RoshaniN: I don't think it is super important, as folks would need to replace it during iteration!

xpk.py
RoshaniN: Thank you!

xpk.py
RoshaniN: Done, thanks!

README.md
Obliviour: > however, we really want the workload to actually train!

Agreed this is the ultimately goal but we should go one step at a time. How about we have both a simple workload and a MaxText example?

I want users to be able to verify they have xpk + system working before trying to train.

xpk.py
Obliviour: > during iteration

Can you go into this more? will customers need to replace this often? curious if we should make it simpler if so.

RoshaniN:
Closing this PR as an alternative approach has been implemented with "create-pathways".

PR #102, opened 2024-03-28
XPK update for hybridsim

tonyjohnchen:
## Fixes / Features
For `AOT+Hybridsim` integration, we need to pull Hybridsim docker image, so we need to mount `/var/run/docker.sock` (https://screenshot.googleplex.com/BD3SMwL57tP5cQY)
For running Hybridsim, we need to mount `/tmp/xla_dump/` so the Hybridsim docker image can access to HLO graph (https://screenshot.googleplex.com/64Fwrjwt6g55orn)

## Testing / Documentation
Tested `AOT+Hybridsim` e2e on a GKE node: https://screenshot.googleplex.com/5LQwBNBp4p6iyLq

This should be a combo PR of Maxtext and XPK:
https://github.com/google/maxtext/pull/564
https://github.com/google/xpk/pull/102

- [ y ] Tests pass

### DIFF
--- a/xpk.py
+++ b/xpk.py
@@ -123,10 +123,15 @@ spec:
                 volumeMounts:
                 - mountPath: /dev/shm
                   name: dshm-2
+                - mountPath: /var/run/docker.sock
+                  name: docker-socket-volume
               volumes:
               - emptyDir:
                   medium: Memory
                 name: dshm-2
+              - name: docker-socket-volume
+                hostPath:
+                  path: /var/run/docker.sock
 """
 
 gpu_workload_create_yaml = """apiVersion: jobset.x-k8s.io/v1alpha2
--- a/xpk.py
+++ b/xpk.py
@@ -125,6 +125,8 @@ spec:
                   name: dshm-2
                 - mountPath: /var/run/docker.sock
                   name: docker-socket-volume
+                - mountPath: /tmp/xla_dump/
+                  name: xla-dump-volume
               volumes:
               - emptyDir:
                   medium: Memory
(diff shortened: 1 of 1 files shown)


xpk.py
```diff
@@ -123,10 +123,20 @@
                 volumeMounts:
                 - mountPath: /dev/shm
                   name: dshm-2
+                - mountPath: /var/run/docker.sock
```
rwitten, line 126: We need to talk about the architecture, this doesn't make sense to me to hard code.

rwitten (changes requested):

xpk.py
Obliviour: +1 to avoiding hardcoding this in the common path for TPUs/ CPUs.

tonyjohnchen: It seems like both GPU and Pathway mount `/tmp` folder

https://github.com/google/xpk/blob/main/xpk.py#L325-L327
https://github.com/google/xpk/blob/main/xpk.py#L271-L272

Maybe I should mount `/tmp` folder instead of `/tmp/xla_dump/`?

Obliviour: GPU / Pathways has specific workload create files (also a sad state of duplication) but that's why it is ok for those mounts. The issue here is we are mounting something that doesn't always exist.

Obliviour:
I assume this PR is not needed anymore? @tonyjohnchen let me know if we can close it!

tonyjohnchen:
> I assume this PR is not needed anymore? @tonyjohnchen let me know if we can close it!

Yeah, I can close it.

PR #104, opened 2024-03-28
Support --env flag and Artifact Registry image validation

jonb377:
## Fixes / Features
- Individual environment variables can be set with `--env`
- Validate existence of ArtifactRegistry images before workload creation

## Testing / Documentation
Ran a few validation workloads:
- Nonexistent AR image will fail validation: https://paste.googleplex.com/6298050852552704#l=1
- Existing AR image will succeed: https://paste.googleplex.com/5135260269936640#l=18
- `--env` format is enforced: https://paste.googleplex.com/5226635967594496
- Specified variables are applied to the container: https://paste.googleplex.com/6198367050268672
- Multiple `--env` may be specified on a single workload: https://paste.googleplex.com/6259857570660352
- `--env` and `--env-file` are mutually exclusive: https://paste.googleplex.com/5719129935314944

- [ y ] Tests pass

### DIFF
--- a/xpk.py
+++ b/xpk.py
@@ -1252,13 +1252,13 @@ def add_env_config(args):
   device_type = args.tpu_type if args.tpu_type else args.device_type
   env = {'JOBSET_NAME': args.workload}
 
+  env_pat = re.compile(r'(^[a-zA-Z_][a-zA-Z0-9_]*?)(?:=(.*))?$', re.M)
   if args.env_file:
     print('Setting container environment from', args.env_file)
-    pat = re.compile(r'(^[a-zA-Z_][a-zA-Z0-9_]*?)(?:=(.*))$', re.M)
     with open(file=args.env_file, mode='r', encoding='utf-8') as f:
-      for match in pat.finditer(f.read()):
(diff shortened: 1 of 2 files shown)


xpk.py
```diff
@@ -4795,7 +4804,8 @@ def directory_path_type(value):
     default=1,
     help='The number of nodes to use, default=1.',
 )
-workload_create_parser_optional_arguments.add_argument(
+workload_env_arguments = workload_create_parser_optional_arguments.add_mutually_exclusive_group()
```
NinaCai, line 4807: Do they need to be exclusive?

jonb377: Not necessarily, we can relax this restriction if it would be useful. The idea was just to keep the two flows separate, either you specify the env from a file or manually with `--env`. cc @Obliviour

Obliviour: I think from a user perspective, it would be confusing to support both at the same time and handle conflicts between the same variables from both.

LMK if you think otherwise

Obliviour (approved):

NinaCai (approved):

PR #106, opened 2024-03-29
Enable Autoprovisioning Support in XPK

Obliviour:
Autoprovisioning allows clusters to be dynamically reprovisioned when based on new workload resource needs.

Add support for autoprovisioning in xpk with a new flag --enable-autoprovisioning, optional flags, and internal logic to make it easy to use in xpk.

## Fixes / Features
  <DONE> AI: service account support
  <DONE> AI: TPU Resource support
  <DONE> AI: GPU Resource support
  <DONE> AI: Cluster Update vs Cluster Create commands
  <DONE> AI: Disable support for NAP with XPK on CPUs. No use case AFAIK.
  <DONE> AI: workload file needs to support on-demand/reservation/spot
  <DONE> AI: Kueue Resource Flavor Support. (Ignore topology, set the resource amount to max_chips)
  <DONE> AI: Config map for the type needs to be changed to "autoprovision" and for gke_accelerator type.
  <DONE> AI: Add to config map what type of capacity type is used in the cluster.
  <DONE> AI: Set created accelerator node pools to be enabled with NAP. Not the default node pool.
  <DONE> AI: Bring default cpu np back into the cluster creation.
  <DONE> AI: Prevent workload from getting created if it above MAXIMUM chips the cluster supports? yes.

## Testing / Documentation

Created a v4 nap cluster.
Ran jobs of various sizes.
Job too large was rejected early.
Increased the cluster provisioning chips.

- [ y ] Tests pass

### DIFF
--- a/xpk.py
+++ b/xpk.py
@@ -33,6 +33,7 @@ Next Steps:
 
 import argparse
 import datetime
+import enum
 import os
 import random
 import re
@@ -69,9 +70,18 @@ default_gke_version="1.29.1-gke.1589017"
 # This is the version for XPK PyPI package
 __version__ = "0.3.0"
 xpk_current_version = __version__
+
+h100_device_type = 'h100-80gb-8'
+
+_AUTOPROVISIONING_CONFIG_VALUE = 'AUTOPROVISION'
+_AUTOPROVISIONING_CONFIG_MINIMUM_KEY = 'minimum_chips'
+_AUTOPROVISIONING_CONFIG_MAXIMUM_KEY = 'maximum_chips'
+
+_CAPACITY_TYPE_CONFIG_KEY = 'capacity_type'
+_RESERVATION_CONFIG_KEY = 'reservation_id'
 _CLUSTER_QUEUE_NAME='cluster-queue'
 _LOCAL_QUEUE_NAME='multislice-queue'
-h100_device_type = 'h100-80gb-8'
+_DEFAULT_POOL_NAME='default-pool'
 _CLUSTER_RESOURCES_CONFIGMAP = 'resources-configmap'
 _CLUSTER_METADATA_CONFIGMAP = 'metadata-configmap'
 _XPK_SERVICE_ACCOUNT = 'xpk-sa'
@@ -82,6 +92,13 @@ _DEFAULT_VERTEX_TENSORBOARD_NAME = 'tb-instance'
 
 if _VERTEX_TENSORBOARD_FEATURE_FLAG:
   from cloud_accelerator_diagnostics import tensorboard
+
+
+class CapacityType(enum.Enum):
+  ON_DEMAND='on_demand'
+  RESERVATION='reservation'
+  SPOT='spot'
+  UNKNOWN='unknown'
 
 
 workload_create_yaml = """apiVersion: jobset.x-k8s.io/v1alpha2
@@ -115,6 +132,7 @@ spec:
               nodeSelector:
                 {accelerator_label}
                 {machine_label}
+                {autoprovisioning_args}
               priorityClassName: {args.priority}
               hostNetwork: true
               dnsPolicy: ClusterFirstWithHostNet
@@ -169,6 +187,7 @@ spec:
               nodeSelector:
                 {accelerator_label}
                 {machine_label}
+                {autoprovisioning_args}
               priorityClassName: {args.priority}
               hostNetwork: true
               dnsPolicy: ClusterFirstWithHostNet
@@ -330,6 +349,7 @@ spec:
             nodeSelector:
               {accelerator_label}
               {machine_label}
+              {autoprovisioning_args}
             priorityClassName: {args.priority}
             volumes:
             - hostPath:
@@ -434,6 +454,7 @@ spec:
                 type: DirectoryOrCreate
               name: shared-tmp
 """
+
 script_dir_dockerfile = """FROM {base_docker_image}
 
 # Set the working directory in the container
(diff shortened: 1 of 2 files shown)


xpk.py
```diff
@@ -1877,23 +2130,176 @@ def create_cluster_network_config(args) -> int:
...
+    case CapacityType.ON_DEMAND.name:
+      node_selector = ""
+    case CapacityType.SPOT.name:
+      node_selector = 'cloud.google.com/gke-spot=\"true\"'
+    case CapacityType.RESERVATION.name:
+      node_selector = (
+        f'cloud.google.com/reservation-name: {args.reservation}'
+      )
+    case _:
+      xpk_print(f'Unknown capacity type: {capacity_type}. Unable to determine capacity args.')
```
SurbhiJainUSC, line 2275: nit: `unable to determine the node selectors`

xpk.py
```diff
@@ -2511,29 +2846,34 @@ def install_kueue_on_cluster(args) -> int:
...
 
 
-def enable_kueue_crds(args, system) -> int:
+def enable_kueue_crds(args, system, autoprovisioning_config: AutoprovisioningConfig | None) -> int:
   """Enable Kueue crds.
 
   Args:
     args: user provided arguments for running the command.
     system: system level arguments.
+    maximum_chips: the custom max chip allocation specified for kueue.
```
SurbhiJainUSC, line 2855: nit: `autoprovisioning_config`

xpk.py
```diff
@@ -1666,6 +1741,182 @@ def add_roles_to_service_account(args) -> int:
...
+  """
+  memory_limits = """
+  minimum: 1
+  maximum: 10000
+  """
+
+  # By default, the maximum chips is set to be the current number of resources used
+  # in the cluster. The minimum is set to zero.
+  if system.accelerator_type == AcceleratorType['GPU']:
+    num_chips_in_cluster = system.vms_per_slice * system.chips_per_vm * args.num_nodes
```
SurbhiJainUSC, line 1771: Why is this calculation different from line 2868?

xpk.py
```diff
@@ -3169,27 +3519,55 @@ def check_if_workload_can_schedule(args, system):
...
+    missing_gke_accelerator_type = True
+  elif cluster_config_map[system.gke_accelerator] == _AUTOPROVISIONING_CONFIG_VALUE:
+    # Run total chip check when in autoprovisioning mode.
+    max_chips_in_cluster = int(cluster_config_map[_AUTOPROVISIONING_CONFIG_MAXIMUM_KEY])
+    if system.accelerator_type == AcceleratorType['GPU']:
+      num_chips_in_workload = system.vms_per_slice * system.chips_per_vm * args.num_nodes
+    else:
+      num_chips_in_workload = system.vms_per_slice * system.chips_per_vm * args.num_slices
+
+    if max_chips_in_cluster > num_chips_in_workload:
```
SurbhiJainUSC, line 3536: Why can't we schedule the workload requesting `num_chips_in_workload` which is less than equal to `max_chips_in_cluster`?

xpk.py
```diff
@@ -3745,6 +4140,98 @@ def get_system_characteristics(args) -> tuple[SystemCharacteristics|None, int]:
...
+  # If the user doesn't specify args, then use the cluster settings.
+  capacity_type, return_code = get_capacity_type(args)
+  capacity_type_str = capacity_type.name
+  if return_code != 0:
+    xpk_print('Unable to get capacity type.')
+    return node_selector_args, return_code
+
+  if capacity_type_str == CapacityType.UNKNOWN.name:
+    # Use default settings from cluster creation.
+    resources_configmap_name = f'{args.cluster}-{_CLUSTER_METADATA_CONFIGMAP}'
```
SurbhiJainUSC, line 4200: nit: `metadata_configmap_name` to avoid confusion

xpk.py
```diff
@@ -1122,12 +1175,33 @@ def make_tmp_files(per_command_name):
...
   """
+  # Supports removal of spaces from command names before converting to file name.
   return [
-      tempfile.NamedTemporaryFile(delete=False, prefix=command + '-')
+      tempfile.NamedTemporaryFile(delete=False, prefix=command.replace(" ", "-") + '-')
       for command in per_command_name
   ]
 
 
+def get_value_from_map(key: str, map_to_search: dict) -> tuple[int, str | None]:
```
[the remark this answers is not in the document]
Obliviour: Cool I moved to use this instead of if key ... in logic.

I like this function since it pulls the common print statement out of each request to get_value from a map.

xpk.py
Obliviour: This should be the same, created a new helper function to reduce the dup code: determine_total_chips_from_args

xpk.py
```diff
@@ -1666,6 +1741,182 @@ def add_roles_to_service_account(args) -> int:
...
+  memory_limits = """
+  minimum: 1
+  maximum: 10000
+  """
+
+  # By default, the maximum chips is set to be the current number of resources used
+  # in the cluster. The minimum is set to zero.
+  if system.accelerator_type == AcceleratorType['GPU']:
+    num_chips_in_cluster = system.vms_per_slice * system.chips_per_vm * args.num_nodes
+  else:
```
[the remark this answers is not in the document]
Obliviour: Moved this logic to a helper function and I think it is possible the helper function could be used for CPUs thought it doesn't execute for CPUs currently. 

Do you think we don't need the comment anymore?

xpk.py
Obliviour: Thank you Surbhi!!

xpk.py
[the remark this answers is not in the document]
Obliviour: Oh actually, the helper function is used in the CPU case now since it is used by enable_kueue_crds.

xpk.py
Obliviour: Oh no I flipped the if statement and havent tested this code yet. Fixed it. Thanks for the catch!

xpk.py
Obliviour: agreed

SurbhiJainUSC (approved):

PR #112, opened 2024-04-03
return pid exit code if it is non-zero

NinaCai:
## Fixes / Features
- gpu xpk returns non-zero code when any pid returns non-zero 

## Testing / Documentation
Testing details.

- [ y ] Tests pass

### DIFF
--- a/xpk.py
+++ b/xpk.py
@@ -263,7 +263,7 @@ spec:
                   - "bash"
                   - "-c"
                   - |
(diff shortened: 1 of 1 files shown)


xpk.py
yangyuwei, line 1: Thanks @NinaCai for the fix!

yangyuwei (approved):

Obliviour (approved):

PR #115, opened 2024-04-04
Nina/unify gpu container yaml

NinaCai:
## Fixes / Features
-
-

## Testing / Documentation
Testing details.

- [ y ] Tests pass

### DIFF
--- a/xpk.py
+++ b/xpk.py
@@ -224,60 +224,7 @@ spec:
                 env:
                 - name: LD_LIBRARY_PATH
                   value: /usr/local/nvidia/lib64
-              - name: maxtext-tcpx
-                image: {docker_image}
-                imagePullPolicy: Always
-                securityContext:
-                  privileged: true
-                ports:
-                    - containerPort: 6002
-                env:
-                  - name: REPLICATED_JOB_NAME
-                    valueFrom:
-                      fieldRef:
-                        fieldPath: metadata.annotations['jobset.sigs.k8s.io/replicatedjob-name']
-                  - name: JOBSET_NAME
-                    valueFrom:
-                      fieldRef:
-                        fieldPath: metadata.annotations['jobset.sigs.k8s.io/jobset-name']
-                  - name: JAX_COORDINATOR_ADDRESS
-                    value: "$(JOBSET_NAME)-$(REPLICATED_JOB_NAME)-0-0.$(JOBSET_NAME)"
-                  - name: NNODES
-                    value: "{args.num_nodes}"
-                  - name: NODE_RANK
-                    valueFrom:
-                      fieldRef:
-                        fieldPath: metadata.annotations['batch.kubernetes.io/job-completion-index']
-                  - name: USE_GPUDIRECT
-                    value: "tcpx"
-                  - name: GPUS_PER_NODE
-                    value: "{chips_per_vm}"
-                  - name: JAX_COORDINATOR_PORT
-                    value: "6002"
-                  - name: LD_LIBRARY_PATH
-                    value: /usr/local/nvidia/lib64
-                  - name: COMMAND
-                    value: "{command}"
-                  {args.env}
-                command:
-                  - "bash"
-                  - "-c"
-                  - |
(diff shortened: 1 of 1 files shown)


xpk.py
```diff
@@ -3373,14 +3320,20 @@ def get_main_container(args, system, docker_image, resource_type) -> str:
   if args.enable_debug_logs:
     command = ('TPU_STDERR_LOG_LEVEL=0 TPU_MIN_LOG_LEVEL=0 TF_CPP_MIN_LOG_LEVEL=0'
                f' TPU_VMODULE=real_program_continuator=1 {args.command}')
+  
+  device_type = args.tpu_type if args.tpu_type else args.device_type
```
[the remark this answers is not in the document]
Obliviour: but it seems like we can use system.accelerator_type = GPU and that would be better?

xpk.py
```diff
@@ -3374,13 +3321,19 @@ def get_main_container(args, system, docker_image, resource_type) -> str:
     command = ('TPU_STDERR_LOG_LEVEL=0 TPU_MIN_LOG_LEVEL=0 TF_CPP_MIN_LOG_LEVEL=0'
                f' TPU_VMODULE=real_program_continuator=1 {args.command}')
 
+  device_type = args.tpu_type if args.tpu_type else args.device_type
+  gpu_workload_terminate_command = ''
+  if device_type == h100_device_type:
+    command = ('cd /deps && bash gpu_multi_process_run.sh')
+    gpu_workload_terminate_command = ('echo Main app is done > /usr/share/workload/workload_terminated; ')
```
Obliviour, line 3328: nit, I think we don't need the () around the strings

xpk.py
```diff
@@ -3426,23 +3381,55 @@ def add_image_pull_policy_for_pw(args):
...
+                  mountPath: /usr/local/nvidia/lib64
+                - name: tcpx-nccl-plugin-volume
+                  mountPath: /usr/local/tcpx
+                - name: tcpd-socket
+                  mountPath: /tmp
+                - name: shared-memory
+                  mountPath: /dev/shm
+                - name: workload-terminated-volume
+                  mountPath: /usr/share/workload"""
+  device_type = args.tpu_type if args.tpu_type else args.device_type
```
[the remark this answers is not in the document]
Obliviour: I am currently assuming that the workload file for GPU has no a100 specific code in it but let me know if I assumed wrong!

Obliviour (commented):
Looking forward to this simplification. Thank you Nina for making this!

xpk.py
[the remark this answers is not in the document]
NinaCai: Right, we could use `system.accelerator_type`

xpk.py
```diff
@@ -3449,23 +3403,55 @@ def add_image_pull_policy_for_pw(args):
...
+  """ Docker name for the main container.
+  Args:
+    args: user provided args.
+    system: system characteristics.
+
+  Returns:
+    str:
+      Workload docker image as a YAML string
+  """
+  # Resources requirements for Pathways workload containers are known.
```
Obliviour, line 3421: comment `  # Resources requirements for Pathways workload containers are known.` not needed

xpk.py
```diff
@@ -3592,6 +3616,11 @@ def add_container_ports(args) -> str:
                 - containerPort: 8080"""
   if args.use_pathways:
     return ''
+
+  gpu_port_yaml = """- containerPort: 6002"""
+  device_type = args.tpu_type if args.tpu_type else args.device_type
```
Obliviour, line 3621: Should this be `if system.accelerator_type == AcceleratorType['GPU']:` as well?

xpk.py
```diff
@@ -3449,23 +3403,55 @@ def add_image_pull_policy_for_pw(args):
...
+    system: system characteristics.
+
+  Returns:
+    str:
+      Workload docker image as a YAML string
+  """
+  # Resources requirements for Pathways workload containers are known.
+
+  if system.accelerator_type == AcceleratorType['GPU']:
+    return "maxtext-tcpx"
```
Obliviour, line 3424: Hmm I think the name is just an identifier for the docker image. We can set this to `gpu-image` instead of `maxtext-tcpx` which might not be correct if the user is not running maxtext.

---

Outside of this PR, xpk should support the args.docker_image arg but set a reasonable default for the gpu / cpu /tpu cases. Right now the default is `jax-tpu` which is not reasonable default for gpus or cpus. I can look at this later.

Obliviour (commented):
Looks good to me other than some small nits!

Obliviour (approved):

PR #120, opened 2024-04-08
Check cluster arguments and update nodepools in existing cluster when requesting different device_type

SurbhiJainUSC:
## Fixes / Features
- Do not update nodepools in existing cluster when `args.zone` is different from the zone in which the cluster was initially created 
- Delete nodepools in existing cluster and create new nodepools when requesting cluster to update with different device_type

## Testing / Documentation
*Scenario 1:*
Cluster already has 4 nodepools of v4-8 in us-central2-b and now we request 2 nodepools of v4-8 in us-central2-b. The end state of the cluster will be 2 nodepools of v4-8 in us-central2-b.

*Scneraio 2:*
Cluster has 2 nodepools of v4-8 in us-central2-b and now we request 4 nodepools of v4-8 in us-central2-b. The end state of the cluster will be  4 nodepools of v4-8 in us-central2-b.

*Scenario 3:*
Cluster already has 2 nodepools of v4-8 in us-central2-b and now we request 2 nodepools of v4-16 in us-central2-b. The end state of the cluster will be 2 nodepools of v4-16 in us-central2-b.

*Scenario 4:*
Cluster already has 2 nodepools of v4-8 in us-central2-b and now we request 3 nodepools of v4-16 in us-central2-b. The end state of the cluster will be 3 nodepools of v4-16 in us-central2-b.

*Scenario 5:*
Cluster already has 2 nodepools of v4-8 in us-central2-b and now we request 2 nodepools of v4-8 in us-central2-a. XPK will fail early and will not update the cluster. The end state of the cluster will be 2 nodepools of v4-8 in us-central2-b.

- [ y ] Tests pass

### DIFF
--- a/xpk.py
+++ b/xpk.py
@@ -2506,6 +2506,51 @@ def create_cluster_if_necessary(args) -> int:
     return run_gke_cluster_create_command(args)
 
 
+def get_nodepool_zone(args, nodepool_name) -> tuple[int, str]:
+  """Return zone in which nodepool exists in the cluster.
+
+  Args:
+    args: user provided arguments for running the command.
+    nodepool_name: name of nodepool.
+
+  Returns:
+    0 if successful, 1 otherwise and the zone of nodepool.
+  """
+  command = (
+      f'gcloud beta container node-pools describe {nodepool_name}'
+      f' --cluster {args.cluster} --project={args.project}'
+      f' --region={zone_to_region(args.zone)} --format="value(locations)"'
+  )
+  return_code, nodepool_zone = (
+      run_command_for_value(command, 'Get Node Pool Zone', args)
+  )
+  if return_code != 0:
+    xpk_print(f'Get Node Pool Zone returned ERROR {return_code}')
+    return 1, None
+
+  return 0, nodepool_zone.strip()
+
+
+def check_device_type_in_cluster(args, device_type) -> bool:
+  """Check if cluster has resources of a specified device_type.
+
+  Args:
+    args: user provided arguments for running the command.
+    device_type: device_type to check.
+
+  Returns:
+    True if device_type exists in the cluster, False otherwise.
+  """
+  resources_configmap_name = f'{args.cluster}-{_CLUSTER_RESOURCES_CONFIGMAP}'
+  resources_config_map = get_cluster_configmap(args, resources_configmap_name)
+  if resources_config_map is None:
+    xpk_print(f'No ConfigMap exist for cluster with the name {resources_config_map}.')
+    return False
+  if device_type in resources_config_map:
+    return True
+  return False
+
+
 def get_all_nodepools_programmatic(args) -> tuple[list[str], int]:
   """Gets all the nodepools associated with the cluster / project / region.
 
@@ -2528,6 +2573,9 @@ def get_all_nodepools_programmatic(args) -> tuple[list[str], int]:
     return [], 1
 
   all_nodepools = [x.split(' ')[0] for x in raw_nodepool_output.splitlines()]
+  # remove header=NAME from the nodepools list
+  if 'NAME' in all_nodepools:
+    all_nodepools.remove('NAME')
   return all_nodepools, 0
 
 
@@ -2612,6 +2660,38 @@ def get_user_input(input_msg):
   return user_input in ('y', 'yes')
 
 
+def get_node_pools_to_delete(args, existing_node_pool_names, desired_node_pool_names) -> list:
+  """Get list of nodepools to delete from the cluster.
+
+  Args:
+    args: user provided arguments for running the command.
+    existing_node_pool_names: names of nodepools that already exist in the cluster.
+    desired_node_pool_names: names of nodepools that should exist in the cluster.
+
+  Returns:
+    List of nodepool names to delete.
+  """
+  node_pools_to_delete = []
+  desired_node_pool_type = args.tpu_type if args.tpu_type else args.device_type
+  is_requested_device_type_in_cluster = check_device_type_in_cluster(args, desired_node_pool_type)
+  for existing_node_pool_name in existing_node_pool_names:
+    # Deletion logic would leave behind any Pathways CPU nodepools.
+    if existing_node_pool_name.find(f'{args.cluster}-np-') != 0:
+      continue
+
+    # Nodepools will be deleted in two scenarios:
+    # Scenario 1: Cluster exists with 3 nodepools of 'x' device_type and now we are updating
+    # the cluster to 2 nodepools of 'x' device_type. In this case, we will delete
+    # '{args.cluster}-np-2' from the cluster.
+    # Scenario 2: Cluster exists with 2 nodepools of 'x' device_type and now we are updating
+    # the cluster to 2 nodepools of 'y' device_type. In this case, we will delete
+    # '{args.cluster}-np-0' and '{args.cluster}-np-1' from the cluster.
+    if existing_node_pool_name not in desired_node_pool_names or not is_requested_device_type_in_cluster:
+      node_pools_to_delete.append(existing_node_pool_name)
+
+  return node_pools_to_delete
+
+
 def run_gke_node_pool_create_command(args, system) -> int:
   """Run the Create GKE Node Pool request.
 
@@ -2651,9 +2731,6 @@ def run_gke_node_pool_create_command(args, system) -> int:
     xpk_print('Parsing capacity arguments failed!')
     return return_code
 
-  commands = []
-  task_names = []
-
   if system.accelerator_type == AcceleratorType['GPU']:
     xpk_print(
       f'Creating 1 node pool with {args.num_nodes} nodes of {system.device_type}\n'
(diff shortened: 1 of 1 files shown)


xpk.py
```diff
@@ -2506,6 +2506,51 @@ def create_cluster_if_necessary(args) -> int:
...
 
+def get_nodepool_zone(args, nodepool_name) -> tuple[int, str]:
+  """Return zone in which nodepool exists in the cluster.
+
+  Args:
+    args: user provided arguments for running the command.
+    nodepool_name: name of nodepool.
+
+  Returns:
+    0 if successful, 1 otherwise and the zone of nodepool.
```
Obliviour, line 2517: nit that the function also returns a string

xpk.py
```diff
@@ -2612,6 +2660,38 @@ def get_user_input(input_msg):
...
+    if existing_node_pool_name.find(f'{args.cluster}-np-') != 0:
+      continue
+
+    # Nodepools will be deleted in two scenarios:
+    # Scenario 1: Cluster exists with 3 nodepools of 'x' device_type and now we are updating
+    # the cluster to 2 nodepools of 'x' device_type. In this case, we will delete
+    # '{args.cluster}-np-2' from the cluster.
+    # Scenario 2: Cluster exists with 2 nodepools of 'x' device_type and now we are updating
+    # the cluster to 2 nodepools of 'y' device_type. In this case, we will delete
+    # '{args.cluster}-np-0' and '{args.cluster}-np-1' from the cluster.
```
Obliviour, line 2688: Can you add in the readme an example of this scenario in the section about recalling cluster create. 

Basically showing: 

How to change your cluster's device_type:

Call xpk clsuter create with a new --device-type=v5-8. The cluster will delete the old node pools.

xpk.py
```diff
@@ -2669,8 +2746,36 @@ def run_gke_node_pool_create_command(args, system) -> int:
       f'{args.cluster}-np-{slice_num}' for slice_num in range(args.num_slices)
     ]
 
+  node_pools_to_remain = []
+  delete_commands = []
+  delete_task_names = []
+  if existing_node_pool_names:
+    return_code, existing_node_pool_zone = get_nodepool_zone(args, existing_node_pool_names[0])
+    if return_code == 0:
```
Obliviour, line 2754: what if return_code != 0?

xpk.py
```diff
@@ -2669,8 +2746,36 @@ def run_gke_node_pool_create_command(args, system) -> int:
...
+      if existing_node_pool_zone and existing_node_pool_zone != args.zone:
+        xpk_print(f'Cluster already has nodepools in zone: {existing_node_pool_zone}.'
+                  ' Use the same zone to update nodepools in the cluster.')
```
Obliviour, line 2757: nit add {args.cluster} name

xpk.py
```diff
@@ -2735,17 +2840,8 @@ def run_gke_node_pool_create_command(args, system) -> int:
...
-  for existing_node_pool_name in existing_node_pool_names:
-    if (
-        existing_node_pool_name.find(f'{args.cluster}-np-') == 0
-        and existing_node_pool_name not in desired_node_pool_names
-    ):
-      node_pools_to_delete.append(existing_node_pool_name)
+      create_commands.append(command)
+      create_task_names.append(task)
 
   will_delete = True
```
Obliviour, line 2846: nit we can move this delete user check earlier to directly after we figure the list of node_pools_to_delete. That would save time of processing the node pool commands if the user selects no.

xpk.py
```diff
@@ -2754,30 +2850,29 @@ def run_gke_node_pool_create_command(args, system) -> int:
...
+  # '{args.cluster}-np-i' nodepool will be re-created for 'y' device_type.
+  for i, command in enumerate(delete_commands):
+    xpk_print(f'To complete {delete_task_names[i]} we are executing {command}')
+  max_return_code = run_commands(
+      delete_commands, 'Delete Nodepools', delete_task_names, dry_run=args.dry_run
+  )
+  if max_return_code != 0:
+    xpk_print(f'Delete Nodepools returned ERROR {max_return_code}')
+    return 1
+
```
Obliviour, line 2868: I think we should update the config map after the deletion is successful with the new device_type.

There are a few scenarios that the node pool commands could succeed / fail, and this would help avoid unnecessary deletion of node pools.

1. User creates v4-8 cluster
2. User changes cluster to v5e-8

Outcome 1:
Command fails to delete, this is ok since the type of the cluster is still v4-8

Outcome 2:
Command succeeds to delete, but fails to create new. This is ok since there are 0 node pools so the type doesnt matter.

Outcome 3:
Command succeeds to delete and partially fails to create new. This is not ok since the tpu type doesnt align with what is in the clsuter. This can cause following calls to delete the node pools and retry creation from 0. We can avoid this by setting the device_type of the cluster after delete. 

What do you think?

Obliviour (commented):
Looks great, I think being able to change the cluster's device type will be really useful. I found one edge case that we can fix.

xpk.py
SurbhiJainUSC: I have mentioned `zone of nodepool`. I will add this as a new line for readability.

xpk.py
SurbhiJainUSC: Should we exit in that case?

xpk.py
SurbhiJainUSC: Yes, that sounds good.

xpk.py
Obliviour: Yes sounds good

xpk.py
SurbhiJainUSC: Updated README.

xpk.py
SurbhiJainUSC: Updated.

xpk.py
SurbhiJainUSC: I am updating the ConfigMap after deletion is successful. 
So, the different scenarios look like this when user updated cluster from 2 slices of v4-8 to 2 slices of v5e-8:
ConfigMap before the update is called: `v4-8: 2`

*Scenario 1: Deletion of v4-8 is unsuccessful*
Outcome: ConfigMap will be `v4-8: 2`.

*Scenario 2: Deletion of v4-8 is successful but creation of v5e-8 fails*
Outcome: ConfigMap will be updated to `v5e-8: 0`.

*Scenario 3: Deletion of v4-8 is successful but creation of v5e-8 partially fails*
Outcome: ConfigMap will be updated to `v5e-8: 0`. The following call to `create cluster` will not delete the v5e-8 that were already created.

xpk.py
```diff
@@ -3408,8 +3506,97 @@ def run_gke_node_pool_create_command(
...
+          f'Planning to delete {len(node_pools_to_delete)} node pools including'
+          f' {node_pools_to_delete}. \nDo you wish to delete: y (yes) / n'
+          ' (no):\n'
+      )
+    if not will_delete:
+      xpk_print(
+          'You have requested to not delete the existing nodepools in the'
+          ' cluster. There will be no change to the cluster.'
+      )
+      xpk_exit(1)
```
Obliviour, line 3563: change to return 1 to allow parent function to exit to follow our current exit pattern. I saw a few other cases that can be changed as well.

xpk.py
```diff
@@ -3243,6 +3253,55 @@ def create_cluster_if_necessary(args, gke_control_plane_version: str) -> int:
...
+  Returns:
+    True if device_type exists in the cluster, False otherwise.
+  """
+  resources_configmap_name = f'{args.cluster}-{_CLUSTER_RESOURCES_CONFIGMAP}'
+  resources_config_map = get_cluster_configmap(args, resources_configmap_name)
+  if resources_config_map is None:
+    xpk_print(
+        f'No ConfigMap exist for cluster with the name {resources_config_map}.'
+    )
+    return False
```
Obliviour, line 3299: we should ignore the device check if the configmap doesn't exist to avoid breaking old clusters.

xpk.py
```diff
@@ -3243,6 +3253,55 @@ def create_cluster_if_necessary(args, gke_control_plane_version: str) -> int:
...
+  if device_type in resources_config_map:
```
Obliviour, line 3300: also return True if `enable_autoprovisioning == true`

Obliviour (commented):
Some nits and comments about supporting autoprovisioning (aka by skipping device checks when autoprovisioning is configured)

xpk.py
```diff
@@ -3346,6 +3408,45 @@ def get_user_input(input_msg):
...
+  Args:
+    args: user provided arguments for running the command.
+    existing_node_pool_names: names of nodepools that already exist in the cluster.
+    desired_node_pool_names: names of nodepools that should exist in the cluster.
+
+  Returns:
+    List of nodepool names to delete.
+  """
+  node_pools_to_delete = []
+  desired_node_pool_type = args.tpu_type if args.tpu_type else args.device_type
```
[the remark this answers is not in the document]
SurbhiJainUSC: Sure!

xpk.py
SurbhiJainUSC: Updated!

xpk.py
```diff
@@ -3408,8 +3506,97 @@ def run_gke_node_pool_create_command(
...
+        delete_commands,
+        'Delete Nodepools',
+        delete_task_names,
+        dry_run=args.dry_run,
+    )
+    if max_return_code != 0:
+      xpk_print(f'Delete Nodepools returned ERROR {max_return_code}')
+      return 1
+
+    # Update {args.cluster}-{_CLUSTER_RESOURCES_CONFIGMAP} ConfigMap to 'y': '0'
```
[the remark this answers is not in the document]
SurbhiJainUSC: Shouldn't we update and check resources ConfigMap when requested GKE accelerator type is different for clusters with autoprovisioning enabled?

xpk.py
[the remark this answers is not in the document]
SurbhiJainUSC: Done!

xpk.py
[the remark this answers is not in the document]
SurbhiJainUSC: As discussed, updating the ConfigMap for autoprovisioning case.

xpk.py
SurbhiJainUSC: Done!

xpk.py
SurbhiJainUSC: Sure, updated!

xpk.py
SurbhiJainUSC: Fixed autoprovisioning checks.

Obliviour (approved):

PR #122, opened 2024-04-12
Update README with Vertex AI Tensorboard information and update pip version to 0.4.0

SurbhiJainUSC:
## Fixes / Features
- Add `cloud-accelerator-diagnostics` as a dependency in pyproject.toml
- Update README with Vertex AI Tensorboard functionality information
- Upgrade pip version to 0.4.0

## Testing / Documentation
Testing details.

- [ y ] Tests pass

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,13 @@ To install xpk, run the following command:
 pip install xpk
 ```
 
+If you are running XPK by cloning GitHub repository, first run the
+following command to begin using XPK commands:
+
+```shell
+pip install cloud-accelerator-diagnostics
+```
+
 # XPK for Large Scale (>1k VMs)
 
 Follow user instructions in [xpk-large-scale-guide.sh](xpk-large-scale-guide.sh)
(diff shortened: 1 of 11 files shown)


README.md
```diff
@@ -58,6 +58,13 @@ To install xpk, run the following command:
 pip install xpk
 ```
 
+If you are running XPK by cloning GitHub repository, first run the
+following command to begin using XPK commands:
+
+```shell
+pip install cloud-accelerator-diagnostics
```
Obliviour, line 65: We can suggest installing all dependencies in case we add more in the future.

```
git clone https://github.com/google/xpk.git
# Install dependencies such as cloud-accelerator-diagnostics
pip install .
```

Also want to add a section about:

```
If you see an error saying: `This environment is externally managed`. Please use a virtual environment:

Example:

```shell
  ## One time step of creating the venv
  VENV_DIR=~/venvp3
  python3 -m venv $VENV_DIR
  ## Enter your venv.
  source $VENV_DIR/bin/activate
  ## Same steps for cloning and installing dependencies.
```

Obliviour (approved):

Obliviour:
I think we should do two things:

1. Let people know that there is a dependencies install step that is needed to use vertex ai
2. Only import the tensorboard module when vertex ai is being used in the xpk script so that it doesn't break current workflows

Obliviour (approved):

PR #124, opened 2024-04-15
Add custom env variables to CPU workloads.

RoshaniN:
## Fixes / Features
- Allow env variables to be added to CPU workloads.
- Jobset name was repeated https://github.com/google/xpk/blob/main/xpk.py#L1302 , hence removed in CPU env config. 
- Renamed add_env_config to parse_env_config as the name is misleading with the task done by this function. parse_env_config also adds cpu env variables as needed.
- Reorganized volume mounts to which were already getting added to the main container.
-  Cluster delete has been flaky for the past few days, added 5 attempts to delete as a short-term fix.

## Testing / Documentation
Testing details.

- [ y ] Tests pass

### DIFF
--- a/xpk.py
+++ b/xpk.py
@@ -139,10 +139,6 @@ spec:
               terminationGracePeriodSeconds: {args.termination_grace_period_seconds}
               containers:
               {container}
-                {env}
-                volumeMounts:
-                - mountPath: /dev/shm
-                  name: dshm-2
               volumes:
               - emptyDir:
                   medium: Memory
@@ -1275,8 +1271,8 @@ def add_zone_and_project(args):
   xpk_print(f'Working on {args.project=} and {args.zone}')
 
 
-def add_env_config(args, tensorboard_config):
-  """Adds environment configurations to the jobset config.
+def parse_env_config(args, tensorboard_config):
+  """Parses the environment configurations to the jobset config.
 
   Args:
     args: user provided arguments for running the command.
@@ -2800,7 +2796,6 @@ def run_gke_cluster_delete_command(args) -> int:
   )
 
   return_code = run_command_with_updates(command, 'Cluster Delete', args)
-
   if return_code != 0:
     xpk_print(f'Cluster delete request returned ERROR {return_code}')
     return 1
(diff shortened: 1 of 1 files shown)


xpk.py
```diff
@@ -4218,10 +4215,12 @@ def get_cpu_env(num_slices, system) -> str:
                   value: "{processes_in_job}"
                 - name: JAX_PROCESS_COUNT
                   value: "{process_count}"
+                {env_vars}
```
jonb377, line 4218: I guess there's no easy way to do it, but it would be good to assert none of the above env vars are being overwritten here. @RoshaniN Did you hit an error when `JOBSET_NAME` was set twice?

jonb377 (approved):
Thanks for the quick turnaround Roshani!

xpk.py
RoshaniN: Fair observation!
No error, @jonb377 - the YAML generated by XPK showed the JOBSET_NAME (previously) mentioned here and also added by add_env_config . Also, overwriting would not error out.

xpk.py
```diff
@@ -4186,10 +4186,11 @@ def calculate_process_count(num_slices, vms_per_slice) -> str:
 
   return f"{num_processes}"
 
-def get_cpu_env(num_slices, system) -> str:
+def get_cpu_env(num_slices, env_vars, system) -> str:
```
Obliviour, line 4189: It looks like cpu flow has its own way of adding {env} env_vars. It attaches the variables directly to the xpk workload yaml. ref: https://github.com/google/xpk/blob/main/xpk.py#L4493

In gpu, pathways, tpu, they use the container {env} template to add these variables. https://github.com/google/xpk/blob/main/xpk.py#L3775

We condense the paths to 1 by merging function `get_cpu_env` into `get_env_container(args, system)` and taking out the {env} in the workload template. What do you think?

RoshaniN: Ohh get_env_container seems to be a better place, great feedback, thanks!

RoshaniN: Addressed the feedback, thanks!

Obliviour (approved):

PR #125, opened 2024-04-16
Add configuration setting for default pool num nodes

Obliviour:
## Fixes / Features
- Adds new configuration setting for num-nodes-cpu to change the initial nodes of the default cpu node pool
- This is important to avoid master resizing in our build / integ tests.

## Testing / Documentation
Testing details.

- [ y] Tests pass

### DIFF
--- a/.github/workflows/build_tests.yaml
+++ b/.github/workflows/build_tests.yaml
@@ -45,8 +45,8 @@ jobs:
         install_components: 'beta,gke-gcloud-auth-plugin'
     - name: Verify gcp setup
       run: gcloud info
-    - name: Create a Pathways-enabled XPK Cluster with 2x v4-8 nodepools.
-      run: python xpk.py cluster create --cluster $TPU_CLUSTER_NAME --enable-pathways  --device-type=v4-8  --num-slices=2 --zone=us-central2-b --default-pool-cpu-machine-type=n1-standard-16 --reservation='${{ secrets.GCP_TPU_V4_RESERVATION }}'
+    - name: Create a Pathways-enabled XPK Cluster with 2x v4-8 nodepools. Larger num-nodes to avoid master resizing.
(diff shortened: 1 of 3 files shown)


SurbhiJainUSC (approved):

RoshaniN (approved):
Thanks Victor!

.github/workflows/build_tests.yaml
```diff
@@ -45,8 +45,8 @@ jobs:
         install_components: 'beta,gke-gcloud-auth-plugin'
     - name: Verify gcp setup
       run: gcloud info
-    - name: Create a Pathways-enabled XPK Cluster with 2x v4-8 nodepools.
-      run: python xpk.py cluster create --cluster $TPU_CLUSTER_NAME --enable-pathways  --device-type=v4-8  --num-slices=2 --zone=us-central2-b --default-pool-cpu-machine-type=n1-standard-16 --reservation='${{ secrets.GCP_TPU_V4_RESERVATION }}'
+    - name: Create a Pathways-enabled XPK Cluster with 2x v4-8 nodepools. Larger num-nodes to avoid master resizing.
+      run: python xpk.py cluster create --cluster $TPU_CLUSTER_NAME --enable-pathways  --device-type=v4-8  --num-slices=2 --zone=us-central2-b --default-pool-cpu-machine-type=n1-standard-16 --default-pool-cpu-num-nodes=16 --reservation='${{ secrets.GCP_TPU_V4_RESERVATION }}'
```
RoshaniN, line 49: I think in the end state of the cluster, the default pool contains 2 nodes. The cluster seems to be sizing down. Do you think we should start with a smaller number than 16 (or even the default currently of 6)?

RoshaniN: https://screenshot.googleplex.com/5s5uDD7FtRa9rfB

Obliviour: 1. We need a larger starting size so that the gke master is scaled up to support more nodes without causing a resize request.

In this case we have a +5 node increase (through gcloud ... container node-pool create) after the cluster is created (gcloud  ... container clusters create). We need at least 8 nodes in the initial cluster creation to avoid a scale up.

2. Starting with a small default pool could have not enough system cpus to install kueue, jobset, and kubedns system pods. We noticed slow downs when the cpu node pool is small.

PR #127, opened 2024-04-17
Add Support for Pathways Expected Instances & Larger Default Worker Backoff Limit

SujeethJinesh:
## Fixes / Features
- Changes to support `pathways_expected_instances` (replaces `pathways_resource_manager_expected_num_worker_jobs`), which is more accurate as it includes topology information.
- Sets default backoff limit for pathways workers to 4.
- Added `terminationGracePeriodSeconds` to workers on Pathways. This allows the user to specify the timeframe workers may need to save state and gracefully exit if preempted. 

## Testing / Documentation
Validated manually with XPK create command below.

```
python3 xpk.py workload create --workload ${WORKLOAD_NAME} --cluster ${CLUSTER} --num-slices=${NUM_SLICES} --tpu-type=${TPU_TYPE} --use-pathways --project $PROJECT --zone $ZONE --docker-image="gcr.io/cloud-tpu-multipod-dev/${USER}_runner" --command="${COMMAND}" --server-image="gcr.io/cloud-tpu-multipod-dev/pathways-prototype/server/${USER}:server" --proxy-server-image="gcr.io/cloud-tpu-multipod-dev/pathways-prototype/gke/${USER}:server" --pathways-gcs-location="${LOCATION}" --termination-grace-period-seconds="${TERMINATION_GRACE_PERIOD}" --max-restarts=4
```

- [ y ] Tests pass

### DIFF
--- a/xpk.py
+++ b/xpk.py
@@ -1899,6 +1899,7 @@ PathwaysExpectedInstancesMap = {
     'v3': 'v3',
 }
 
+
 def chunks(lst, n):
   """Return a list of n-sized chunks from lst.
 
@@ -4846,27 +4847,34 @@ def get_pathways_rm_args(args, system: SystemCharacteristics) -> str:
               - --pathways_tmp_dir_pattern={args.pathways_gcs_location}
               - --pathways_expected_instances={expected_instances}"""
   if args.use_pathways:
-    return yaml.format(args=args, expected_instances=compute_pathways_expected_instances(args, system))
+    return yaml.format(
+        args=args,
+        expected_instances=compute_pathways_expected_instances(args, system),
+    )
   else:
     return ''
 
-def compute_pathways_expected_instances(args, system: SystemCharacteristics) -> str:
+
+def compute_pathways_expected_instances(
+    args, system: SystemCharacteristics
+) -> str:
   """Computes the expected instances from the system characteristics.
   Args:
     args: user provided args.
     system: system characteristics.
 
   Returns:
-    str: formatted string representing the expected instances (eg: 
+    str: formatted string representing the expected instances (eg:
     "tpuv4:2x2x2,tpuv4:2x2x2" for 2 slices of v4-16).
   """
   expected_instances = ','.join([
-      f'tpu{get_pathways_expected_tpu_type(system.device_type)}:{system.topology}' 
+      f'tpu{get_pathways_expected_tpu_type(system.device_type)}:{system.topology}'
       for _ in range(args.num_slices)
   ])
 
   xpk_print(f'Pathways expected instances are: {expected_instances}')
   return expected_instances
+
 
 def get_pathways_expected_tpu_type(device_type: str) -> str:
   """Returns the device type expected by Pathways
@@ -4879,9 +4887,13 @@ def get_pathways_expected_tpu_type(device_type: str) -> str:
   raw_type = device_type.split('-')[0].lower()
   pathways_expected_instance = PathwaysExpectedInstancesMap[raw_type]
   if not pathways_expected_instance:
-    xpk_print(f'Passed in device_type {device_type} is incorrect. Please pass in a valid device type')
+    xpk_print(
+        f'Passed in device_type {device_type} is incorrect. Please pass in a'
+        ' valid device type'
+    )
     xpk_exit(1)
   return pathways_expected_instance
+
 
 def get_pathways_worker_args(args) -> str:
   """Arguments for the Pathways workers.
--- a/xpk.py
+++ b/xpk.py
@@ -1113,7 +1113,7 @@ PathwaysExpectedInstancesMap = {
     "v5p": "v5",
     "v5e": "v5e",
     "v4": "v4",
-    "v3"" "v3,
+    "v3"" "v3",
 }
 
 def chunks(lst, n):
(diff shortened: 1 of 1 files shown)


xpk.py
```diff
@@ -1108,6 +1109,13 @@ class SystemCharacteristics:
 Map in MaxText/accelerator_to_spec_map.py """
 # ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
+PathwaysExpectedInstancesMap = {
+    "v5p": "v5",
+    "v5e": "v5e",
```
Obliviour, line 1114: probably `v5litepod: v5e`

SujeethJinesh: Thanks for the catch!

xpk.py
```diff
@@ -263,11 +263,12 @@ class CapacityType(enum.Enum):
         labels:
           xpk.google.com/workload: {args.workload}
       spec:
-        backoffLimit: 0
+        backoffLimit: 4
```
shaurya89, line 266: Can we have a special exit code for when pathways workers are dying after suspension? We should ideally not have that restart count towards the backofftLimit. I am ok if you wan to address that in a separate PR.

xpk.py
```diff
@@ -3862,12 +3870,43 @@ def get_pathways_rm_args(args) -> str:
...
               - --pathways_tmp_dir_pattern={args.pathways_gcs_location}
-              - --pathways_resource_manager_expected_num_worker_jobs={args.num_slices}"""
+              - --pathways_expected_instances={expected_instances}"""
   if args.use_pathways:
-    return yaml.format(args=args)
+    return yaml.format(args=args, expected_instances=compute_pathways_expected_instances(args, system))
   else:
     return ""
 
+def compute_pathways_expected_instances(args, system: SystemCharacteristics) -> str:
```
shaurya89, line 3879: Can we add some tests for these new functions? Like testing at least one valid and one invalid accelerator type.

shaurya89 (commented):
Thanks!

xpk.py
```diff
@@ -263,11 +263,12 @@ class CapacityType(enum.Enum):
         labels:
           xpk.google.com/workload: {args.workload}
       spec:
-        backoffLimit: 0
+        backoffLimit: 4
         completions: {system.vms_per_slice}
         parallelism: {system.vms_per_slice}
         template:
           spec:
+            terminationGracePeriodSeconds: {args.termination_grace_period_seconds}
```
RoshaniN, line 271: Could you elaborate on the need for this change?

xpk.py
```diff
@@ -3862,12 +3870,43 @@ def get_pathways_rm_args(args) -> str:
...
+  Args:
+    device_type: the system characteristic device type
+
+  Returns:
+    str: the device type expected by pathways.
+  """
+  raw_type = device_type.split('-')[0].lower()
+  pathways_expected_instance = PathwaysExpectedInstancesMap[raw_type]
+  if not pathways_expected_instance:
+    raise argparse.ArgumentTypeError(f'Passed in device_type {device_type} is incorrect. Please pass in a valid device type')
```
RoshaniN, line 3908: I would use existing xpk helper functions such as xpk_exit() with exit_code and xpk_print() to maintain uniformity.

xpk.py
```diff
@@ -3862,12 +3870,43 @@ def get_pathways_rm_args(args) -> str:
               - --pathways_persistent_compilation_cache=false
               - --pathways_compilation_mode=compile_at_worker
               - --pathways_tmp_dir_pattern={args.pathways_gcs_location}
-              - --pathways_resource_manager_expected_num_worker_jobs={args.num_slices}"""
+              - --pathways_expected_instances={expected_instances}"""
```
RoshaniN, line 3873: Thinking ahead, in terms of exposure is there a need to add this argument (and relevant logic) in XPK? 

Would it be better to have this argument as part of the RM image?

RoshaniN (commented):
Thanks for the change, Sujeeth!

xpk.py
SujeethJinesh: This would likely be a much more involved change as backoffLimit is purposely generic to all failures. I think making a change specifically for suspension would still impact the backoffLimit, or rather show the job as completed (would not be intended).

I don't think it's a change we'd want to support in the long term either. I'm open to make this value tunable for the user though in the future.

shaurya89: I thought GKE supported configuring the error types which should count towards your backoffLimit. My concern here would be that we would only be able to tolerate as many planned failures as the backoffLimit and that may be an issue for larger jobs.

shaurya89 (approved):

xpk.py
RoshaniN, line 271: I am not sure if this arg is actually added. 

You would need to add this argument, towards the end of this file. For example - https://github.com/google/xpk/blob/main/xpk.py#L6977-L6982

RoshaniN (changes requested):
Please make sure that the argument termination_grace_period_seconds is parsed like other XPK args.

xpk.py
SujeethJinesh: The arg is already added here. Thanks for checking!

https://github.com/google/xpk/blob/37fe185d664bc2111526cb1e4276a2a4f27c57f9/xpk.py#L5743-L5751

RoshaniN: Talked offline, this is part of workload create args!

RoshaniN: Talked offline, this is part of workload create args!

xpk.py
SujeethJinesh: Right now this is how Pathways takes in the argument, and it likely won't be in the roadmap to put this logic in the near future to support various accelerator options (possibly shapes and sizes and number).

xpk.py
SujeethJinesh: Done!

RoshaniN (approved):

xpk.py
SujeethJinesh: It looks like GKE does not consider SIGTERMs against the backoffLimit. I think we can explore this option as well in a different CL if we want to consider SIGTERM counting against restarts.

Obliviour (approved):

PR #130, opened 2024-04-17
Set gcloud zone property for build and nightly tests

SurbhiJainUSC:
## Fixes / Features
- Set Google Cloud CLI zone property for build and nightly tests. This will catch failures where gcloud command is missing zone flag in the workflow

## Testing / Documentation
Testing details.

- [ y ] Tests pass

### DIFF
--- a/.github/workflows/build_tests.yaml
+++ b/.github/workflows/build_tests.yaml
@@ -45,7 +45,7 @@ jobs:
         install_components: 'beta,gke-gcloud-auth-plugin'
     - name: Verify gcp setup
       run: gcloud info
-    - name: Set Google Cloud CLI properties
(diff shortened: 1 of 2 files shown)


.github/workflows/build_tests.yaml
```diff
@@ -45,6 +45,10 @@ jobs:
         install_components: 'beta,gke-gcloud-auth-plugin'
     - name: Verify gcp setup
       run: gcloud info
+    - name: Set Google Cloud CLI properties
```
Obliviour, line 48: nit to add `Set Google Cloud CLI properties` `to a unused zone to verify --zone arg is passed properly in commands.

Obliviour (approved):

.github/workflows/build_tests.yaml
SurbhiJainUSC: Do you mean to update the name?

PR #133, opened 2024-04-18
enable create workload for h150

NinaCai:
## Fixes / Features
- hot fix for reservation
- enable create workload for h150
- remove nccl plugin installation in workload creation

## Testing / Documentation
Testing details.

- [ y/n ] Tests pass

### DIFF
--- a/xpk.py
+++ b/xpk.py
@@ -72,6 +72,7 @@ __version__ = "0.3.0"
 xpk_current_version = __version__
 
 h100_device_type = 'h100-80gb-8'
+h150_device_type = 'h150-80gb-8'
 
 _AUTOPROVISIONING_CONFIG_VALUE = 'AUTOPROVISION'
 _AUTOPROVISIONING_CONFIG_MINIMUM_KEY = 'minimum_chips'
@@ -147,7 +148,7 @@ kind: JobSet
 metadata:
   name: {args.workload}
   labels:
-    kueue.x-k8s.io/queue-name: multislice-queue  # Name of the LocalQueue
+    {gpu_queue_name}
     xpk.google.com/workload: {args.workload}
 spec:
   failurePolicy:
@@ -181,7 +182,7 @@ spec:
                 {accelerator_label}
                 {machine_label}
                 {autoprovisioning_args}
-              priorityClassName: {args.priority}
+              {gpu_priority}
               hostNetwork: true
               dnsPolicy: ClusterFirstWithHostNet
               terminationGracePeriodSeconds: {args.termination_grace_period_seconds}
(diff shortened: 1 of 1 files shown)


xpk.py
```diff
@@ -3753,13 +3732,20 @@ def set_cluster_command(args) -> int:
...
     0 if successful and 1 otherwise.
   """
-  command = (
-      'gcloud container clusters get-credentials'
-      f' {args.cluster} --region={zone_to_region(args.zone)}'
-      f' --project={args.project} &&'
-      ' kubectl config view && kubectl config set-context --current'
-      ' --namespace=default'
-  )
+  if args.device_type == h150_device_type:
```
NinaCai, line 3735: @Obliviour Here we need to add `--device-type` back to workload deletion, because A3 and A3+ handle zone/region differently.

xpk.py
```diff
@@ -2123,7 +2102,7 @@ def parse_env_config(args, tensorboard_config):
     for key, value in tensorboard_config.items():
       env[key.upper()] = value
 
-  if device_type == h100_device_type:
+  if device_type == h100_device_type or device_type == h150_device_type:
```
Obliviour, line 2105: maybe change this to `if accelerator_type = AcceleratorType['gpu']`?

xpk.py
```diff
@@ -3120,7 +3099,7 @@ def get_capacity_arguments_from_capacity_type(
       capacity_args = '--spot'
     case CapacityType.RESERVATION:
       capacity_args = (
-          f'--reservation-affinity=specific --reservation={args.reservation}'
+        f'--reservation-affinity=specific --reservation={args.reservation} --placement-policy={args.reservation}'
```
Obliviour, line 3102: Is this needed in TPUs / CPUs? I see the build test failing with TPUs, I assume related to this.

xpk.py
Obliviour: `get-credentials` command is searching for the cluster. XPK only creates regional clusters so it should only need to search for a cluster using --region arg.

---

A scenario where you would need --zone is if the cluster was created outside of xpk using --zone. I am guessing that is what happened in the h150 cluster you are testing with. So we shouldn't need a different code path.

xpk.py
```diff
@@ -5181,7 +5191,52 @@ def get_gke_debugging_dashboard(args):
...
+  Args:
+    args: user provided arguments for running the command.
+
+  Returns:
+    str: name of the localQueue
+  """
+  if args.device_type == h100_device_type:
+    return "kueue.x-k8s.io/queue-name: multislice-queue"  # Name of the LocalQueue
+  else:
+    return "" #A3+ doesn't support Kueue yet
```
Obliviour, line 5206: Curious why A3+ doesn't support kueue? and what is missing there

Obliviour (changes requested):
Initial read, probably need to take a deeper look Monday.

PR #135, opened 2024-04-19
Dynamically determine GKE Version for Cluster and Node Pool Creation

Obliviour:
## Fixes / Features
- Determine the GKE version based on gcloud recommended versions
- Support addition node pools with the correct version (use the cluster's master version) instead of accidently using a new version by default
- Support --gke-version=x to override the default behavior.

This is super useful as xpk no longer has to change their gke version it uses whenever it disappears as a gke master version AND xpk will no longer by default recommend bad node pool versions.

## Testing / Documentation
Testing details.

- [ y ] Tests pass

### DIFF
--- a/xpk.py
+++ b/xpk.py
@@ -67,7 +67,6 @@ if (
 
 default_docker_image = 'python:3.10'
 default_script_dir = os.getcwd()
-default_gke_version = '1.29.1-gke.1589017'
 # This is the version for XPK PyPI package
 __version__ = '0.4.0'
 xpk_current_version = __version__
@@ -2171,7 +2170,12 @@ def append_temporary_file(payload, file):
 
 
 def run_command_for_value(
-    command, task, global_args, dry_run_return_val='0', print_timer=False
+    command,
+    task,
+    global_args,
+    dry_run_return_val='0',
+    print_timer=False,
+    hide_error=False,
 ) -> tuple[int, str]:
   """Runs the command and returns the error code and stdout.
 
@@ -2182,6 +2186,8 @@ def run_command_for_value(
     task: user provided task name for running the command.
     global_args: user provided arguments for running the command.
     dry_run_return_val: return value of this command for dry run.
+    print_timer: print out the time the command is running.
+    hide_error: hide the error from the command output upon success.
 
   Returns:
     tuple[int, str]
@@ -2225,7 +2231,7 @@ def run_command_for_value(
       output = subprocess.check_output(
           command,
           shell=True,
-          stderr=subprocess.STDOUT,
+          stderr=subprocess.STDOUT if not hide_error else None,
       )
     except subprocess.CalledProcessError as e:
       xpk_print(f'Task {task} failed with {e.returncode}')
@@ -2755,11 +2761,12 @@ def enable_autoprovisioning_on_cluster(
   return autoprovisioning_config, return_code
 
 
-def run_gke_cluster_create_command(args) -> int:
+def run_gke_cluster_create_command(args, gke_control_plane_version: str) -> int:
   """Run the Create GKE Cluster request.
 
   Args:
     args: user provided arguments for running the command.
+    gke_control_plane_version: version used if creating the cluster.
 
   Returns:
     0 if successful and 1 otherwise.
@@ -2784,7 +2791,7 @@ def run_gke_cluster_create_command(args) -> int:
       f' {args.cluster} --project={args.project}'
       f' --region={zone_to_region(args.zone)}'
       f' --node-locations={args.zone}'
-      f' --cluster-version={args.gke_version}'
+      f' --cluster-version={gke_control_plane_version}'
       f' --machine-type={machine_type}'
       ' --enable-autoscaling'
       ' --total-min-nodes 1 --total-max-nodes 1000'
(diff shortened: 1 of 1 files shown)


xpk.py
```diff
@@ -4072,6 +4076,183 @@ def install_nccl_on_cluster(args) -> int:
...
+  return_code, default_rapid_gke_version = run_command_for_value(
+      default_rapid_gke_version_cmd, f'{base_command_description} {task}.', args
+  )
+  if return_code != 0:
+    xpk_print(f'Unable to get server config for {task}.')
+    return return_code, None
+
+  task = ' valid master versions'
+  return_code, valid_master_versions = run_command_for_value(
+      valid_master_versions_cmd, f'{base_command_description} .', args
```
SurbhiJainUSC, line 4128: Did you miss {task}?

xpk.py
```diff
@@ -4072,6 +4076,183 @@ def install_nccl_on_cluster(args) -> int:
...
+
+  task = 'default rapid gke version'
+  return_code, default_rapid_gke_version = run_command_for_value(
+      default_rapid_gke_version_cmd, f'{base_command_description} {task}.', args
+  )
+  if return_code != 0:
+    xpk_print(f'Unable to get server config for {task}.')
+    return return_code, None
+
+  task = ' valid master versions'
```
SurbhiJainUSC, line 4126: nit: unnecessary whitespace at the start.

xpk.py
```diff
@@ -4072,6 +4076,183 @@ def install_nccl_on_cluster(args) -> int:
...
+  return_code, valid_master_versions = run_command_for_value(
+      valid_master_versions_cmd, f'{base_command_description} .', args
+  )
+  if return_code != 0:
+    xpk_print(f'Unable to get server config for {task}.')
+    return return_code, None
+
+  task = 'valid node versions'
+  return_code, valid_node_versions = run_command_for_value(
+      valid_node_versions_cmd, f'{base_command_description} .', args
```
SurbhiJainUSC, line 4136: nit: missed {task}

xpk.py
```diff
@@ -4072,6 +4076,183 @@ def install_nccl_on_cluster(args) -> int:
...
+      " {print $2}' "
+  )
+  valid_node_versions_cmd = (
+      base_command
+      + " | awk '/validNodeVersions/,/END/' | awk '/-/,/-/ {print $2}' "
+  )
+  base_command_description = 'Determine server supported GKE versions for'
+
+  task = 'default rapid gke version'
+  return_code, default_rapid_gke_version = run_command_for_value(
```
SurbhiJainUSC, line 4119: Can we move these 3 run_command_for_value() into a single method that reduces human error in the commands, logs etc.
Something like this:
```
gke_version_command_map = {}
gke_version_command_map['default rapid gke version'] =  default_rapid_gke_version_cmd
for task_name, command in gke_version_command_map.items():
      return_code, gke_version = run_command_for_value(
           command, f'{base_command_description} {task_name}.', args)
      if return_code != 0:
         xpk_print(f'Unable to get server config for {task}.')
      return return_code, None
```

xpk.py
```diff
@@ -4072,6 +4076,183 @@ def install_nccl_on_cluster(args) -> int:
...
+        f'\nRecommended GKE Version: {gke_server_config.valid_node_versions}'
+    )
+    xpk_print(
+        'Error: Planned GKE Version {master_gke_version} is not valid.'
+        f'Checks failed: Is Master Valid: {is_valid_master_version}'
+        f'\nIs Valid Node Version: {is_valid_node_version}'
+    )
+    xpk_print(
+        'Please select a gke version from the above list using --gke-version=x'
+        ' argument. or rely on the default gke version:'
```
SurbhiJainUSC, line 4193: nit: remove `.` after argument?

xpk.py
```diff
@@ -4072,6 +4076,183 @@ def install_nccl_on_cluster(args) -> int:
...
+        ' {gke_server_config.valid_node_versions}'
```
SurbhiJainUSC, line 4194: Is this the default gke version? Isn't `valid_node_versions` a set of node pool gke version?

xpk.py
```diff
@@ -4072,6 +4076,183 @@ def install_nccl_on_cluster(args) -> int:
...
+  # Override with user provide gke version if specified.
+  if args.gke_version is not None:
+    master_gke_version = args.gke_version
+  else:
+    master_gke_version = gke_server_config.default_rapid_gke_version
+
+  is_valid_master_version = (
+      master_gke_version in gke_server_config.valid_master_versions
+  )
+  is_valid_node_version = (
```
SurbhiJainUSC, line 4173: Why do we need to check the node pool gke version here? Isn't `get_gke_node_pool_version()` doing this?

Obliviour: This is needed for cpu node pools created through cluster create command.

xpk.py
Obliviour: fixed

xpk.py
Obliviour: fixed

xpk.py
Obliviour: Done! Thanks for the suggestion!

SurbhiJainUSC (approved):

PR #137, opened 2024-04-19
Update xpk.py

kyle-google:
update default version for gke

## Fixes / Features
- tcpx does not work with the default cluster version

### DIFF
--- a/xpk.py
+++ b/xpk.py
@@ -67,7 +67,7 @@ if (
 
 default_docker_image = 'python:3.10'
 default_script_dir = os.getcwd()
-default_gke_version = '1.29.1-gke.1589017'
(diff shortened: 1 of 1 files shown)


Obliviour:
Thank you @kyle-google for this change! we notice the gke master versions supported change over time creating this scenario where xpk can't create a cluster.

I created a future looking fix here: https://github.com/google/xpk/pull/135 that dynamically figures the gke version. I am hoping we can depend on this instead of manually increasing the gke version. This should be merged today. LMK if you still need the update change you have here.

Obliviour:
@kyle-google #135 was merged in. LMK if you are no longer seeing issues, and we can close this PR.

PR #139, opened 2024-04-23
Making exit flow similar to other XPK commands.

RoshaniN:
## Fixes / Features
- Making exit flow similar to other XPK commands, to use in Pathways integration tests.
- Ensuring that all the functions exit using xpk_exit() flow, cleanup return types.

## Testing / Documentation
Testing details.

- [ y ] Tests pass

### DIFF
--- a/xpk.py
+++ b/xpk.py
@@ -4098,7 +4098,7 @@ def default_subcommand_function(
   return 0
 
 
-def cluster_create(args) -> int:
+def cluster_create(args) -> None:
   """Function around cluster creation.
 
   Args:
@@ -4228,7 +4228,7 @@ def cluster_create(args) -> int:
   xpk_exit(0)
 
 
-def cluster_delete(args):
+def cluster_delete(args) -> None:
   """Function around cluster delete.
 
   Args:
@@ -4246,7 +4246,7 @@ def cluster_delete(args):
   xpk_exit(0)
 
 
-def cluster_cacheimage(args) -> int:
+def cluster_cacheimage(args) -> None:
   """Function around cluster cacheimage.
 
   Args:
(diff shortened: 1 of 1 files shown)


xpk.py
```diff
@@ -4243,7 +4243,7 @@ def cluster_delete(args) -> int:
   if run_gke_cluster_delete_command_code != 0:
     xpk_exit(run_gke_cluster_delete_command_code)
   xpk_print(f'GKE commands done! Cluster {args.cluster} deleted.\n')
-  return 0
+  return xpk_exit(0)
```
Obliviour, line 4246: xpk_exit doesn't return any value.

Probably what you want is without the return statement:
`xpk_exit(0)`

---

Most of the subcommand functions like cluster_delete are actually misleading to say they return int. They actually never return. Some longer term clean up exists there.

RoshaniN: Aah, makes sense. Thank you

Obliviour (approved):
Thanks for the change!

I agree we should make the major functions ( subcommand functions like xpk cluster [create / delete / list], xpk workload [xxx, yyy, zzz]) remove the return value from them
do you want to do that in this PR as well?

xpk.py
```diff
@@ -4228,7 +4228,7 @@ def cluster_create(args) -> int:
   xpk_exit(0)
 
 
-def cluster_delete(args) -> int:
+def cluster_delete(args):
```
Obliviour, line 4231: nit`  -> None`

RoshaniN: Addressed and took care of the other high level functions also.

RoshaniN:
> Thanks for the change!
> 
> I agree we should make the major functions ( subcommand functions like xpk cluster [create / delete / list], xpk workload [xxx, yyy, zzz]) remove the return value from them do you want to do that in this PR as well?

Sure, done!

PR #141, opened 2024-05-02
Pathways in headless mode.

RoshaniN:
## Fixes / Features
- Enable Cloud DNS on Pathways clusters.
- Add user job conditionally based on whether headless mode is defined or not.
- Print out proxy address for Pathways in headless mode. Users would connect to this address from Vertex AI notebook / GCE VMs, to execute their JAX workloads.
- Organized XPK subcommands for Pathways.
- - Add cluster create-pathways subcommand. Deprecate --enable-pathways.
- -  Add workload create-pathways subcommand. Deprecate --use-pathways.
- Update build tests and nightly tests, reordered a few tests.
- Restricting some features to create as they are not suitable for create-pathways, at the moment. (eg: autoprovisioning,  debug dump to gcs, stack trace sidecar and restart on user failures.)

## Testing / Documentation
Testing details.

- [ y ] Tests pass

### DIFF
--- a/xpk.py
+++ b/xpk.py
@@ -367,28 +367,7 @@ spec:
                   memory: 100G
             nodeSelector:
               cloud.google.com/gke-nodepool: cpu-proxy-np
-  - name: main
-    replicas: 1
-    template:
-      metadata:
-        labels:
-          xpk.google.com/workload: {args.workload}
-      spec:
-        backoffLimit: 0
-        completions: 1
-        parallelism: 1
-        template:
-          spec:
-            containers:
-              {container}
-            nodeSelector:
-              cloud.google.com/gke-nodepool: cpu-user-np
-            restartPolicy: OnFailure
-            volumes:
-            - hostPath:
-                path: /tmp
-                type: DirectoryOrCreate
-              name: shared-tmp
+  {user_workload}
 """
 
 script_dir_dockerfile = """FROM {base_docker_image}
@@ -5162,6 +5141,85 @@ def get_pathways_proxy_args(args) -> str:
     return ''
 
 
+def get_user_workload_container(args, system):
+  """Deploy user workload container
+
+  Args:
+      args: user provided args.
+      system: system characteristics.
+
+  Returns:
+      container: main container
+      debugging_dashboard_id: id of the GKE dashboard
+  """
+
+  setup_docker_image_code, docker_image = setup_docker_image(args)
+  if setup_docker_image_code != 0:
+    xpk_exit(setup_docker_image_code)
+
+  # Determine if we deploy a sidecar and if we deploy a container.
+  debugging_dashboard_id = None
+  resource_type = AcceleratorTypeToAcceleratorCharacteristics[
+      system.accelerator_type
+  ].resource_type
+  if (
+      system.accelerator_type == AcceleratorType['TPU']
+      and args.deploy_stacktrace_sidecar
+  ):
+    xpk_print(
+        'Sidecar container to display stack traces for TPU workloads will also'
+        ' be deployed.'
+    )
+    container = get_main_and_sidecar_container(args, system, docker_image)
+    # Get GKE debugging dashboard only when sidecar container is deployed for TPU workloads
+    debugging_dashboard_id = get_gke_debugging_dashboard(args)
+  else:
+    container = get_main_container(args, system, docker_image, resource_type)
+  return container, debugging_dashboard_id
+
+
+def get_user_workload_for_pathways(args, system) -> str:
+  """
+  Create a user workload container for Pathways.
+  Don't create one for Pathways headless mode.
+
+  Args:
+    args: user provided args.
+    container: container with the user workload.
+
+  Returns:
+    str:
+      Pathways server port as a YAML string
+  """
+  user_workload_yaml = """- name: main
+    replicas: 1
+    template:
+      metadata:
+        labels:
+          xpk.google.com/workload: {args.workload}
+      spec:
+        backoffLimit: 0
+        completions: 1
+        parallelism: 1
+        template:
+          spec:
+            containers:
+              {container}
+            nodeSelector:
+              cloud.google.com/gke-nodepool: cpu-user-np
+            restartPolicy: OnFailure
+            volumes:
+            - hostPath:
+                path: /tmp
+                type: DirectoryOrCreate
+              name: shared-tmp"""
+  if args.use_pathways and not args.headless:
+    container, _ = get_user_workload_container(args, system)
+    return user_workload_yaml.format(args=args, container=container)
+  else:
+    return ''
+
+
 def get_env_container(args, system: SystemCharacteristics):
   """Environment configuration for the main container.
   Args:
@@ -5723,9 +5781,7 @@ def workload_create(args) -> None:
         ' cluster create`.'
     )
 
-  setup_docker_image_code, docker_image = setup_docker_image(args)
-  if setup_docker_image_code != 0:
-    xpk_exit(setup_docker_image_code)
+  debugging_dashboard_id = None
 
   tensorboard_config = {}
   if _VERTEX_TENSORBOARD_FEATURE_FLAG and args.use_vertex_tensorboard:
@@ -5750,31 +5806,14 @@ def workload_create(args) -> None:
     if return_code != 0:
       xpk_exit(return_code)
 
-  # Determine if we deploy a sidecar and if we deploy a container.
-  debugging_dashboard_id = None
-  resource_type = AcceleratorTypeToAcceleratorCharacteristics[
-      system.accelerator_type
-  ].resource_type
-  if (
-      system.accelerator_type == AcceleratorType['TPU']
-      and args.deploy_stacktrace_sidecar
-  ):
-    xpk_print(
-        'Sidecar container to display stack traces for TPU workloads will also'
-        ' be deployed.'
-    )
-    container = get_main_and_sidecar_container(args, system, docker_image)
-    # Get GKE debugging dashboard only when sidecar container is deployed for TPU workloads
-    debugging_dashboard_id = get_gke_debugging_dashboard(args)
-  else:
-    container = get_main_container(args, system, docker_image, resource_type)
-
   # Create the workload file based on accelerator type or workload type.
   if system.accelerator_type == AcceleratorType['GPU']:
+    container, debugging_dashboard_id = get_user_workload_container(
+        args, system
+    )
     yml_string = gpu_workload_create_yaml.format(
         args=args,
         container=container,
-        docker_image=docker_image,
         command=args.command,
         accelerator_label=create_accelerator_label(
             system.accelerator_type, system
@@ -5803,7 +5842,6 @@ def workload_create(args) -> None:
     yml_string = pw_workload_create_yaml.format(
         args=args,
         system=system,
-        container=container,
         accelerator_label=create_accelerator_label(
             system.accelerator_type, system
         ),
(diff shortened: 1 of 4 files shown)


xpk.py
```diff
@@ -2609,21 +2585,13 @@ def run_gke_cluster_create_command(args, gke_control_plane_version: str) -> int:
   Returns:
     0 if successful and 1 otherwise.
   """
-  # cluster_cpu_machine_type is soon to be deprecated!
   machine_type = args.default_pool_cpu_machine_type
-  if args.cluster_cpu_machine_type != '':
```
Obliviour, line 2614: Lets depreciate this in a separate PR. Couldn't potentially cause issues if folks are still using this arg.

xpk.py
```diff
@@ -4378,12 +4357,14 @@ def cluster_create(args) -> None:
...
       xpk_exit(install_nccl_code)
 
-  xpk_print('Creating ConfigMap for cluster')
-  create_cluster_configmaps_code = create_cluster_configmaps(
-      args, system, tensorboard_config, autoprovisioning_config
-  )
-  if create_cluster_configmaps_code != 0:
-    xpk_exit(create_cluster_configmaps_code)
+  # Currently autoprovisioning is not supported with Pathways.
+  if not args.enable_pathways:
```
Obliviour, line 4361: This code is not autoprovisioning specific. It will create a configmap describing what accelerator types are used in the cluster.

xpk.py
```diff
@@ -5118,6 +5099,85 @@ def get_pathways_proxy_args(args) -> str:
     return ''
 
 
+def get_user_workload_container(args, system):
```
Obliviour, line 5102: nit `system: SystemCharacteristics`

xpk.py
```diff
@@ -5118,6 +5099,85 @@ def get_pathways_proxy_args(args) -> str:
...
+    )
+    container = get_main_and_sidecar_container(args, system, docker_image)
+    # Get GKE debugging dashboard only when sidecar container is deployed for TPU workloads
+    debugging_dashboard_id = get_gke_debugging_dashboard(args)
+  else:
+    container = get_main_container(args, system, docker_image, resource_type)
+  return container, debugging_dashboard_id
+
+
+def get_user_workload_for_pathways(args, system) -> str:
```
Obliviour, line 5139: nit system: SystemCharacteristics

xpk.py
```diff
@@ -5694,44 +5772,28 @@ def workload_create(args) -> None:
...
-  autoprovisioning_enabled, return_code = is_autoprovisioning_enabled(
-      args, system
-  )
-  if return_code != 0:
-    xpk_exit(return_code)
-  if autoprovisioning_enabled:
-    # Determine NAP capacity type
-    autoprovisioning_args, return_code = (
-        get_autoprovisioning_node_selector_args(args)
+  if not args.use_pathways:
```
Obliviour, line 5775: in theory we shouldn't need to check use_pathways since auto provisioning won't be enabled in the first place.

xpk.py
```diff
@@ -5833,6 +5890,37 @@ def workload_create(args) -> None:
         f' {args.workload} from the JobName filter on the dashboard.'
     )
 
+  if args.use_pathways:
```
Obliviour, line 5893: Have you considering have a Cloud Console UI page for pathways that could show these links. Then XPK only needs to direct to one link for example.

xpk.py
```diff
@@ -6497,6 +6585,398 @@ def add_shared_arguments(custom_parser):
...
+        '--spot',
+        action='store_true',
+        help=(
+            'Sets node pool creation to use spot resources.'
+            ' See `--reservation` or `--on-demand` for other capacity types.'
+        ),
+    )
+
+
+def add_shared_workload_create_required_arguments(create_args_parsers):
```
Obliviour, line 6776: nit workload_arg_parsers or just arg_parsers. I see some inconsistencies in the other workload shared argument functions.

xpk.py
```diff
@@ -6497,6 +6585,398 @@ def add_shared_arguments(custom_parser):
   )
 
 
+def add_shared_cluster_create_required_arguments(create_args_parsers):
```
Obliviour, line 6588: These shared functions are great!!!

xpk.py
```diff
@@ -7265,6 +7516,79 @@ def directory_path_type(value):
...
+)
+workload_create_pathways_docker_image_arguments = workload_create_pathways_parser.add_argument_group(
+    'Docker Image Arguments',
+    '`--base-docker-image` is used by default. Set this argument if the'
+    ' user wants the docker image to be used directly by the xpk workload.',
+)
+
+### "workload create-pathways" Required arguments, specific to Pathways
+workload_create_pathways_parser_required_arguments.add_argument(
+    '--tpu-type',
```
Obliviour, line 7550: hmm a bit confused, why does pathways_parser for workload create need --tpu-type specifically when it is part of workload_create arguments as well

xpk.py
RoshaniN: Thanks for the catch! I udpated the logic to disable autoprovisioning specific code path within this function.

xpk.py
RoshaniN: For sure, we now have a lot of items to deprecate. Perhaps a PR targeting cleanup would be helpful!

xpk.py
RoshaniN: Copy Paste error, fixed now!

xpk.py
RoshaniN: Copy Paste error, fixed now!

xpk.py
RoshaniN: I agree!
I added this earlier to avoid accessing autoprovisioning related arguments which were not "defined" for the create-pathways subcommand. I may be able to remove this condition now.

xpk.py
RoshaniN: workload create supports provision of one out of tpu-type or device-type in a mutually exclusive group (GPUs and CPUs are supported).
workload create-pathways can only support tpu-type.

These args are hence not added in the shared parsers command!

xpk.py
RoshaniN: The intention here was to talk about the action "create" for both clusters and workloads. I will change it if that causes confusion, to keep something more generic.

xpk.py
```diff
@@ -5833,6 +5902,38 @@ def workload_create(args) -> None:
...
+          f' {pathways_proxy_link} '
+      )
+    else:
+      pathways_job_link = f'https://console.cloud.google.com/kubernetes/job/{zone_to_region(args.zone)}/{args.cluster}/default/{args.workload}-{args.targetReplicatedJob}-0/details?project={args.project}'
+      xpk_print(
+          'Follow your Pathways workload here:'
+          # pylint: disable=line-too-long
+          f' {pathways_job_link} '
+      )
+    xpk_print(
```
shaurya89, line 5926: We should probably have a unified debugging view here as well (just remove the "main" job from the query)

xpk.py
```diff
@@ -6497,88 +6598,477 @@ def add_shared_arguments(custom_parser):
...
 
-def workload_name_type(value, pat=re.compile(r'[a-z]([-a-z0-9]*[a-z0-9])?')):
-  """Validate that the workload name matches the expected pattern."""
-  match = pat.fullmatch(value)
-  if not match or len(match.group(0)) > 40:
-    raise argparse.ArgumentTypeError(
-        'Workload name must be less than 40 characters and match the pattern'
-        f' `{pat.pattern}`'
-        f' Name is currently {value}'
+def add_shared_cluster_create_optional_arguments(args_parsers):
```
shaurya89, line 6620: What does "shared" in the function name imply? If it indicates common functionality across use cases, maybe consider "common"?

shaurya89 (commented):
Thanks, sorry for the delay.

xpk.py
RoshaniN: Sure, will do.

xpk.py
RoshaniN: Sure, "shared" in the function names implies that these arguments are shared between cluster create and cluster create-pathways (analogously shared between `workload create` and `workload create-pathways`) subcommands.

xpk.py
Obliviour: It looks like the code is still being removed in this PR?

xpk.py
```diff
@@ -4262,6 +4240,11 @@ def default_subcommand_function(
   return 0
 
 
+def cluster_create_pathways(args) -> None:
```
Obliviour, line 4243: Nit for some docstring

xpk.py
```diff
@@ -5680,12 +5767,15 @@ def workload_create(args) -> None:
...
+  debugging_dashboard_id = None
 
   tensorboard_config = {}
-  if _VERTEX_TENSORBOARD_FEATURE_FLAG and args.use_vertex_tensorboard:
+  if (
+      not args.use_pathways
+      and _VERTEX_TENSORBOARD_FEATURE_FLAG
+      and args.use_vertex_tensorboard
+  ):
+    # Profiling is handled differently in Pathways workloads
```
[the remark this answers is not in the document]
Obliviour: Hmm tensorboard is a very nice for users to inspect their workloads. I am guessing it can be used nicely in pathways so i am curious to see how we integrate pathways and tensorboard in the future

xpk.py
Obliviour: Hmm maybe some changes are missing, I still see if not args.use_pathways here

xpk.py
RoshaniN: The change to revert this was lost in a stash :) 
Sorry about that!

xpk.py
RoshaniN: I don't see it on my end!

xpk.py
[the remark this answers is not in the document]
shaurya89: I am not sure I follow the reasoning. I would suspect maxtext already has the functionality to do the upload. What specifically is different in the Pathways setup that you see as a concern?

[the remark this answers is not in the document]
RoshaniN: There is nothing specific to Pathways, this feature is available through Maxtext + XPK. 

My only concern is that this feature is not extensible to other JAX workloads, that might be used with Pathways, without the specific changes mentioned above.

xpk.py
```diff
@@ -4889,15 +4889,16 @@ def get_main_container(args, system, docker_image, resource_type) -> str:
...
+      and system.accelerator_type == AcceleratorType['TPU']
       and args.deploy_stacktrace_sidecar
   ):
     tpu_stacktrace_terminate_command = (
         'touch /shared-volume/stacktrace_signal; '
     )
 
   xpk_return_user_exit_code = ''
-  if args.restart_on_user_code_failure:
+  if not args.use_pathways and args.restart_on_user_code_failure:
```
Obliviour, line 4901: Do we need many of these `if not args.use_pathways`? It the argument doesn't exist in the create_pathways subfunction, then that prevents this code path right? 

I feel like having several if not args.use_pathways will be hard to maintain.

RoshaniN: It is necessary to have the condition to not go through this code path. Without the argument in the create_pathways subfunction, the code path is not prevented, but arg not found error is observed.

Obliviour: It sounds cleaner to use the arguments set in subfunctions to determine which code paths can be run instead of having to do if checks in the code? What do you think, and let me know if there is some error that prevents this from working!

RoshaniN: I agree, I tried to keep it as simple and clean as I could. However, some args are not shared (present in create and not in create_pathways). When a user runs create_pathways and XPK goes through these code paths, the args are "not defined" and hence not found. 
Example -
"AttributeError: 'Namespace' object has no attribute 'deploy_stacktrace_sidecar'"

I have hence meticulously abstracted the code paths that are not needed for Pathways.

xpk.py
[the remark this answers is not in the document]
RoshaniN: I have added the feature back in, with the caveat that users would need to modify their workload (similar to Maxtext) to upload the logs, in order to be able to use them in Pathways + XPK.

shaurya89 (approved):

Obliviour (approved):

xpk.py
```diff
@@ -5121,7 +5125,7 @@ def get_pathways_worker_args(args) -> str:
   """
   yaml = """- --alsologtostderr
               - --pathways_server_port=38677
-              - --pathways_resource_manager={args.workload}-rm-0-0.{args.workload}:38677
+              - --pathways_resource_manager={args.workload}-rm-0-0.{args.workload}.default.svc.{args.cluster}-domain.:38677
```
shaurya89, line 5128: I think this and all the rest of the changes to the DNS addresses would break existing users who are using clusters that do not use cloud DNS. We should probably share the steps to do that with all users or hide this change behind some check for cloud DNS.

PR #145, opened 2024-05-14
Enable cluster and workload creation on A3+.

yangyuwei:
## Fixes / Features
- remove nccl plugin installation in workload creation on A3
- enable cluster and workload creation on A3+

## Testing / Documentation
Manual tests passed for cluster creation and workload creation on A3+.

- [ y ] Tests pass

### DIFF
--- a/xpk.py
+++ b/xpk.py
@@ -2742,12 +2742,11 @@ def create_cluster_subnet(args, index) -> int:
   return 0
 
 
-def delete_cluster_subnets(args, system: SystemCharacteristics) -> int:
+def delete_cluster_subnets(args) -> int:
   """Delete GKE Cluster subnets.
 
   Args:
     args: user provided arguments for running the command.
-    system: system characteristics.
 
   Returns:
     0 if successful and 1 otherwise.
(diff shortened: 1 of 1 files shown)


xpk.py
```diff
@@ -2660,17 +2646,20 @@ def run_gke_cluster_create_command(args, gke_control_plane_version: str) -> int:
...
+  and there are 8 NICs for GPU-GPU bw and 1 NIC for host in an A3+ node.
 
   Args:
     args: user provided arguments for running the command.
 
   Returns:
     0 if successful and 1 otherwise.
   """
-  for i in range(1, 5):
+  device_type = args.tpu_type if args.tpu_type else args.device_type
```
Obliviour, line 2660: Could we pass the SystemCharacteristics into`set_up_cluster_network_for_gpu` instead of checking the tpu/device_type args? Hoping to avoid duplication of this arg check in the code and propagate SystemCharacteristics instead

xpk.py
```diff
@@ -4049,11 +4008,20 @@ def install_nccl_on_cluster(args) -> int:
   Returns:
     0 if successful and 1 otherwise.
   """
-  command = (
-      'kubectl apply -f'
-      # pylint: disable=line-too-long
-      'https://raw.githubusercontent.com/GoogleCloudPlatform/container-engine-accelerators/master/gpudirect-tcpx/nccl-tcpx-installer.yaml'
-  )
+  device_type = args.tpu_type if args.tpu_type else args.device_type
```
Obliviour, line 4011: Nit to pass in SystemCharacteristics

xpk.py
```diff
@@ -5602,6 +5573,96 @@ def get_autoprovisioning_node_selector_args(args) -> tuple[str, int]:
...
+def get_gpu_volume(args) -> str:
+  """Get gpu volume based on user provided arguments.
+
+  Args:
+    args: user provided arguments for running the command.
+
+  Returns:
+    str: yaml containing gpu volume
+  """
+  gpu_volumn = ""
```
Obliviour, line 5585: nit volume

xpk.py
```diff
@@ -5602,6 +5573,96 @@ def get_autoprovisioning_node_selector_args(args) -> tuple[str, int]:
...
+  if args.device_type == h100_device_type:
```
Obliviour, line 5586: Can pass systemcharacteristics instead and use system.device_type

xpk.py
```diff
@@ -5722,6 +5783,10 @@ def workload_create(args) -> None:
         node_pool_name=f'{args.cluster}-np-0',
         chips_per_vm=system.chips_per_vm,
         autoprovisioning_args=autoprovisioning_args,
+        gpu_volume=get_gpu_volume(args),
+        gpu_rxdm_image=get_gpu_rxdm_image(args),
+        gpu_rxdm_cmd=get_gpu_rxdm_cmd(args),
+        gpu_tcp_volume=get_gpu_tcp_volume(args)
```
Obliviour, line 5789: I think we can pass system instead of args here? Mainly to have consistency in how we parse the device usage. This will be useful for future refactors

Obliviour (commented):
LGTM overall! Just some small requests

yangyuwei:
> LGTM overall! Just some small requests

Thanks for the review! I've updated the code based on your comments in the latest commit (https://github.com/google/xpk/commit/091edf65ba29b6311c098665d911527b314ec60f). PTAL.

xpk.py
```diff
@@ -2764,16 +2755,17 @@ def delete_cluster_subnets(args) -> int:
   if return_code > 0:
     xpk_print('Listing all subnets failed!')
     return return_code
-  for index in range(1, 5):
+
+  for index in range(1, 9):
```
Obliviour, line 2759: Does this need to be 1,5 for a3 scenario and 1,9 for a3+?

Obliviour (approved):
Just one small comment! LGTM

xpk.py
yangyuwei: Nice catch! Based on offline discussion, I updated get_all_subnets_programmatic() to get more accurate results for subnets, so that we don't need to hardcode the number of subnets here. I also cleaned up the code based on pylint checks. PTAL.

Obliviour (approved):
Thank you Yuwei!

michelle-yooh (approved):

PR #151, opened 2024-05-29
Add a check and update existing Pathways clusters to use CloudDNS.

RoshaniN:
## Fixes / Features
- Update existing Pathways clusters to use CloudDNS without user intervention.
- This would avoid DNS resolution failures on existing Pathways clusters.
- Add a check to find if the cluster is using CloudDNS or not and upgrade automatically, as needed. 
- Upgrade reflects after 3 steps:
- - Step 1: Update cluster dns to use cloud dns
- - Step 2: Update GKE control plane version
- - Step 3: Update nodepools, which recreates nodes.

## Testing / Documentation
1. Upgrade existing cluster during the first "cluster create-pathways".
2. Upgrade existing cluster during the first "workload create-pathways".
3. No extra steps needed from the user

- [ y] Tests pass

### DIFF
--- a/xpk.py
+++ b/xpk.py
@@ -2642,6 +2642,7 @@ def run_gke_pathways_cluster_update_command(args) -> int:
       f' --cluster-dns-domain={args.cluster}-domain'
       ' --quiet'
   )
+  xpk_print('Updating Pathways cluster to use Cloud DNS, may take a while!')
   return_code = run_command_with_updates(
       command, 'Pathways Cluster Update to enable Cloud DNS', args
   )
@@ -3299,6 +3300,33 @@ def create_cluster_if_necessary(
     )
 
 
+def is_cluster_using_clouddns(args) -> int:
+  """Checks if cluster is using CloudDNS.
+  Args:
+    args: user provided arguments for running the command.
+
+  Returns:
+    0 if cluster is using CloudDNS and 1 otherwise.
+  """
+  command = (
+      f'gcloud container clusters describe {args.cluster}'
+      f' --project={args.project} --region={zone_to_region(args.zone)}'
+      ' | grep "clusterDns: CLOUD_DNS" | wc -l'
+  )
+  return_code, cloud_dns_matches = run_command_for_value(
+      command,
+      'Check if Cloud DNS is enabled in cluster describe.',
+      args,
+  )
+  if return_code != 0:
+    xpk_exit(return_code)
+  cloud_dns_matches = int(cloud_dns_matches)
+  if cloud_dns_matches > 0:
+    xpk_print('Cloud DNS is enabled on the cluster, no update needed.')
+    return 0
+  return 1
+
+
 def update_pathways_cluster_if_necessary(args) -> int:
   """Updates a Pathways cluster to use CloudDNS.
 
@@ -3313,6 +3341,9 @@ def update_pathways_cluster_if_necessary(args) -> int:
     xpk_print('Listing all clusters failed!')
     return 1
   if args.cluster in all_clusters:
+    # If cluster is already using clouddns, no update necessary!
+    if is_cluster_using_clouddns(args) == 0:
+      return 0
     cluster_update_return_code = run_gke_pathways_cluster_update_command(args)
     if cluster_update_return_code > 0:
       xpk_print('Updating Pathways cluster to use CloudDNS failed!')
@@ -5927,6 +5958,10 @@ def workload_create(args) -> None:
     0 if successful and 1 otherwise.
   """
   add_zone_and_project(args)
+
+  if args.use_pathways:
+    update_pathways_cluster_if_necessary(args)
+
   if args.command is None and not args.headless:
     xpk_print(
         'Please provide a command using "--command" for the docker container to'
--- a/xpk.py
+++ b/xpk.py
@@ -5248,6 +5248,38 @@ def get_pathways_expected_tpu_type(device_type: str) -> str:
   return pathways_expected_instance
 
 
+def get_rm_address_based_on_clouddns(args) -> str:
+  """Generates the Pathways resource manager address based on whether CloudDNS is enabled or not.
+  Args:
+    args: user provided arguments for running the command.
+
+  Returns:
+    str: Fully qualified RM address.
+  """
+  suffix = ''
+  if is_cluster_using_clouddns(args):
+    suffix = f'.default.svc.{args.cluster}-domain.'
+  rm_address = f'{args.workload}-rm-0-0.{args.workload}{suffix}:38677'
+  return rm_address
+
+
+def get_proxy_address_based_on_clouddns(args) -> str:
+  """Generates the Pathways proxy address based on whether CloudDNS is enabled or not.
+  Args:
+    args: user provided arguments for running the command.
+
+  Returns:
+    str: Fully qualified proxy address.
+  """
+  suffix = ''
+  if is_cluster_using_clouddns(args):
+    suffix = f'.default.svc.{args.cluster}-domain.'
+  proxy_address = (
+      f'grpc://{args.workload}-proxy-0-0.{args.workload}{suffix}:38676'
+  )
+  return proxy_address
+
+
 def get_pathways_worker_args(args) -> str:
   """Arguments for the Pathways workers.
   Args:
@@ -5258,7 +5290,7 @@ def get_pathways_worker_args(args) -> str:
   """
   yaml = """- --alsologtostderr
               - --pathways_server_port=38677
-              - --pathways_resource_manager={args.workload}-rm-0-0.{args.workload}.default.svc.{args.cluster}-domain.:38677
+              - --pathways_resource_manager={rm_address}
               - --pathways_persistent_compilation_cache=false
               - --pathways_compilation_mode=compile_at_worker
               - --xla_tpu_enable_data_parallel_all_reduce_opt=true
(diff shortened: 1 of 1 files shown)


xpk.py
```diff
@@ -2624,6 +2624,107 @@ def run_gke_cluster_create_command(
   return 0
 
 
+def run_gke_pathways_cluster_update_command(args) -> int:
```
shaurya89, line 2627: Maybe rename as `update_gke_cluster_with_clouddns`

xpk.py
```diff
@@ -2624,6 +2624,107 @@ def run_gke_cluster_create_command(
...
+      ' --quiet'
+  )
+  xpk_print('Updating Pathways cluster to use Cloud DNS, may take a while!')
+  return_code = run_command_with_updates(
+      command, 'Pathways Cluster Update to enable Cloud DNS', args
+  )
+  if return_code != 0:
+    xpk_print(f'Pathways Cluster Update request returned ERROR {return_code}')
+    return 1
+  return 0
```
shaurya89, line 2652: Remove Pathways from these logs since these actions are not pathways specific but help enable a feature on the pathways stack

xpk.py
```diff
@@ -2624,6 +2624,107 @@ def run_gke_cluster_create_command(
...
+
+
+def upgrade_control_plane_version(args) -> int:
+  """Upgrade Pathways cluster's control plane version before updating Pathways nodepools to use CloudDNS.
```
shaurya89, line 2656: Please use GKE instead of Pathways throughout this funtcion since Pathways has nothing to do with GKE control plane version here

xpk.py
```diff
@@ -2624,6 +2624,107 @@ def run_gke_cluster_create_command(
...
+  if return_code != 0:
+    xpk_print(
+        "Pathways cluster's control plane version update request returned"
+        f' ERROR {return_code}'
+    )
+    return 1
+  return 0
+
+
+def upgrade_pathways_node_pools_command(args) -> int:
```
shaurya89, line 2688: Same comment as above

xpk.py
```diff
@@ -3231,6 +3332,65 @@ def create_cluster_if_necessary(
...
+      'Check if Cloud DNS is enabled in cluster describe.',
+      args,
+  )
+  if return_code != 0:
+    xpk_exit(return_code)
+  cloud_dns_matches = int(cloud_dns_matches)
+  if cloud_dns_matches > 0:
+    xpk_print('Cloud DNS is enabled on the cluster, no update needed.')
+    return 0
+  return 1
```
shaurya89, line 3359: Returning values the other way round seem more intuitive (1 means true/cluster is using clouddns and 0 otherwise)

xpk.py
```diff
@@ -3231,6 +3332,65 @@ def create_cluster_if_necessary(
...
+
+
+def update_pathways_cluster_if_necessary(args) -> int:
```
shaurya89, line 3362: Same as before: No need to mention pathways anywhere here since this is not required for the pathways functionality but can apply to any GKE cluster. Instead say `update_gke_cluster_with_clouddns_if_necessary`

shaurya89 (changes requested):
Thank you

xpk.py
Obliviour: Could return a bool instead? true / false. ints are great for error checking vs this looks like a good use of bool logic

xpk.py
```diff
@@ -3231,6 +3332,65 @@ def create_cluster_if_necessary(
...
+    args: user provided arguments for running the command.
+
+  Returns:
+    0 if successful and 1 otherwise.
+  """
+  all_clusters, return_code = get_all_clusters_programmatic(args)
+  if return_code > 0:
+    xpk_print('Listing all clusters failed!')
+    return 1
+  if args.cluster in all_clusters:
```
Obliviour, line 3373: Is it an error case `if args.cluster in all_clusters:` is not true

xpk.py
```diff
@@ -2624,6 +2624,107 @@ def run_gke_cluster_create_command(
...
+def upgrade_control_plane_version(args) -> int:
+  """Upgrade Pathways cluster's control plane version before updating Pathways nodepools to use CloudDNS.
+
+  Args:
+    args: user provided arguments for running the command.
+
+  Returns:
+    0 if successful and 1 otherwise.
+  """
+  command = (
```
Obliviour, line 2664: What version are we updating the control plane to?

Obliviour: We currently use the RAPID default release version in TPU flow, which will be the latest recommended gke version. 

If this update is run, I believe it would downgrade the gke version to something else. Can we use the same RAPID default release version?

xpk.py
```diff
@@ -5829,6 +5994,12 @@ def workload_create(args) -> None:
     0 if successful and 1 otherwise.
   """
   add_zone_and_project(args)
+
+  if args.use_pathways:
+    update_cluster_command_code = update_pathways_cluster_if_necessary(args)
+    if update_cluster_command_code != 0:
+      xpk_exit(update_cluster_command_code)
+
```
Obliviour: I don't think we should run cluster update commands in `workload_create`. This mixes the intent of the two command apis.

xpk.py
RoshaniN: Done, thanks!

xpk.py
RoshaniN: Done, thanks!

xpk.py
RoshaniN: I agree, I followed the convention followed in all of the XPK functions where 0 is successful and 1 means error.

xpk.py
RoshaniN: Updating to default version (there will be an upgrade, not a downgrade) - https://screenshot.googleplex.com/4TSa8KQT3XyZDSe

Is there a difference?

xpk.py
RoshaniN: No, it is not an error case. 

if args.cluster in all_clusters: is not true --- it means the cluster is not created and will follow the "create" flow, not the "update" flow. (Hence no error from update perspective.)

xpk.py
RoshaniN: Thanks, I agree with the rationale. 
However, I am hoping to have a one-time update to the cluster (if needed) on the first "workload create" on an existing cluster. New clusters won't need this change.

Handling this one-time at workload create would mean that users don't have to run "cluster create" command explicitly to upgrade their cluster. (Hoping for less intervention from users.)

Happy to chat offline, in case of concerns.

xpk.py
RoshaniN: Thanks, sure, can switch to boolean without issues!

xpk.py
RoshaniN: Upgrading to rapid default version, reusing some utilities in XPK, please check the diff :)

xpk.py
```diff
@@ -3231,6 +3334,76 @@ def create_cluster_if_necessary(
...
+    # If cluster is already using clouddns, no update necessary!
+    if is_cluster_using_clouddns(args):
+      return 0
+    cluster_update_return_code = update_cluster_with_clouddns(args)
+    if cluster_update_return_code > 0:
+      xpk_print('Updating GKE cluster to use CloudDNS failed!')
+      return cluster_update_return_code
+
+    # Find default rapid control plane version and update the control plane to the same.
+    server_config_return_code, gke_server_config = get_gke_server_config(args)
```
RoshaniN, line 3387: @Obliviour Imbibed your feedback on rapid version. Please check!

xpk.py
Obliviour: Thought I like how it reduces the steps on the user, I think this can provide confusion between what the commands do. I would like to avoid opening up workload create API to cluster changes including and beyond this change. It might work well for this feature but it opens up a lot of potential features that I don't think we should have. What do you think Roshani?

RoshaniN: Thanks @Obliviour, for your thoughts!
I can see a way forward where I exit from XPK workload create-pathways to ask the user to update the cluster using XPK cluster create-pathways. I can conditionally update the Pathways server and proxy addresses based on whether clouddns is enabled. However, this way needs an extra step from the user.

I would prefer to keep the current implementation, even for short term (until existing clusters are updated) and deprecate this path. This additional condition won't be needed in case of new clusters anyway.

WDYT, @shaurya89

RoshaniN: I went ahead with the approach I mentioned above. That would remove cluster update from workload create flow and prompt to user to update (only when needed - Pathways in headless mode).

Obliviour (approved):

xpk.py
```diff
@@ -2624,6 +2624,109 @@ def run_gke_cluster_create_command(
...
+
+  Returns:
+    0 if successful and 1 otherwise.
+  """
+  existing_node_pool_names, return_code = get_all_nodepools_programmatic(args)
+  if return_code != 0:
+    xpk_print('Listing all node pools failed!')
+    return return_code
+
+  # Batch execution to upgrade Pathways node pools simultaneously
```
shaurya89, line 2703: Remove pathways from the comment

xpk.py
```diff
@@ -2624,6 +2624,109 @@ def run_gke_cluster_create_command(
...
+  if return_code != 0:
+    xpk_print(
+        "GKE cluster's control plane version update request returned"
+        f' ERROR {return_code}'
+    )
+    return 1
+  return 0
+
+
+def upgrade_nodepools_with_clouddns(args, default_rapid_gke_version) -> int:
```
shaurya89, line 2688: clouddns change is only at the cluster level. This function can be renamed to just reflect a regular gke upgrade for nodepools

xpk.py
```diff
@@ -5067,6 +5248,38 @@ def get_pathways_expected_tpu_type(device_type: str) -> str:
   return pathways_expected_instance
 
 
+def get_rm_address_based_on_clouddns(args) -> str:
```
shaurya89, line 5251: Rename these methods to simply say `get_rm_address` and such. clouddns or not can be an internal detail.

shaurya89 (approved):
Just a few minor comments. OK to merge after addressing them.

xpk.py
RoshaniN: Thanks for catching this!

PR #152, opened 2024-06-05
Add logic to fail Pathways jobs on user code errors.

RoshaniN:
## Fixes / Features
- Set "--restart-on-user-code-failure" to fail JobSet on user code errors, as default for Pathways jobs. Note that max_restarts is 0 by default (unless altered for suspend/resume use-cases.)
- Pool the many Pathways workload create checks and setting into one function called ensure_pathways_workload_prerequisites()

## Testing / Documentation
Tested locally with changes
Screenshot 2024-06-05 at 11 49 05 AM

Here roshanin-pw-test3 is correct and roshanin-pw-test4 contains a typo in user code.

- [ y ] Tests pass

### DIFF
--- a/.github/workflows/build_tests.yaml
+++ b/.github/workflows/build_tests.yaml
@@ -62,7 +62,7 @@ jobs:
     - name: Wait for workload completion and confirm it succeeded
       run: python3 xpk.py workload list --cluster $TPU_CLUSTER_NAME --zone=us-central2-b --wait-for-job-completion $WORKLOAD_NAME --timeout 300
     - name: Run a Pathways workload on Ubuntu base image
-      run: python xpk.py workload create-pathways --cluster $TPU_CLUSTER_NAME --workload $PATHWAYS_WORKLOAD_NAME --docker-image='marketplace.gcr.io/google/ubuntu2004' --tpu-type=v4-8 --num-slices=2 --zone=us-central2-b --command "bash test.sh" 
+      run: python xpk.py workload create-pathways --cluster $TPU_CLUSTER_NAME --workload $PATHWAYS_WORKLOAD_NAME --docker-image='marketplace.gcr.io/google/ubuntu2004' --tpu-type=v4-8 --num-slices=2 --zone=us-central2-b --command "echo \"Hello world from a test script! \"" 
     - name: Wait for Pathways workload completion and confirm it succeeded
       run: python3 xpk.py workload list --cluster $TPU_CLUSTER_NAME --zone=us-central2-b --wait-for-job-completion $PATHWAYS_WORKLOAD_NAME --timeout 300
     - name: List out the workloads on the cluster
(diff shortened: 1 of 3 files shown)


.github/workflows/build_tests.yaml
```diff
@@ -62,7 +62,7 @@ jobs:
     - name: Wait for workload completion and confirm it succeeded
       run: python3 xpk.py workload list --cluster $TPU_CLUSTER_NAME --zone=us-central2-b --wait-for-job-completion $WORKLOAD_NAME --timeout 300
     - name: Run a Pathways workload on Ubuntu base image
-      run: python xpk.py workload create-pathways --cluster $TPU_CLUSTER_NAME --workload $PATHWAYS_WORKLOAD_NAME --docker-image='marketplace.gcr.io/google/ubuntu2004' --tpu-type=v4-8 --num-slices=2 --zone=us-central2-b --command "bash test.sh"
```
Obliviour, line 65: Hmm not following why we are not using the test.sh script?

Obliviour: ah since it is using a --docker-image approach not the ---base-docker-image approach. Surprised actually this test worked in the past interesting!

xpk.py
```diff
@@ -7216,6 +7229,16 @@ def add_shared_workload_create_optional_arguments(args_parsers):
             ' the workload.'
         ),
     )
+    custom_parser.add_argument(
+        '--restart-on-user-code-failure',
+        action='store_true',
+        help=(
+            'Adding this argument will return user failures back to the jobset'
+            ' manager allowing restarts on user code when --max-restarts is set'
+            ' greater than 0. By default, this is not enabled, and workloads'
```
Obliviour, line 7238: nit to add that this is enabled by default in pathways workloads

.github/workflows/build_tests.yaml
RoshaniN: I think earlier this test was not failing with user workload failure, which a good confirmation that the current fix works.

xpk.py
RoshaniN: Okay, sure! Thanks

.github/workflows/build_tests.yaml
Obliviour: Ah I see, thanks for helping me understand!

Could you also add the flag to other workloads that are created in the test? I think those will face the same problem potentially?

shaurya89 (approved):
Thanks. A "for future" comment, let's revisit all the xpk args that are not supported by pathways like --debug-dump-gcs, --scheduler etc. and make sure we have bugs for all of those that can be supported with some work.

RoshaniN:
> Thanks. A "for future" comment, let's revisit all the xpk args that are not supported by pathways like --debug-dump-gcs, --scheduler etc. and make sure we have bugs for all of those that can be supported with some work.



Yes, thanks for noticing this - I have a backlog item to evaluate these args for Pathways here - b/342189448

Obliviour (approved):

PR #153, opened 2024-06-05
Update the CloudDNS check.

lukebaumann:
If there is an upgrad available for the cluster, additional text is printed to stderr
```
* - There is an upgrade available for your cluster(s).

To upgrade nodes to the latest available version, run
  $ gcloud container clusters upgrade <cluster>
```
That text is captured by `run_command_for_value` and a ValueError is returned when trying to convert it to an integer.

The fix is to inspect the return_code of `grep` instead of the text output of `wc -l`.

## Fixes / Features
-
-

## Testing / Documentation
Testing details.

- [ y/n ] Tests pass

### DIFF
--- a/xpk.py
+++ b/xpk.py
@@ -3346,17 +3346,14 @@ def is_cluster_using_clouddns(args) -> bool:
   command = (
       f'gcloud container clusters describe {args.cluster}'
       f' --project={args.project} --region={zone_to_region(args.zone)}'
-      ' | grep "clusterDns: CLOUD_DNS" | wc -l'
+      ' | grep "clusterDns: CLOUD_DNS"'
   )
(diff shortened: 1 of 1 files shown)


lukebaumann:
I see this is not the only place this type of check occurs.

Obliviour (approved):

Obliviour:
Thank you for the change Luke! I added you to the repo as a writer so you can create a new branch based PR. I think the forking method doesn't work with our build tests.

RoshaniN (approved):
Thanks Luke

lukebaumann:
> Thank you for the change Luke! I added you to the repo as a writer so you can create a new branch based PR. I think the forking method doesn't work with our build tests.

Can you add me again? The invite expired

PR #154, opened 2024-06-05
Update XPK to support topology-aware scheduler for GPU workloads.

yangyuwei:
## Fixes / Features
- Support topology-aware scheduler for GPU workloads

## Testing / Documentation
Manually tested two cases:
1. Use topology-aware scheduler for running MaxText at a single node on A3+.
2. Use default-scheduler for running MaxText at a single node on A3+.

- [ y ] Tests pass

### DIFF
--- a/xpk.py
+++ b/xpk.py
@@ -138,6 +138,24 @@ spec:
               {volumes}
 """
 
+gpu_scheduler_yaml = """schedulerName: {scheduler_name}
+              affinity:
+                nodeAffinity:
+                  requiredDuringSchedulingIgnoredDuringExecution:
+                    nodeSelectorTerms:
+                    - matchExpressions:
+                      - key: cloud.google.com/gke-accelerator
+                        operator: Exists
+                      - key: cloud.google.com/gke-nodepool
+                        operator: In
+                        values: [{node_pool_name}]
+              nodeSelector:
+                {accelerator_label}
+                {machine_label}
+                {autoprovisioning_args}
+              """
+
+
 gpu_workload_create_yaml = """apiVersion: jobset.x-k8s.io/v1alpha2
 kind: JobSet
 metadata:
@@ -161,8 +179,7 @@ spec:
               labels:
                 xpk.google.com/workload: {args.workload}
             spec:
-              schedulingGates:
-              - name: "gke.io/topology-aware-auto-{args.workload}"
+              {gpu_scheduler}
               priorityClassName: {args.priority}
               restartPolicy: Never
               hostNetwork: true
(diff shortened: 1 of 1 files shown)


Obliviour (commented):
Hi @yangyuwei,

Very excited for these changes, some questions:

1. Does the topology-aware-auto scheduler need to be installed on the cluster before it can be used? If so we can include that as part of xpk cluster creation steps.

2. Is there any downside to not using the topology-aware-auto scheduler? I am wondering if we should make it a first class use case in the xpk API? Basically make it the default for xpk gpu users?

yangyuwei:
Thanks Victor for the review. Please see my replies below:

> 1. Does the topology-aware-auto scheduler need to be installed on the cluster before it can be used? If so we can include that as part of xpk cluster creation steps.

Yes, that's correct. If you don't mind, I'd add that support in a followup PR, so that we can get this PR in first to support running workloads with topology-aware scheduler on GPU clusters which may or may not be created by XPK.

> 2. Is there any downside to not using the topology-aware-auto scheduler? I am wondering if we should make it a first class use case in the xpk API? Basically make it the default for xpk gpu users?

Good question. I don't think there is any downside. But topology-aware scheduler can only be used on machines which have compact placement.

There might be two options:
a. Relying on users to tell whether to use it or not (as in the current PR)
b. Adding some logic to check whether topology-aware scheduler is installed on the cluster, if so, we will use it in the workload config, otherwise, we fall back to the original way. But I discussed with grigsby@ and he said topology-aware scheduler should be able to install on either compact-placement or non-compact-placement clusters. He said they could add precheck for installation, but it's not there yet.

Therefore, we can go with option a for now. Does it make sense?

xpk.py
```diff
@@ -5709,6 +5713,37 @@ def get_autoprovisioning_node_selector_args(args) -> tuple[str, int]:
...
+    autoprovisioning_args: a string of arguments for Autoprovisioning.
+
+  Returns:
+    str: yaml containing gpu scheduler configuration
+  """
+  if args.scheduler == 'gke.io/topology-aware-auto':
+    gpu_scheduler = f"""schedulingGates:
+              - name: "{args.scheduler}-{args.workload}"
+              """
+  else:
```
Obliviour, line 5733: Can we change this case to: elif args.scheduler == 'default-scheduler'

and add an else case with an error message? The error message can say x, y and supported cases. This would help in case a user has a typo. If a current user made a typo 'gke.io/topology-aware-autoOOO' then it would use the default scheduler?

Obliviour (approved):
LGTM, if we find that the topology-aware scheduler gives better performance, we should make that the default case for GPU workloads and have XPK install it as well. This can be in future PRs.

Thanks Yuwei for the change!

xpk.py
yangyuwei: Sounds good. I've updated the code and tested it manually. Since you are OOO, I'll go ahead and merge the PR to unblock the XLML tests depending on it.

PR #168, opened 2024-08-08
Create `parser` package. Move logic from `xpk.py` to `parser` package. 

PBundyra:
## Fixes / Features
- Create `parser` package
- Initialize `core` package
- Move logic from `xpk.py` to `parser` package. 
- Wrap some of the parser logic into new functions e.g. `set_parser`
- Adjust `pyproject.toml` accordingly

## Testing / Documentation
Tested manually

- [ y ] Tests pass
- [ n ] Appropriate changes to documentation are included in the PR - no user-facing changes


PBundyra:
Once I'll get LGTM I'll merge it when #166 is merged

.github/workflows/build_tests.yaml
```diff
@@ -49,31 +53,34 @@ jobs:
       run: |
         gcloud config set compute/zone us-east4-a
         gcloud config get compute/zone
+    - name: Install xpk
```
Obliviour, line 56: Oh perfect I was looking for this in the previous PR. I think we want to test both flows with python3 xpk.py and xpk (through pip)?

Maybe a build test that checks that [xpk.py / xpk] --help can be run is a good check for both flows in the build test, the rest of the build test can use just one of them. We could add a nightly test that tests our exhaustive command list?

.github/workflows/build_tests.yaml
```diff
@@ -49,31 +53,34 @@ jobs:
...
-      run: python xpk.py cluster create-pathways --cluster $TPU_CLUSTER_NAME --tpu-type=v4-8  --num-slices=2 --zone=us-central2-b --default-pool-cpu-machine-type=n1-standard-16 --default-pool-cpu-num-nodes=16 --reservation='${{ secrets.GCP_TPU_V4_RESERVATION }}'
+      run: xpk cluster create-pathways --cluster $TPU_CLUSTER_NAME --tpu-type=v4-8  --num-slices=2 --zone=us-central2-b --default-pool-cpu-machine-type=n1-standard-16 --default-pool-cpu-num-nodes=16 --reservation='${{ secrets.GCP_TPU_V4_RESERVATION }}'
     - name: Authenticate Docker
       run: gcloud auth configure-docker --quiet
     - name: Create test script to execute in workloads
       run: echo -e '#!/bin/bash \n echo "Hello world from a test script!"' > test.sh
     - name: Run a base-docker-image workload
-      run: python xpk.py workload create --cluster $TPU_CLUSTER_NAME --workload $WORKLOAD_NAME  --command "bash test.sh"  --tpu-type=v4-8 --num-slices=2 --zone=us-central2-b
+      run: python main.py workload create --cluster $TPU_CLUSTER_NAME --workload $WORKLOAD_NAME  --command "bash test.sh"  --tpu-type=v4-8 --num-slices=2 --zone=us-central2-b
```
Obliviour, line 66: I propose we keep the name xpk.py as the entry point, a handful of our documentation and user workflows is build upon xpk.py and would be a large change from calling python3 xpk.py to python3 main.py.

Obliviour (commented):
Awesome refactor, hoping we can keep the name as xpk.py for user experience.

.github/workflows/build_tests.yaml
PBundyra: I see. At first I also wanted to keep it as `xpk.py` but I encountered some limitation due to python import system. A package cannot have the same name as module within it(in this case `xpk` and `xpk.py`) - see [second answer from this Stackoverflow post](https://stackoverflow.com/questions/54333865/python-no-module-named-error-package-is-not-a-package). At least, this is what happens when running xpk directly using python interpreter. It works correctly when installing first with pip, and then running it as a command. To solve this we have a few alternatives:
### Introduce `main.py` with drawbacks you mentioned
I agree that it's important to be consistent with our documentation 
### Recommend always installing xpk with pip first and discourage running it directly using python interpreter
I like this approach the most. I believe it's a standard practice to use it that way, and it's what other tools do. So from users' perspective, they would need to use it following `README.md`
1. Run `pip install xpk`
2. Run `xpk [subcommand]` 

However, I may not be aware of some cons you are - maybe it would break some customers' workflows?

Another advantage of this approach is that it would be easier to maintain the project, because of simplified e2e tests - we would need to cover only one instead of two ways of installing XPK

### Change name of the package `xpk`
Cons: I believe it's common practice to name packages the same as the project in Python, so we wouldn't be following the convention.


Please let me know what you think

.github/workflows/build_tests.yaml
PBundyra: Please see my comment below. I believe it's easier to maintain just xpk through pip, and stop supporting using it through python interpreter - unless there are some cons I'm not aware of (maybe some users already use it with the python interpreter and it would break their workflows(?))

.github/workflows/build_tests.yaml
Obliviour: We have most users using xpk through python3 xpk.py so changing that flow would require a lot of user awareness :( 

No good answer currently

xpk/core/core.py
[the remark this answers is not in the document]
PBundyra: >> Do you think if we adjust some naing of the files/folders we can eliminate confusion? There are two files and a folder/package with the name of "core".

Looking at different Python projects (e.g. [JAX](https://github.com/google/jax/tree/main/jax)) I believe it's fine to have two (or even more) files/package with the same name. 

>> I'm new to python but is it a good practice to have a separate module for the only core.py?

You're absolutely right. This is only temp solution until I divide `core.py` into smaller modules. I didn't want to make a huge refactor at once because it would be harder to review and iterate. My goal is to have modules for at least `cluster.py`, `workload.py` and `inspector.py` - similarly to `parser` package, maybe even more, and a separate file/folder for the constants.

.github/workflows/build_tests.yaml
PBundyra: I see, let me investigate a little further then

PBundyra: I came up with [the solution](https://github.com/google/xpk/pull/166), PTAL

Obliviour: Thanks @PBundyra LGTM

xpk/core/core.py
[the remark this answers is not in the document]
Obliviour: Can even have a file for the template jobset / kueue configs.

There may be some common core code between cluster.py, workload.py and inspector.py as well. Inspector logic reuses logic from workload list for example.

Obliviour (approved):
Thank you for the refactor changes, nice way to start work in XPK!

PR #175, opened 2024-08-19
Create `commands` package and `core/` modules for NAP, Kueue and Pathways

PBundyra:
## Fixes / Features
- Creates `commands` package
- Creates `core/` modules for NAP, Kueue and Pathways
- Clean up

## Testing / Documentation
Tested manually

- [ y ] Tests pass
- [ n ] Appropriate changes to documentation are included in the PR - no user-facing changes

### DIFF
--- a/src/xpk/core/core.py
+++ b/src/xpk/core/core.py
@@ -39,14 +39,19 @@ import re
 import string
 import subprocess
 import sys
-import time
-from .. import utils as xpk_utils
 from dataclasses import dataclass
+
+from .. import utils as xpk_utils
+from .commands import (
+    run_command_for_value,
+    run_command_with_updates,
+    run_command_with_updates_retry,
+    run_commands,
+)
 from .system_characteristics import (
     AcceleratorType,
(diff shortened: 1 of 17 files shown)


src/xpk/core/core.py
Obliviour, line 1: This one looks like the hardest to split mainly since there are a lot of features that are somewhat common but also maybe unique to different flows. This looks good to me, I am guessing that we will put most general features into this file. It also sounds that splitting up past this may get confusing to users / devs?

Obliviour (commented):
LGTM, could you add into the readme or somewhere that describes how the code is split generally to help new people / this transition to the new split of code?

Obliviour (approved):

src/xpk/core/core.py
PBundyra: Absolutely, I'll add the README with the code structure.

My main goal was to create some base structure we can work on in the future. Right now `core/core.py` is still a big file but it's much smaller than it used to be, which was the priority for me. There's definitely room for improvement, but I believe we'll improve the codebase gradually.

PR #177, opened 2024-08-21
Enabling Workload Identity and GCSFuse driver flags added.

sharabiani:
## Fixes / Features
- `--enable-workload-identity` flag implemented for `cluster create' command. It enables Workload Identity Federation on the cluster and node-pools.
- `--enable-gcsfuse-csi-driver` flag implemented for `cluster create' command. It enables GSCFuse driver on the cluster. This also enables Workload Identity Federation.'

## Testing / Documentation
Manual tests.

- [ y/n ] Tests pass

### DIFF
--- a/src/xpk/commands/cluster.py
+++ b/src/xpk/commands/cluster.py
@@ -90,7 +90,7 @@ def cluster_create(args) -> None:
     xpk_exit(create_cluster_command_code)
 
   # Enable WorkloadIdentity if not enabled already.
-  if args.enable_workload_identity or args.enable_gcsfuse_driver:
+  if args.enable_workload_identity or args.enable_gcsfuse_csi_driver:
     update_cluster_command_code = (
         update_cluster_with_workload_identity_if_necessary(args)
     )
@@ -98,7 +98,7 @@ def cluster_create(args) -> None:
       xpk_exit(update_cluster_command_code)
 
   # Enable GCSFuse CSI Driver if not enabled already.
-  if args.enable_gcsfuse_driver:
+  if args.enable_gcsfuse_csi_driver:
     update_cluster_command_code = (
         update_cluster_with_gcsfuse_driver_if_necessary(args)
     )
@@ -491,7 +491,7 @@ def run_gke_cluster_create_command(
           f' --cluster-dns-domain={args.cluster}-domain'
       )
 
-  if args.enable_workload_identity or args.enable_gcsfuse_driver:
+  if args.enable_workload_identity or args.enable_gcsfuse_csi_driver:
     command += f' --workload-pool={args.project}.svc.id.goog'
 
   if args.enable_gcsfuse_csi_driver:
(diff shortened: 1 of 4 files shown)


src/xpk/commands/cluster.py
```diff
@@ -87,6 +89,22 @@ def cluster_create(args) -> None:
   if create_cluster_command_code != 0:
     xpk_exit(create_cluster_command_code)
 
+  # Enable WorkloadIdentity if not enabled already.
+  if args.enable_workload_identity or args.enable_gcsfuse_driver:
```
Obliviour, line 93: Nit: curious if you want to check for the case `if not args.enable_workload_identity and args.enable_gcsfuse_driver: ... return error`

src/xpk/parser/cluster.py
```diff
@@ -449,6 +449,21 @@ def add_shared_cluster_create_optional_arguments(args_parsers):
             ' --custom-tpu-nodepool-arguments="--enable-ip-alias"'
         ),
     )
+    custom_parser.add_argument(
+        '--enable-workload-identity',
```
Obliviour, line 453: One discussion we can think through is if we want to support workload-identity separately from gcsfuse. Do we see a need to support workload-identity by itself such as the additional dev support of the feature by itself in XPK? The alternative is we just support the current request (from what I am aware) is just gcsfuse-csi-driver support. If we don't currently see value for `--enable-workload-identity` flag by itself, we don't need to add it now (one way door)! Curious what you think!

src/xpk/core/core.py
```diff
@@ -1406,6 +1617,35 @@ def run_gke_node_pool_create_command(
       xpk_print(f'Delete Nodepools returned ERROR {max_return_code}')
       return 1
 
+  # Enable Workload Identity on existing Nodepools
```
Obliviour, line 1620: nit to add comment saying the xpk doesn't support disabling WI on clusters that have it enabled already? (Which makes sense, not saying we should add this)

Obliviour (approved):
LGTM, great changes! I have some questions about --enable-workload-identity support by itself and some nits!

src/xpk/commands/cluster.py
sharabiani: does false value for args.enable_workload_identity means user explicitly set it to false? or it can also mean user did not provided the flag?
I assumed the latter is the case, and we'll automatically enable WI if gsfuse is enabled.
What do you think?

src/xpk/parser/cluster.py
sharabiani: That's what I'm waiting for the decision. For now it's possible to only provide only --enable-gcsfuse-csi-driver flag which will enable WI, too.
We can remove the WI flag later if we no need for that. What do you think?

src/xpk/core/core.py
sharabiani: done

PBundyra:
LGTM

src/xpk/parser/cluster.py
Obliviour: > We can remove the WI flag later if we no need for that. What do you think?

Yes I agree we can remove it, though I would prefer to remove it now since it is harder to remove later! But I'll leave the decision up to you here :)

sharabiani: I'll leave it to @PBundyra as we discussed and close this PR.

PR #181, opened 2024-08-28
Fix debug logging

Obliviour:
## Fixes / Features
- Debug logs were missing, move to exporting each variable per line

## Testing / Documentation
Ran tests with and without --enable-debug-logs

- [ y ] Tests pass


RoshaniN (commented):
Thanks for the change, Victor!
Do you think the debug changes can be separated from the System Characteristic changes?

Obliviour:
Moved it to https://github.com/google/xpk/pull/183 due to conflicting change. Found it easier to just move the change.

PR #184, opened 2024-09-08
better core dump for debugging

ZhiyuLi-goog:
## Features
upload more detailed info to gcs bucket in failures

## Testing / Documentation

- [x] Verified in burnin tests

### DIFF
--- a/src/xpk/core/core.py
+++ b/src/xpk/core/core.py
@@ -2104,8 +2104,12 @@ def get_main_container(args, system, docker_image, resource_type) -> str:
         ' is required but not installed. Aborting"; exit 24;};'
     )
     xpk_internal_commands += (
+        'set -x;'
         'WORKER_ID=$HOSTNAME;'
-        f'gsutil -m cp -r /tmp/xla_dump/ {args.debug_dump_gcs}/$WORKER_ID;'
+        f'gcloud storage cp -r /tmp/xla_dump/ {args.debug_dump_gcs}/$WORKER_ID;'
+        f'gcloud storage cp /tmp/*.dump {args.debug_dump_gcs}/$WORKER_ID;'
+        f'gcloud storage cp -r /tmp/tpu_logs {args.debug_dump_gcs}/$WORKER_ID;'
+        'sleep 120;'
     )
 
   command = args.command


src/xpk/core/core.py
```diff
@@ -2104,8 +2104,12 @@ def get_main_container(args, system, docker_image, resource_type) -> str:
         ' is required but not installed. Aborting"; exit 24;};'
     )
     xpk_internal_commands += (
+        'set -x;'
         'WORKER_ID=$HOSTNAME;'
-        f'gsutil -m cp -r /tmp/xla_dump/ {args.debug_dump_gcs}/$WORKER_ID;'
+        f'gcloud storage cp -r /tmp/xla_dump/ {args.debug_dump_gcs}/$WORKER_ID;'
+        f'gcloud storage cp /tmp/*.dump {args.debug_dump_gcs}/$WORKER_ID;'
+        f'gcloud storage cp -r /tmp/tpu_logs {args.debug_dump_gcs}/$WORKER_ID;'
+        'sleep 120;'
```
Obliviour, line 2112: What is the sleep for?

src/xpk/core/core.py
```diff
@@ -2104,8 +2104,12 @@ def get_main_container(args, system, docker_image, resource_type) -> str:
         ' is required but not installed. Aborting"; exit 24;};'
```
Obliviour, line 2104: I don't think you need the gsutil_test_command anymore since you move to gcloud storage

src/xpk/core/core.py
```diff
@@ -2104,8 +2104,12 @@ def get_main_container(args, system, docker_image, resource_type) -> str:
...
     )
     xpk_internal_commands += (
+        'set -x;'
         'WORKER_ID=$HOSTNAME;'
-        f'gsutil -m cp -r /tmp/xla_dump/ {args.debug_dump_gcs}/$WORKER_ID;'
+        f'gcloud storage cp -r /tmp/xla_dump/ {args.debug_dump_gcs}/$WORKER_ID;'
+        f'gcloud storage cp /tmp/*.dump {args.debug_dump_gcs}/$WORKER_ID;'
```
Obliviour, line 2110: Do we need additional flags set to find the /tmp/*.dump and /tmp/tpu_logs? just want to make sure these additional dumps always exist and don't crash a users job if they can't be found (like only available with additional flags)

src/xpk/core/core.py
ZhiyuLi-goog: Good point. Let me change this part accordingly.

src/xpk/core/core.py
ZhiyuLi-goog: 1. No flags for dumping /tmp/tpu_logs.   
2. As for `/tmp/*.dump`, this file will be dumped in case of some specific hardware failure. We can also force core dump with some flags as well.

There will be an error without finding the specific path. And the pipeline will still continue after the error since we didn't `set -e;`.

src/xpk/core/core.py
ZhiyuLi-goog: This is a workaround, not an ideal solution. The main host pod requires a significant amount of time to upload `/tmp/*.dump` which are approximately 10 GiB in size.
I did find the main host pod was terminated earlier before finishing upload. I just want to add a buffer time to each pod's lifecycle, ensuring the host pod has sufficient time to finish uploading.

Let me know if you have better solution.

src/xpk/core/core.py
```diff
@@ -2104,8 +2104,12 @@ def get_main_container(args, system, docker_image, resource_type) -> str:
         ' is required but not installed. Aborting"; exit 24;};'
     )
     xpk_internal_commands += (
+        'set -x;'
```
Obliviour, line 2107: Curious why we add `set -x;` here? Is it related to the debug_dump_gcs feature or a more general feature? If it is a more general feature, maybe it should be part of --enable-debug-logs arg?

src/xpk/core/core.py
Obliviour: > We can also force core dump with some flags as well.

SG, Should we explicitly set the flags for the core dumps?

src/xpk/core/core.py
Obliviour: Hmm buffer times are scary to put into production code because you don't know if you have set the time for too long or too short!

It looks like `gcloud storage cp ..` a blocking call? Why would the main host terminate earlier is strange... do you mind looking into this more? Thank you Zhiyu!

PR #185, opened 2024-09-17
Fix GKE node version selection logic

44past4:
## Fixes / Features

- Fixes problem with GKE nodes version selection if the current GKE cluster version is no longer a valid GKE rapid version.

## Testing / Documentation

Tested manually

- [ y/n ] Tests pass

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,17 @@ and the following GPU types:
 and the following CPU types:
 * n2-standard-32
 
+# Permissions needed on Cloud Console:
+
+Artifact Registry Writer
+Compute Admin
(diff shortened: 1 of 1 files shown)


Obliviour (approved):
LGTM, Is it possible that the rapid release will not have a release of the same major version or lower? The current default rapid release version is 1.31 and it is currently  and has support for versions old as 1.27x. 1.27x was introduced in May 2023. So I am guessimating that a major version will be in the rapid release for 1.5 years. If a user doesn't upgrade their cluster for that long then, this logic may break again.

PR #187, opened 2024-09-18
Fix autoprovisioning with spot nodes

avrittrohwer:
## Fixes / Features
- Fixes workload rendering when using spot, without this change xpk workload create errors like:

  ```
  [XPK] Waiting for `Creating Workload`, for 0 seconds
  error: error parsing /tmp/tmp242uhnfs: error converting YAML to JSON: yaml: line 33: could not find expected ':'
  [XPK] Task: `Creating Workload` terminated with code `1`
  ```
  
- Adds required pod tolerations when using node auto-provisioning with spot nodes.  Without the tolerations cluster autoscaler will not create new spot node pools.

### Node auto-provisioning with spot

1. Created a xpk cluster with `--spot` flag.
2. Created a workload with a different topology than the cluster default.
3. Observed a nodepool being created with the new workload topology using spot TPU nodes.

### Node auto-provisioning without spot

TODO

### Not auto-provisioning

TODO

- [ y/n ] Tests pass
- [ y, not needed ] Appropriate changes to documentation are included in the PR

### DIFF
--- a/src/xpk/core/nap.py
+++ b/src/xpk/core/nap.py
@@ -240,6 +240,29 @@ def create_autoprovisioning_config(
   return autoprovisioning_config, 0
 
 
+def get_cluster_metadata_configmap(args) -> tuple[dict, int]:
+  """Gets the cluster metadata configmap.
+
+  Args:
+    args: user provided arguments for running the command.
+
+  Returns:
+    configmap and 0 if found, None and 1 otherwise.
+  """
+  configmap = get_cluster_configmap(
+      args, f'{args.cluster}-{CLUSTER_METADATA_CONFIGMAP}'
+  )
+  if configmap is None:
+    xpk_print(
+        'Unable to find config map. Please specify a capacity type'
+        ' --on-demand, --spot, --reservation=$RESERVATION_ID) to continue'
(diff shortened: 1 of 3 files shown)


avrittrohwer:
`zone: 'us-central2'>] finished with error: Try a different location, or try again later: Google Compute Engine does not have enough resources available to fulfill request: us-central2-b`

src/xpk/core/nap.py
```diff
@@ -240,6 +240,29 @@ def create_autoprovisioning_config(
   return autoprovisioning_config, 0
 
 
+def get_cluster_metadata_configmap(args) -> tuple[dict, int]:
```
PBundyra, line 243: What's the benefit of this function with respect to `get_cluster_configmap`?

avrittrohwer: It reduces duplication in `get_autoprovisioning_node_selector_args` and `get_capacity_type_str_from_args_or_cluster_default`.  I don't have strong opinion here, happy to inline the func

PBundyra: Maybe instead of inlining we could just use `get_cluster_configmap`? I would prefer not to add functions which main functionality is calling a different function with a const argument

PR #192, opened 2024-09-25
Introduce Storage API

PBundyra:
## Fixes / Features
- Creates Storage CRD that allows attaching existing GCS Bucket to workloads

## Testing / Documentation
Tested manually, CI/CD tests in progress

- [ y ] Tests pass

### DIFF
--- a/src/xpk/api/storage_crd.yaml
+++ b/src/xpk/api/storage_crd.yaml
@@ -1,0 +1,52 @@
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  name: storages.xpk.x-k8s.io
+spec:
+  group: xpk.x-k8s.io
+  versions:
+    - name: v1
+      served: true
+      storage: true
+      schema:
+        openAPIV3Schema:
+          type: object
+          properties:
+            spec:
+              type: object
+              properties:
+                type:
+                  type: string
+                cluster:
+                  type: string
+                auto_mount:
+                  type: boolean
+                mount_point:
+                  type: string
+                readonly:
+                  type: boolean
+                manifest:
+                  type: string
+                pv:
+                  type: string
+                pvc:
+                  type: string
+              required:
+                - type
+                - cluster
+                - auto_mount
+                - mount_point
+                - readonly
+                - manifest
+                - pvc
+                - pv
+          x-kubernetes-validations:
+            - message: Value is immutable
+              rule: self == oldSelf
+  scope: Cluster
+  names:
+    plural: storages
+    singular: storage
+    kind: Storage
+    shortNames:
+      - stg
--- a/src/xpk/commands/storage.py
+++ b/src/xpk/commands/storage.py
@@ -17,7 +17,7 @@ limitations under the License.
 from argparse import Namespace
 
 from kubernetes import client as k8s_client
-from kubernetes.client.exceptions import ApiException
+from kubernetes.client.rest import ApiException
 
 from ..core.core import (
     setup_k8s_env,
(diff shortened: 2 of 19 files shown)


PBundyra:
PTAL @sharabiani @44past4

src/xpk/api/storage_crd.yaml
```diff
@@ -0,0 +1,52 @@
+apiVersion: apiextensions.k8s.io/v1
```
sharabiani, line 1: A general question: I'm new to k8s world, Do you think a configMap can be a better fit for this reason especially as we are not going to have a controller for the objects created based on this crd? What are pros and cons of using a CRD over ConfigMap?
@44past4

src/xpk/commands/storage.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+def storage_delete(args: Namespace) -> None:
+  k8s_api_client = setup_k8s_env(args)
+  install_storage_crd(k8s_api_client)
+  api_instance = k8s_client.CustomObjectsApi(k8s_api_client)
+  core_api = k8s_client.CoreV1Api()
+  try:
+    storage = get_storage(k8s_api_client, args.name)
+    core_api.delete_namespaced_persistent_volume_claim(storage.pvc, "default")
```
sharabiani, line 69: The order or deletion is correct now. 
- Is it guaranteed that if any failure happens during  the deletion of this objects we will get an exception? I'm not sure about this particular case, but for example when you "kubectl apply -f aDeployment.yaml", you get the okay that the request is sent, but the deployment may fail while the controller is provisioning the deployment.

- Let's say pvc and pv were deleted successfully, but custom_object failed. What's the behavior when retrying to delete the already deleted pv and pvc?

src/xpk/api/storage_crd.yaml
sharabiani: If we decicded to go ahead with a CRD, do you think we can have a version for it so that if later we wanted to add new fields to it we can distinguish crd/objects versions there?

Let's say a CRD is created by a user with an older version of xpk, but upgrades xpk to a newer version that introduces new fields there. What will be the behavior?

src/xpk/core/storage.py
```diff
@@ -0,0 +1,340 @@
...
+  Args:
+      k8s_api_client: An ApiClient object for interacting with the Kubernetes API.
+      names: A list of Storage resource names to retrieve.
+
+  Returns:
+      A list of Storage objects matching the given names and any auto-mounted storages.
+  """
+  storages: list[Storage] = []
+  for storage in list_storages(k8s_api_client):
+    if storage.name in names:
```
sharabiani, line 197: If "names" is the list of storages requested by users to mount to the workload, here we are ignoring the ones that are not in the list_storages. This can be a problematic behavior because user might think all the storage they requested are mounted.

We need to throw an error if any requested storage doesn't exist and let the user know the name of invalid storages. Maybe there is a typo there and they can fix it.

src/xpk/core/storage.py
```diff
@@ -0,0 +1,340 @@
...
+      A list of Storage objects that have `auto_mount` set to True.
+  """
+  auto_mount_storages: list[Storage] = []
+  for storage in list_storages(k8s_api_client):
+    if storage.auto_mount is True:
+      auto_mount_storages.append(storage)
+  return auto_mount_storages
+
+
+def get_storages(k8s_api_client: ApiClient, names: list[str]) -> list[Storage]:
```
sharabiani, line 184: The naming for this method can be more descriptive. We have a method `get_storage` that retrieves a single storage. Therefore, we expect if get_storages retrieves the list of storage objects based on names list passed in, but it also add some other "auto_mount" objects here.

I would suggest to
- keep this `get_storages` method for the names only.
- add a new method `get_storages_to_mount` that unions the data from `get_storages` and `get_auto_mount_storages`

sharabiani (commented):
src/xpk/templates/pod.yaml is an empty file

src/xpk/core/storage.py
PBundyra: Good point, thanks. Done!

src/xpk/core/storage.py
PBundyra: Good point, thanks. Done!

src/xpk/commands/storage.py
PBundyra: > Is it guaranteed that if any failure happens during the deletion of this objects we will get an exception? I'm not sure about this particular case, but for example when you "kubectl apply -f aDeployment.yaml", you get the okay that the request is sent, but the deployment may fail while the controller is provisioning the deployment.

I couldn't find an exhausting list of Exceptions that can be throw by any of these methods, but by looking at examples provided in the k8s python client documentation I deduced they can only throw ApiExceptions

References:
- https://github.com/kubernetes-client/python/blob/master/kubernetes/docs/CoreV1Api.md#delete_persistent_volume
- https://github.com/kubernetes-client/python/blob/master/kubernetes/docs/CoreV1Api.md#delete_namespaced_persistent_volume_claim
- https://github.com/kubernetes-client/python/blob/master/kubernetes/docs/CustomObjectsApi.md#delete_cluster_custom_object

> Let's say pvc and pv were deleted successfully, but custom_object failed. What's the behavior when retrying to delete the already deleted pv and pvc?

Good catch. Currently if we tried to retry we would get 404 error and the storage wouldn't be deleted. Let me fix that

PBundyra: Done, PTAL @sharabiani

src/xpk/api/storage_crd.yaml
PBundyra: WDYT @44past4?

44past4: The main pros of using CRDs over ConfigMaps are well defined structure, validation and documentation provided by CRD. Therefore in this case where we want to reuse the information about the storages in multiple places (listing storages, mounting them to workloads etc.) for me it makes sense to use CRD.

When it comes to versioning the definition below uses v1 version. As long as we will not be adding/removing required fields to CRD we should be able to keep v1. If we will need to add or remove some required fields we will need to bump the version to v2.

sharabiani: Thanks Pawel!

sharabiani (approved):

src/xpk/commands/storage.py
```diff
@@ -0,0 +1,112 @@
...
+    list_storages,
+    print_storages_for_cluster,
+)
+from ..utils import apply_kubectl_manifest, xpk_exit, xpk_print
+
+
+def storage_create(args: Namespace) -> None:
+  k8s_api_client = setup_k8s_env(args)
+
+  install_storage_crd(k8s_api_client)
```
44past4, line 43: I wonder if it wouldn't be better to have this only on cluster create instead of running it with each storage subcommand.

src/xpk/core/storage.py
```diff
@@ -0,0 +1,365 @@
...
+      "NAME",
+      "TYPE",
+      "AUTO MOUNT",
+      "MOUNT POINT",
+      "READONLY",
+      "MANIFEST",
+  ]
+  storage_tab = []
+  for storage in storages:
+    if storage.cluster == cluster:
```
44past4, line 303: How can we have storages for different clusters here?

src/xpk/parser/storage.py
```diff
@@ -0,0 +1,121 @@
...
+  add_shared_arguments(req_args)
+  req_args.add_argument(
+      'name',
+      type=str,
+      help='The name of storage',
+  )
+  req_args.add_argument(
+      '--type',
+      type=str,
+      help='The type of storage',
```
44past4, line 60: I would add information with supported values here.

src/xpk/commands/storage.py
```diff
@@ -0,0 +1,112 @@
...
+  return_code = update_cluster_with_workload_identity_if_necessary(args)
+  if return_code > 0:
+    xpk_exit(return_code)
+  return_code = update_cluster_with_gcsfuse_driver_if_necessary(args)
```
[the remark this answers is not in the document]
PBundyra: Thanks, done

src/xpk/core/storage.py
PBundyra: That's a valid point, thanks, done

src/xpk/parser/storage.py
PBundyra: Added validation as well

src/xpk/commands/storage.py
PBundyra: I agree, done

44past4 (approved):

PR #194, opened 2024-09-26
Update the CloudDNS check.

lukebaumann:
If there is an upgrad available for the cluster, additional text is printed to stderr

```
* - There is an upgrade available for your cluster(s).

To upgrade nodes to the latest available version, run
  $ gcloud container clusters upgrade <cluster>
```

That text is captured by run_command_for_value and a ValueError is returned when trying to convert it to an integer.

The fix is to inspect the return_code of grep instead of the text output of wc -l.

## Fixes / Features
-
-

## Testing / Documentation
Testing details.

- [ y/n ] Tests pass


RoshaniN (approved):
Thanks for the change!

Obliviour (approved):

PR #200, opened 2024-10-09
Implement info command

pawloch00:
## Fixes / Features
- implement xpk info command which lists clusterqueues and localqueues in cluster

## Testing / Documentation
Testing details.

- [ y ] Tests pass


README.md
```diff
@@ -63,6 +63,13 @@ and the following CPU types:
 * Storage Admin
 * Vertex AI Administrator
 
+# Prerequisites
```
PBundyra, line 66: Please add information about the required version of Python and `gcloud`

pawloch00: https://github.com/AI-Hypercomputer/xpk/pull/204

src/xpk/commands/info.py
```diff
@@ -0,0 +1,219 @@
...
+
+  prepare_kueuectl(args)
+
+  lqs = run_kueuectl_list_localqueue(args)
+  cqs = run_kueuectl_list_clusterqueue(args)
+
+  aggregate_results(cqs, lqs)
+
+
+def apply_shared_flags(args: Namespace) -> None:
```
PBundyra, line 80: Use `add_zone_and_project()` instead

src/xpk/commands/info.py
```diff
@@ -0,0 +1,219 @@
...
+
+  xpk_print(
+      '\n', tabulate(cq_usages + lq_usages, headers='keys', tablefmt=table_fmt)
+  )
+
+
+def parse_queue_lists(
+    qs: list[dict],
+    usage_key: Optional[str] = 'flavorUsage',
+    reservation_key: Optional[str] = 'flavorsReservation',
```
PBundyra, line 126: I believe the dividend for this equation should be ClusterQueue.status.flavorsReservation and the divisor should be the nominal quota in ClusterQueue.spec

src/xpk/commands/info.py
```diff
@@ -0,0 +1,219 @@
...
+    None
+  """
+  cq_list = json.loads(cqs)['items']
+  lq_list = json.loads(lqs)['items']
+
+  cq_usages = parse_queue_lists(cq_list, usage_key='flavorsUsage')
+  lq_usages = parse_queue_lists(lq_list)
+
+  xpk_print(
+      '\n', tabulate(cq_usages + lq_usages, headers='keys', tablefmt=table_fmt)
```
PBundyra, line 119: I believe users could benefit more it there was a clear distinction between ClusterQueues and LocalQueues in the table, currently this mixes up two different types of queues that have different properties. WDYT @44past4 ?

src/xpk/commands/info.py
pawloch00: add_zone_and_project only retrieve values from gcloud. If we want to set different values for invoking info command, we should apply gcloud commands. Correct me if I am wrong.

PBundyra: `add_zone_and_project()` retrives these values only if they are not set by the user. In other words, if a user specifies these values, the method doesn't change them. However it comes handy if a user doesn't specify them. That way a user doesn't need to worry about setting these values in every xpk invocation.

README.md
```diff
@@ -1099,5 +1099,14 @@ To explore the stack traces collected in a temporary directory in Kubernetes Pod
   xpk-test --tpu-type=v5litepod-16 --deploy-stacktrace-sidecar
  ```
 
+### Get information about jobs, queues and resources.
+
+To list available resources and queues use ```xpk info``` command. It allows to see localqueues and clusterqueues and check for available resources.
+
+To see clusterqueues with reservations and usage:
```
PBundyra, line 1106: Do you mean available resources and queue in general?

src/xpk/commands/info.py
```diff
@@ -0,0 +1,240 @@
...
+    add_zone_and_project,
+)
+import json
+from tabulate import tabulate
+from argparse import Namespace
+
+table_fmt = 'plain'
+
+
+def prepare_kueuectl(args: Namespace) -> None:
```
PBundyra, line 33: WDYT about moving this function to `core/kueue.py`? It seems like it could be used outside of the `xpk info` command, similarly to `verify_kueuectl_installation`

src/xpk/commands/info.py
```diff
@@ -0,0 +1,241 @@
...
```
PBundyra, line 33: I think this is redundant now

src/xpk/commands/info.py
```diff
@@ -233,6 +222,8 @@ def run_kueuectl_list_clusterqueue(args: Namespace) -> str:
     kueuectl localqueue formatted as json string
   """
   command = 'kubectl kueue list clusterqueue -o json'
+  if args.namespace != '':
+    command += f' --namespace {args.namespace}'
```
PBundyra, line 226: I wonder if this is necessary since ClusterQueue is an cluster-scoped object so it does not belong to any namespace

src/xpk/commands/info.py
```diff
@@ -233,6 +222,8 @@ def run_kueuectl_list_clusterqueue(args: Namespace) -> str:
     kueuectl localqueue formatted as json string
```
PBundyra, line 222: I believe you meant clusterqueue

PBundyra (dismissed):
Good job, thanks!

src/xpk/core/kueue.py
```diff
@@ -140,6 +141,43 @@
...
+  Args:
+    args: user provided arguments.
+  Returns:
+    None
+  """
+  xpk_print('Veryfing kueuectl installation')
+
+  verify_kueuectl_installed_code = verify_kueuectl_installation(args)
+  if verify_kueuectl_installed_code == 0:
+    xpk_print('kueuectl installed')
```
sharabiani, line 155: `kueuectl found`

src/xpk/core/kueue.py
```diff
@@ -140,6 +141,43 @@
...
+
+  if verify_kueuectl_installed_code != 0:
+    xpk_print(
+        'kueuectl not installed. Please follow'
```
sharabiani, line 159: `kueuectl not found. ....`

src/xpk/core/kueue.py
```diff
@@ -140,6 +141,43 @@
...
+        ' https://kueue.sigs.k8s.io/docs/reference/kubectl-kueue/installation/'
+        ' to install kueuectl.'
+    )
+    xpk_exit(verify_kueuectl_installed_code)
+
+
+def verify_kueuectl_installation(args: Namespace) -> int:
+  """Verify if if kueuectl is installed.
```
sharabiani, line 167: there are double `if`

src/xpk/commands/info.py
```diff
@@ -0,0 +1,231 @@
...
+)
+import json
+from tabulate import tabulate
+from argparse import Namespace
+
+table_fmt = 'plain'
+
+
+def info(args: Namespace) -> None:
+  """Provide info about localqueues, clusterqueues and their resources.
```
sharabiani, line 34: Localqueue, Clusterqueue, Flavour, ...  are kueue concepts. Is telling the user about these concepts in line with the Experience we are working on now? or are we going to translate them to something familiar to people coming from the other worlds?

This comment is probably concerns PMs. I ask because we are installing/configuring kueue on the cluster in the background and I'm wonder if we should assume users already have the knowledge about kueue.

src/xpk/core/kueue.py
```diff
@@ -140,6 +141,43 @@
 """
 
 
+def prepare_kueuectl(args: Namespace) -> None:
```
sharabiani, line 144: I think this method is a leftover from before when you were installing kueuectl because now it doesn't seem to prepare anything and is basically just a wrapper for `verify_.." method. Maybe we can combine them?

src/xpk/commands/info.py
pawloch00: Probably shoud be named partition or sth, but I thing I will leave it as it is now, and we can come back to this topic after consultation.

sharabiani: I would bring it up in the team's chat before completing this PR and get the go-ahead.

src/xpk/core/kueue.py
sharabiani: they are combined but the name is still `prepare..` . I think `verify...` is more descriptive.

sharabiani (approved):

sharabiani (approved):

PR #201, opened 2024-10-09
Job info command implementation

BluValor:
## Fixes / Features
-
-

## Testing / Documentation
Testing details.

- [ y/n ] Tests pass


examples/batch.md
```diff
@@ -0,0 +1,29 @@
...
+
+### Preparation
+
+This section should describe how to prepare volumes and add external files executed in script.
+
+### Execution
+
+To run xpk batch script run:
+```bash
+python3 xpk.py batch slurm_sample.sh
```
sharabiani, line 11: are we allowed to use `slurm`?

pawloch00: I think kjob mode we use is named slurm, that is why i named script this way

pawloch00:
It's not building. Please switch to https://pypi.org/project/ruamel.yaml/ if possible, as it is more convenient to use with classes and will be added in future PRs

pawloch00:
Please add tests in .github files

.github/workflows/build_tests.yaml
```diff
@@ -88,6 +88,10 @@ jobs:
       run: python3 xpk.py workload delete --workload $WORKLOAD_NAME --cluster $TPU_CLUSTER_NAME --zone=us-central2-b
     - name: Delete the Pathways workload on the cluster
       run: python3 xpk.py workload delete --workload $PATHWAYS_WORKLOAD_NAME --cluster $TPU_CLUSTER_NAME --zone=us-central2-b
+    - name: Run a batch job on the cluster
+      run: python3 xpk.py batch --cluster $TPU_CLUSTER_NAME --zone=us-central2-b test.sh
+    - name: Get job info for the last job created on the cluster
+      run: python3 xpk.py job info $(kubectl get jobs --sort-by=.metadata.creationTimestamp -o jsonpath="{.items[-1:].metadata.name}") | grep "Exiting XPK cleanly"
```
pawloch00, line 94: Please check if command output format is correct

pawloch00: And add this tests to nightly

.github/workflows/nightly_tests.yaml
```diff
@@ -95,6 +95,12 @@ jobs:
       run : python3 xpk.py info --cluster $TPU_CLUSTER_NAME --zone=us-central2-b
     - name: Delete the workload on the cluster
       run: python3 xpk.py workload delete --workload $WORKLOAD_NAME --cluster $TPU_CLUSTER_NAME --zone=us-central2-b
+    - name: Run a batch job on the cluster
+      run: python3 xpk.py batch --cluster $TPU_CLUSTER_NAME --zone=us-central2-b test.sh
+    - name: Get job info for the last job created on the cluster
+      run: |
+        export MOST_RECENT_JOB_NAME=$(kubectl get jobs --sort-by=.metadata.creationTimestamp -o jsonpath="{.items[-1:].metadata.name}") 
+        python3 xpk.py job info $MOST_RECENT_JOB_NAME | grep "Exiting XPK cleanly"
```
pawloch00, line 103: grep EXiting xpk cleanly makes no sensie. If xpk does not exits cleany it will exit with code != 0 and pipeline will fail

src/xpk/commands/job.py
```diff
@@ -0,0 +1,97 @@
...
+  pods_command = f'kubectl get pods -l=job-name={job_name} --no-headers'
+  pods_code, pods_text = run_command_for_value(
+      pods_command, 'Getting pods list', args
+  )
+  if pods_code != 0:
+    xpk_print(f'Pods list request returned ERROR {pods_code}')
+    xpk_exit(pods_code)
+
+  yaml = YAML(typ='safe')
+  job_yaml = yaml.load(job_text)['items'][0]
```
pawloch00, line 63: should we check for presence of items and if it has item at 0 index

src/xpk/commands/job.py
```diff
@@ -0,0 +1,97 @@
...
+      ],
+      'Environment variables': get_ev_vars(desc_text),
+      'Pods': get_pods(pods_text),
+  }
+
+  yaml.default_flow_style = False
+  yaml.sort_keys = False
+  yaml.dump(output, sys.stdout)
+
+  xpk_exit(0)
```
pawloch00, line 80: I am not use if xpk_exit(0) is neccessary

.github/workflows/nightly_tests.yaml
BluValor: Right, forgot that the command itself exists with a number.

src/xpk/commands/job.py
BluValor: The question is more like - should we assume, that we trust kjobctls output? Do we trust the "0" output code they provide?

BluValor: We came to conclusion that relying on kjobctl output code should be enough, also because the alternative is very cumbersome.

src/xpk/commands/job.py
BluValor: Right, I don't think it is. The code exits properly without it. Although some old commands (eg. cluster) add this at the end. Possible cleanup target.

BluValor: The only difference is the text prompt we get at the end ("Exiting XPK cleanly", "XPK Done.").

src/xpk/commands/job.py
```diff
@@ -14,12 +14,87 @@
...
+
+
+def get_profile(job_yaml: dict) -> str:
+  env_vars = job_yaml['spec']['template']['spec']['containers'][0]['env']
+  profile = next((x['value'] for x in env_vars if x['name'] == 'PROFILE'), '')
+  return profile
+
+
+def get_ev_vars(job_desc_text: str) -> list[tuple[str, str]]:
+  regex = r'(SLURM_[A-Z_]*=.*)'
```
mbobrovskyi, line 90: It will not work well. We can't see env variables with describe, because it is generating on runtime.

src/xpk/commands/job.py
```diff
@@ -14,12 +14,87 @@
...
+  search_res = re.findall(regex, job_desc_text)
+  return search_res
```
mbobrovskyi, line 92: It will not work well. We can't see env variables with describe, because it is generating on runtime and it is different for each pod.

mbobrovskyi:
cc: @IrvingMg

.github/workflows/nightly_tests.yaml
```diff
@@ -89,6 +89,12 @@ jobs:
       run : python3 xpk.py info --cluster $TPU_CLUSTER_NAME --zone=us-central2-b
     - name: Delete the workload on the cluster
       run: python3 xpk.py workload delete --workload $WORKLOAD_NAME --cluster $TPU_CLUSTER_NAME --zone=us-central2-b
+    - name: Run a batch job on the cluster
+      run: python3 xpk.py batch --cluster $TPU_CLUSTER_NAME --zone=us-central2-b test.sh
+    - name: Get job info for the last job created on the cluster
+      run: |
+        export MOST_RECENT_JOB_NAME=$(kubectl get jobs --sort-by=.metadata.creationTimestamp -o jsonpath="{.items[-1:].metadata.name}")
```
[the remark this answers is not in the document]
pawloch00: can items be empty?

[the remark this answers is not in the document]
BluValor: It shouldn't be, since the previous test (get job list) would fail if it is. I will also add the "get job list" test to nightly tests for this reason.

src/xpk/commands/job.py
```diff
@@ -14,12 +14,87 @@
...
+
+  Args:
+    args: user provided arguments for running the command.
+
+  Returns:
+    0 if successful and 1 otherwise.
+  """
+  job_name = args.name
+
+  desc_command = f'kubectl-kjob describe slurm {job_name}'
```
IrvingMg, line 37: For readibility of `job_info`, I'd suggest splitting the command executions - describe, list, get - into functions like `run_describe_slurm_command(args)` that return an `int`.

src/xpk/commands/job.py
```diff
@@ -14,12 +14,87 @@
...
+
+
+def job_info(args):
+  """Run commands obtaining information about a job given by name.
+
+  Args:
+    args: user provided arguments for running the command.
+
+  Returns:
+    0 if successful and 1 otherwise.
```
IrvingMg, line 33:
```suggestion
    None
```

src/xpk/commands/job.py
```diff
@@ -14,12 +14,87 @@
...
+  yaml.dump(output, sys.stdout)
+
+
+def get_profile(job_yaml: dict) -> str:
+  env_vars = job_yaml['spec']['template']['spec']['containers'][0]['env']
+  profile = next((x['value'] for x in env_vars if x['name'] == 'PROFILE'), '')
+  return profile
+
+
+def get_ev_vars(job_desc_text: str) -> list[tuple[str, str]]:
```
IrvingMg, line 89:
```suggestion
def get_env_vars(job_desc_text: str) -> list[tuple[str, str]]:
```

src/xpk/commands/job.py
BluValor: Here is and example of the output of this command (only the relevant part):
```
- SLURM_ARRAY_JOB_ID=1
- SLURM_ARRAY_TASK_COUNT=3
- SLURM_ARRAY_TASK_MAX=3
- SLURM_ARRAY_TASK_MIN=1
- SLURM_TASKS_PER_NODE=1
- SLURM_CPUS_PER_TASK=
- SLURM_CPUS_ON_NODE=
- SLURM_JOB_CPUS_PER_NODE=
- SLURM_CPUS_PER_GPU=
- SLURM_MEM_PER_CPU=
- SLURM_MEM_PER_GPU=
- SLURM_MEM_PER_NODE=
- SLURM_GPUS=
- SLURM_NTASKS=1
- SLURM_NTASKS_PER_NODE=1
- SLURM_NPROCS=1
- SLURM_NNODES=2
- SLURM_SUBMIT_DIR=/slurm/scripts
- SLURM_SUBMIT_HOST=$HOSTNAME
- SLURM_JOB_NODELIST=xpk-def-app-profile-slurm-hvz49-0.xpk-def-app-profile-slurm-hvz49,xpk-def-app-profile-slurm-hvz49-1.xpk-def-app-profile-slurm-hvz49
- SLURM_JOB_FIRST_NODE=xpk-def-app-profile-slurm-hvz49-0.xpk-def-app-profile-slurm-hvz49
- SLURM_JOB_ID=$(expr $JOB_COMPLETION_INDEX \* 1 + $i + 1)
- SLURM_JOBID=$(expr $JOB_COMPLETION_INDEX \* 1 + $i + 1)
- SLURM_ARRAY_TASK_ID=$container_index
- SLURM_JOB_FIRST_NODE_IP=${SLURM_JOB_FIRST_NODE_IP:-""}
```

sure some of them are empty / unevaluated expressions, but some are set. There is value in printing it out. 

For running containers I could exec into the container and print out the values of the variables. It would be easy to do for ordinary env variables, but a hassle for this ones since they are written to some file in the filesystem of the pod, not exported into the env itself.

Should I just print out ordinary env variables and only for running pods?
Should I print out these statements, or maybe only the ones which values are known at runtime?
Are they really env variables if they are not being set? Are they even relevant for the user if they are not set in the pod?

Personally I would suggest printing env variables only for running pods, and only the real env variables, set and visible when using printenv. What do you think? Is there value in that?

src/xpk/commands/job.py
BluValor: Can I create a helper function run_command_for_value_or_exit? Because this is essentially what's happening, and it's supposed to be taken care of by refactor. If not I would leave it as is and wait for exit codes refactor, because then each of these blocks would transform into simple:
```
some_info = run_command_for_value_or_exit("command text for some info")
```
and they would be simple and readable one liners just with variable names.

IrvingMg: I was thinking of following a similar approach to what we have in `cluster_create`:

https://github.com/AI-Hypercomputer/xpk/blob/3c3b1b54806cdedffa96bbbfb2df00de1e93d2ca/src/xpk/commands/cluster.py#L63

where the command handler - in that case `cluster_create` is in charge of exits based on the returned code by the functions invoked within it - such as `get_gke_server_config`, `set_cluster_command`, etc.-.

However, it's just a suggestion and we may refactor later if needed, so I'm fine if we leave it as is.

BluValor: I will leave it as it is then, since invoking commands is a subject for refactor in the future, and it will simplify it even more.

src/xpk/commands/job.py
BluValor: They are not shown using printenv, they sit in files in slurm/env/i (where i in what I used is 0, but I'm not sure if  can count on it). All I'm getting using printenv are:
```
KUBERNETES_SERVICE_PORT_HTTPS, KUBERNETES_SERVICE_PORT, USER_ID, HOSTNAME, PROFILEPWD, JOB_COMPLETION_INDEX, HOME, KUBERNETES_PORT_443_TCP, LS_COLORS, TIMESTAMP, TERM, TASK_NAME, SHLVL, KUBERNETES_PORT_443_TCP_PROTO, JOB_CONTAINER_INDEX, KUBERNETES_PORT_443_TCP_ADDR, KUBERNETES_SERVICE_HOST, KUBERNETES_PORT, KUBERNETES_PORT_443_TCP_PORT, PATH, TASK_ID, OLDPWD, _
```

mbobrovskyi: I mean execute printenv on the script file.

BluValor: So I guess I will provide status of the pod, and if it's running I will also print environment variables that are set within the pod (obtainable with printenv)? It seems to me like this is what is usable by user. Or maybe the info from the slurm/env files would also be beneficial?

mbobrovskyi: Yes, but for each container it can be different.

mbobrovskyi: I'm not sure that we need to show this variables on the describe command. Is it requirementes?

mbobrovskyi: > I mean execute printenv on the script file and get it by logs.

Ah we can't log on script.

mbobrovskyi: > Or maybe the info from the slurm/env files would also be beneficial?

On that case I think this is only one way. But again, it can be different for each container.

.github/workflows/nightly_tests.yaml
```diff
@@ -89,6 +89,14 @@ jobs:
       run : python3 xpk.py info --cluster $TPU_CLUSTER_NAME --zone=us-central2-b
     - name: Delete the workload on the cluster
       run: python3 xpk.py workload delete --workload $WORKLOAD_NAME --cluster $TPU_CLUSTER_NAME --zone=us-central2-b
+    - name: Run a batch job on the cluster
+      run: python3 xpk.py batch --cluster $TPU_CLUSTER_NAME --zone=us-central2-b test.sh
+    - name: List out the jobs on the cluster
+      run: python3 xpk.py job ls | grep 'xpk-def-app-profile-slurm-'
```
pawloch00, line 95: are we checking for presence of job we ran?

pawloch00 (approved):

.github/workflows/nightly_tests.yaml
BluValor: In run 113 we are getting the last job started, which - if the task "Run a batch job on the cluster" completed successfully - should be the batch job.

pawloch00 (approved):

PR #202, opened 2024-10-09
Add create-ray-cluster subcommand

mbzomowski:
## Fixes / Features
-Adds the create-ray subcommand, which creates a RayCluster with a head node and a workergroup that utilizes all TPU resources

## Testing / Documentation
Manually tested `python3 xpk.py cluster create-ray ...` with several different TPU configurations. Also added a nightly test for RayCluster creation.

- [ y ] Tests pass

### DIFF
--- a/.github/workflows/nightly_tests.yaml
+++ b/.github/workflows/nightly_tests.yaml
@@ -185,6 +185,12 @@ jobs:
       run: |
         gcloud config set compute/zone us-east4-a
         gcloud config get compute/zone
+    - name: Install xpk dependencies
+      run: |
+        make install 
+        echo $PWD/bin >> "$GITHUB_PATH"
+    - name: Check xpk installation
+      run: xpk --help
     - name: Create a RayCluster-enabled XPK Cluster with 2 x v4-8 nodepools
       run: python xpk.py cluster create-ray --cluster $RAYCLUSTER_TPU_CLUSTER_NAME --tpu-type=v4-8  --num-slices=2 --zone=us-central2-b --default-pool-cpu-machine-type=n1-standard-16 --default-pool-cpu-num-nodes=16 --reservation='${{ secrets.GCP_TPU_V4_RESERVATION }}' --custom-cluster-arguments='${{ secrets.CLUSTER_ARGUMENTS}}'
     - name: Delete the RayCluster-enabled XPK cluster
--- a/.github/workflows/nightly_tests.yaml
+++ b/.github/workflows/nightly_tests.yaml
@@ -165,7 +165,7 @@ jobs:
       run: python xpk.py cluster delete --cluster $PATHWAYS_TPU_CLUSTER_NAME --zone=us-central2-b
 
   rc-cluster:
-    runs-on: [ubuntu-20.04]
+    runs-on: [ubuntu-22.04]
     concurrency: # We support one build test to run at a time currently.
       group: nightly-rc-test-cluster-group
       cancel-in-progress: false
--- a/.github/workflows/nightly_tests.yaml
+++ b/.github/workflows/nightly_tests.yaml
@@ -186,7 +186,7 @@ jobs:
         gcloud config set compute/zone us-east4-a
         gcloud config get compute/zone
     - name: Create a RayCluster-enabled XPK Cluster with 2 x v4-8 nodepools
-      run: python xpk.py cluster create-ray --cluster $RAYCLUSTER_TPU_CLUSTER_NAME --tpu-type=v4-8  --num-slices=2 --zone=us-central2-b --default-pool-cpu-machine-type=n1-standard-16 --default-pool-cpu-num-nodes=16 --reservation='${{ secrets.GCP_TPU_V4_RESERVATION }}'
+      run: python xpk.py cluster create-ray --cluster $RAYCLUSTER_TPU_CLUSTER_NAME --tpu-type=v4-8  --num-slices=2 --zone=us-central2-b --default-pool-cpu-machine-type=n1-standard-16 --default-pool-cpu-num-nodes=16 --reservation='${{ secrets.GCP_TPU_V4_RESERVATION }}' --custom-cluster-arguments="${CLUSTER_ARGUMENTS}"
     - name: Delete the RayCluster-enabled XPK cluster
       if: always()
       run: python xpk.py cluster delete --cluster $RAYCLUSTER_TPU_CLUSTER_NAME --zone=us-central2-b
--- a/.github/workflows/nightly_tests.yaml
+++ b/.github/workflows/nightly_tests.yaml
@@ -192,7 +192,7 @@ jobs:
     - name: Check xpk installation
       run: xpk --help
     - name: Create a RayCluster-enabled XPK Cluster with 2 x v4-8 nodepools
-      run: python xpk.py cluster create-ray --cluster $RAYCLUSTER_TPU_CLUSTER_NAME --tpu-type=v4-8  --num-slices=2 --zone=us-central2-b --default-pool-cpu-machine-type=n1-standard-16 --default-pool-cpu-num-nodes=16 --reservation='${{ secrets.GCP_TPU_V4_RESERVATION }}' --custom-cluster-arguments='${{ secrets.CLUSTER_ARGUMENTS}}'
+      run: python xpk.py cluster create-ray --cluster $RAYCLUSTER_TPU_CLUSTER_NAME --tpu-type=v4-8  --num-slices=2 --zone=us-central2-b --ray-version=2.39.0 --default-pool-cpu-machine-type=n1-standard-16 --default-pool-cpu-num-nodes=16 --reservation='${{ secrets.GCP_TPU_V4_RESERVATION }}' --custom-cluster-arguments='${{ secrets.CLUSTER_ARGUMENTS}}'
     - name: Delete the RayCluster-enabled XPK cluster
       if: always()
       run: python xpk.py cluster delete --cluster $RAYCLUSTER_TPU_CLUSTER_NAME --zone=us-central2-b
--- a/.github/workflows/nightly_tests.yaml
+++ b/.github/workflows/nightly_tests.yaml
@@ -186,7 +186,7 @@ jobs:
         gcloud config set compute/zone us-east4-a
         gcloud config get compute/zone
     - name: Create a RayCluster-enabled XPK Cluster with 2 x v4-8 nodepools
-      run: python xpk.py cluster create-ray --cluster $RAYCLUSTER_TPU_CLUSTER_NAME --tpu-type=v4-8  --num-slices=2 --zone=us-central2-b --default-pool-cpu-machine-type=n1-standard-16 --default-pool-cpu-num-nodes=16 --reservation='${{ secrets.GCP_TPU_V4_RESERVATION }}' --custom-cluster-arguments="${CLUSTER_ARGUMENTS}"
+      run: python xpk.py cluster create-ray --cluster $RAYCLUSTER_TPU_CLUSTER_NAME --tpu-type=v4-8  --num-slices=2 --zone=us-central2-b --default-pool-cpu-machine-type=n1-standard-16 --default-pool-cpu-num-nodes=16 --reservation='${{ secrets.GCP_TPU_V4_RESERVATION }}' --custom-cluster-arguments='${{ secrets.CLUSTER_ARGUMENTS}}'
     - name: Delete the RayCluster-enabled XPK cluster
       if: always()
       run: python xpk.py cluster delete --cluster $RAYCLUSTER_TPU_CLUSTER_NAME --zone=us-central2-b
--- a/.github/workflows/nightly_tests.yaml
+++ b/.github/workflows/nightly_tests.yaml
@@ -186,7 +186,7 @@ jobs:
         gcloud config set compute/zone us-east4-a
         gcloud config get compute/zone
     - name: Create a RayCluster-enabled XPK Cluster with 2 x v4-8 nodepools
-      run: python xpk.py cluster create-ray-cluster --cluster $RAYCLUSTER_TPU_CLUSTER_NAME --tpu-type=v4-8  --num-slices=2 --zone=us-central2-b --default-pool-cpu-machine-type=n1-standard-16 --default-pool-cpu-num-nodes=16 --reservation='${{ secrets.GCP_TPU_V4_RESERVATION }}'
+      run: python xpk.py cluster create-ray --cluster $RAYCLUSTER_TPU_CLUSTER_NAME --tpu-type=v4-8  --num-slices=2 --zone=us-central2-b --default-pool-cpu-machine-type=n1-standard-16 --default-pool-cpu-num-nodes=16 --reservation='${{ secrets.GCP_TPU_V4_RESERVATION }}'
     - name: Delete the RayCluster-enabled XPK cluster
       if: always()
       run: python xpk.py cluster delete --cluster $RAYCLUSTER_TPU_CLUSTER_NAME --zone=us-central2-b
--- a/.github/workflows/nightly_tests.yaml
+++ b/.github/workflows/nightly_tests.yaml
@@ -29,6 +29,8 @@ env:
   PATHWAYS_WORKLOAD_NAME: xpkpw-nightly-${{ github.run_attempt }}
   PW_CLUSTER_ARGUMENTS: "--network=${{secrets.NETWORK_NAME}}"
   CLUSTER_NETWORK_ARGUMENTS: "--network=${{secrets.NETWORK_NAME}} --subnetwork=${{secrets.SUBNETWORK_NAME}}"
+  RAYCLUSTER_TPU_CLUSTER_NAME: rc-nightly-test-2-v4-8-nodepools
+
 jobs:
   cluster-create-and-delete:
     runs-on: [ubuntu-22.04]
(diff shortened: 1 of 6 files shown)


src/xpk/core/ray.py
```diff
@@ -0,0 +1,112 @@
...
+from ..utils import write_tmp_file, xpk_print
+from .commands import run_command_with_updates_retry
+
+
+ray_cluster_crd_yaml = """apiVersion: ray.io/v1
+kind: RayCluster
+metadata:
+  name: raycluster
+spec:
+  rayVersion: '2.34.0'
```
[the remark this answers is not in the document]
mbzomowski: Yeah good idea

src/xpk/commands/cluster.py
```diff
@@ -337,6 +343,21 @@ def cluster_create_pathways(args) -> None:
...
+
+def cluster_create_ray_cluster(args) -> None:
+  """Function around cluster creation for RayCluster.
+
+  Args:
+    args: user provided arguments for running the command.
+
+  Returns:
+    None
+  """
```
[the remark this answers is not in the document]
mbzomowski: All the checks passed, which includes running `create-pathways` (https://github.com/AI-Hypercomputer/xpk/actions/runs/11264616690/workflow?pr=202#L57), so for some reason, that isn't needed.

.github/workflows/nightly_tests.yaml
```diff
@@ -115,6 +116,33 @@ jobs:
       if: always()
       run: python xpk.py cluster delete --cluster $PATHWAYS_TPU_CLUSTER_NAME --zone=us-central2-b
 
+  rc-cluster:
+    runs-on: [ubuntu-20.04]
```
pawloch00, line 120: Could you move to Ubuntu 22.04 please? Some packages eg. kueuectl that we are incorporating into this repo have some issues with Ubuntu 20.04

src/xpk/core/ray.py
```diff
@@ -0,0 +1,115 @@
...
+    system: system characteristics.
+
+  Returns:
+    0 if successful and 1 otherwise.
+  """
+
+  command = 'kubectl delete rayclusters --all'
+  task = 'Deleting old RayCluster'
+  retry_attempts = 1
+  return_code = run_command_with_updates_retry(
```
pawloch00, line 93: return_code value is never checked

src/xpk/core/ray.py
```diff
@@ -0,0 +1,115 @@
...
+                  cpu: 1
+                  google.com/tpu: {chips_per_vm}
+                  memory: 40G
+          nodeSelector:
+            cloud.google.com/gke-tpu-accelerator: {accelerator}
+            cloud.google.com/gke-tpu-topology: {topology}
+"""
+
+
+def install_ray_cluster(args, system) -> int:
```
pawloch00, line 79: This function does two things: it deletes old cluster and installs a new one. Can we split it into two methods?

mbzomowski: Good call

src/xpk/core/ray.py
```diff
@@ -0,0 +1,132 @@
...
+  task = 'Applying RayCluster'
+  retry_attempts = 1
+  return_code = run_command_with_updates_retry(
+      command, task, args, num_retry_attempts=retry_attempts
+  )
+  if return_code != 0:
+    xpk_print(f'{task} not successful.')
+  return return_code
+
+def delete_ray_cluster(args) -> None:
```
pawloch00, line 112: signature should be changed to return int

src/xpk/commands/cluster.py
```diff
@@ -177,6 +178,11 @@ def cluster_create(args) -> None:
   if create_cluster_configmaps_code != 0:
     xpk_exit(create_cluster_configmaps_code)
 
+  if args.enable_ray_cluster:
+    return_code = install_ray_cluster(args, system)
+  if return_code != 0:
```
pawloch00, line 183: what if args.enable_ray_cluster is not defined? Shall we check return code than?

src/xpk/core/ray.py
```diff
@@ -0,0 +1,133 @@
...
+  )
+
+  tmp = write_tmp_file(yml_string)
+  command = f'kubectl apply -f {str(tmp.file.name)}'
+  task = 'Applying RayCluster'
+  retry_attempts = 1
+  return_code = run_command_with_updates_retry(
+      command, task, args, num_retry_attempts=retry_attempts
+  )
+  if return_code != 0:
```
pawloch00, line 108: Please return immediately and don't propagate return code up

mbzomowski: Sorry, can you clarify what you mean here?

mbzomowski: do you mean xpk_exit(1)?

src/xpk/commands/cluster.py
mbzomowski: Oh, whoops. Meant to indent 183-184

src/xpk/core/ray.py
pawloch00: yes, xpk_exit(return_code)

src/xpk/parser/cluster.py
```diff
@@ -169,25 +170,87 @@ def set_cluster_parser(cluster_parser):
...
+  )
+  cluster_create_ray_cluster_tensorboard_arguments = (
+      cluster_create_ray_cluster_parser.add_argument_group(
+          'Optional Vertex AI Tensorboard Arguments',
+          'Arguments for creating Vertex AI Tensorboard in cluster create.',
+      )
+  )
+
+  ### RayCluster required arguments specific to "cluster create"
+  cluster_create_ray_cluster_required_arguments.add_argument(
```
pawloch00, line 205: If this argument is required it should have required=True I think

src/xpk/core/ray.py
```diff
@@ -0,0 +1,134 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+
+from ..utils import write_tmp_file, xpk_print, xpk_exit
+from .commands import run_command_with_updates_retry
+
+
+RAY_VERSION = '2.34.0'
```
allenwang28, line 21: This should be configurable from an argument OR, better yet, auto detected from the Ray version the user has specified. Ray versions update too frequently for this to be fixed - as of today we're already at 2.38.0

I figure that the latter will be harder to implement, so I would suggest starting with an argument today

src/xpk/core/ray.py
```diff
@@ -0,0 +1,134 @@
...
+      rayStartParams:
+        block: 'true'
+      template:
+        spec:
+          containers:
+            - name: ray-worker
+              image: rayproject/ray:{version}
+              resources:
+                limits:
+                  cpu: 1
```
allenwang28, line 66: 1 cpu and 40 GB memory limits are not useful.. Can we autodetect what the correct CPU and memory are for the TPU generation?

src/xpk/core/ray.py
```diff
@@ -0,0 +1,134 @@
...
+    rayStartParams: {{}}
+    #pod template
+    template:
+      spec:
+        containers:
+        - name: ray-head
+          image: rayproject/ray:{version}
+          resources:
+            limits:
+              cpu: 1
```
allenwang28, line 39: Similar comment about cpu and memory limits

src/xpk/core/ray.py
```diff
@@ -0,0 +1,134 @@
...
+    - replicas: {replicas} # TODO: Set min and max replicas
+      numOfHosts: {num_hosts}
+      minReplicas: {replicas}
+      maxReplicas: {replicas}
+      groupName: workergroup0
+      rayStartParams:
+        block: 'true'
+      template:
+        spec:
+          containers:
```
allenwang28, line 61: We will need to port forward the following:
- 8081 for multislice
- 8471 for TPU communication

src/xpk/commands/cluster.py
```diff
@@ -177,6 +178,11 @@ def cluster_create(args) -> None:
   if create_cluster_configmaps_code != 0:
     xpk_exit(create_cluster_configmaps_code)
 
+  if args.enable_ray_cluster:
+    return_code = install_ray_cluster(args, system)
+    if return_code != 0:
+      xpk_exit(return_code)
```
richardsliu, line 184: Print an error here?

src/xpk/core/ray.py
```diff
@@ -0,0 +1,134 @@
...
+  """Delete all RayClusters on the cluster
+
+  Args:
+    args: user provided arguments for running the command.
+
+  Returns:
+    0 if successful and 1 otherwise.
+  """
+
+  command = 'kubectl delete rayclusters --all'
```
richardsliu, line 124: Should we have an option for deleting specific ray clusters instead of all?

src/xpk/core/ray.py
```diff
@@ -0,0 +1,134 @@
...
+              cpu: 500m
+              memory: 2Gi
+          ports:
+          - containerPort: 6379
+            name: gcs-server
+          - containerPort: 8265 # Ray dashboard
+            name: dashboard
+          - containerPort: 10001
+            name: client
+  workerGroupSpecs:
```
richardsliu, line 51: What if we need to create multiple worker groups (e.g. for heterogenous clusters)?

src/xpk/core/ray.py
```diff
@@ -0,0 +1,134 @@
...
+from ..utils import write_tmp_file, xpk_print, xpk_exit
+from .commands import run_command_with_updates_retry
+
+
+RAY_VERSION = '2.34.0'
+
+ray_cluster_crd_yaml = """apiVersion: ray.io/v1
+kind: RayCluster
+metadata:
+  name: raycluster
```
richardsliu, line 26: Is namespace needed here?

src/xpk/core/ray.py
richardsliu, line 39: These resources are probably insufficient for the ray head

mbzomowski: Added a mechanism to requisition memory and cpu as a percent of the available resources in the node. I set it to 50% for the head node, let me know what you think.

src/xpk/core/ray.py
mbzomowski: Added a ray namespace

src/xpk/core/ray.py
mbzomowski: That'll be handled in a follow-up PR

src/xpk/core/ray.py
mbzomowski: Is that something you think we need? I don't see us ever having more than one RayCluster in an XPK cluster at a time.

src/xpk/core/ray.py
mbzomowski: Set the cpu and memory to be 90% of the available resources in the TPU.

src/xpk/core/ray.py
```diff
@@ -0,0 +1,134 @@
...
+
+  Args:
+    args: user provided arguments for running the command.
+    system: system characteristics.
+
+  Returns:
+    0 if successful and 1 otherwise.
+  """
+
+  delete_ray_cluster(args)
```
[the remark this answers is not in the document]
mbzomowski: Added check inside delete_ray_cluster which exits if it's not successful. Now this function returns None.

src/xpk/core/ray.py
mbzomowski: Resolved offline. Added in 8081 for multislice

.github/workflows/nightly_tests.yaml
```diff
@@ -138,6 +139,33 @@ jobs:
...
+        credentials_json: '${{ secrets.GCP_SA_KEY }}'
+    - uses: google-github-actions/setup-gcloud@v2
+      with:
+        version: '>= 363.0.0'
+        install_components: 'beta,gke-gcloud-auth-plugin'
+    - name: Set Google Cloud CLI properties to a unused zone to verify --zone arg is passed properly in commands.
+      run: |
+        gcloud config set compute/zone us-east4-a
+        gcloud config get compute/zone
+    - name: Create a RayCluster-enabled XPK Cluster with 2 x v4-8 nodepools
```
pawloch00, line 163: Is there any check that can verify if ray cluster is indeed enabled?

pawloch00 (approved):

.github/workflows/nightly_tests.yaml
mbzomowski: Not afaik, but it will print out an error if RayCluster installation fails (which will happen if the cluster isn't RayCluster-enabled): https://github.com/AI-Hypercomputer/xpk/pull/202/files/64a513324e92667626398145345694d92bd7768b#diff-389a6611cb89e528a2bff81de17ce1457ce4927af04361f1ff8d19b378c33455R203

richardsliu (approved):

.github/workflows/nightly_tests.yaml
Obliviour, line 163: I saw a fix to the nightly tests just submitted: https://github.com/AI-Hypercomputer/xpk/pull/225/files, can you also add the "--network=${{secrets.NETWORK_NAME}}" flag here to contain that fix in your tests as well

src/xpk/parser/cluster.py
```diff
@@ -169,25 +170,95 @@ def set_cluster_parser(cluster_parser):
...
+      '--tpu-type',
```
Obliviour, line 206: might want to add --device-type in the future to support GPUs/CPUs. def not needed now

src/xpk/core/ray.py
```diff
@@ -0,0 +1,198 @@
...
+  available_worker_cpu, available_worker_mem = generate_available_resources(
+      label, args, 0.9
+  )
+
+  yml_string = ray_cluster_crd_yaml.format(
+      accelerator=system.gke_accelerator,
+      topology=system.topology,
+      chips_per_vm=system.chips_per_vm,
+      num_hosts=system.vms_per_slice,
+      replicas=args.num_slices,
```
Obliviour, line 114: maybe this is suppose to the be arg instead of num_nodes?

src/xpk/parser/cluster.py
```diff
@@ -169,25 +170,95 @@ def set_cluster_parser(cluster_parser):
...
+  cluster_create_ray_cluster_required_arguments.add_argument(
+      '--ray-version',
+      type=str,
+      default=None,
+      help="The Ray version to use, e.g. '2.38.0'",
+      required=True,
+  )
+  ### Optional arguments specific to "cluster create"
+  cluster_create_ray_cluster_optional_arguments.add_argument(
+      '--num-nodes',
```
Obliviour, line 221: --num-nodes is the number of TPU slices? I couldn't find it used in the code. maybe it --num-slices

mbzomowski: Yeah I think this was just leftover from when I copied from the cluster create args. I removed it and everything still works fine.

src/xpk/core/ray.py
mbzomowski: Still works without explicitly adding num_slices as an arg.

src/xpk/parser/cluster.py
mbzomowski: Good call. Added a TODO

.github/workflows/nightly_tests.yaml
mbzomowski: Done

Obliviour (dismissed):

.github/workflows/nightly_tests.yaml
```diff
@@ -162,6 +164,33 @@ jobs:
       if: always()
       run: python xpk.py cluster delete --cluster $PATHWAYS_TPU_CLUSTER_NAME --zone=us-central2-b
 
+  rc-cluster:
```
pawloch00, line 167: Can we have at least workflow run and checked for execution status to ensure this is usable? Also, if you modify niggtly tests, please run them before merging and attach results to task description. If you are not able to run them please let me know.

src/xpk/core/ray.py
```diff
@@ -0,0 +1,203 @@
...
+          image: rayproject/ray:{version}
+          resources:
+            limits:
+              cpu: {head_cpu}
+              memory: {head_mem}
+            requests:
+              cpu: {head_cpu}
+              memory: {head_mem}
+          ports:
+          - containerPort: 6379
```
pawloch00, line 51: I don't like this hardcoded values, can you change them please?

src/xpk/core/ray.py
```diff
@@ -0,0 +1,203 @@
...
+
+  Returns:
+    0 if successful and 1 otherwise.
+  """
+
+  delete_ray_cluster(args)
+
+  label = 'cloud.google.com/gke-nodepool=default-pool'
+  available_head_cpu, available_head_mem = generate_available_resources(
+      label, args, 0.5
```
pawloch00, line 102: 0.5 seems a magic number a bit

src/xpk/core/ray.py
```diff
@@ -0,0 +1,203 @@
...
+  )
+
+  label = f'cloud.google.com/gke-tpu-accelerator={system.gke_accelerator}'
+  available_worker_cpu, available_worker_mem = generate_available_resources(
+      label, args, 0.9
```
pawloch00, line 107: 0.9 is hardcoded

src/xpk/core/ray.py
```diff
@@ -0,0 +1,203 @@
...
+  _, node_name = run_command_for_value(command, task, args)
+
+  command = (
+      f"kubectl get node {node_name} -o jsonpath='{{.status.allocatable.cpu}}'"
+  )
+  task = 'Fetching available CPU on node'
+  _, available_cpu = run_command_for_value(command, task, args)
+  match = re.match(r'(\d+)([a-zA-Z]+)', available_cpu)
+  if not match:
+    xpk_exit(1)
```
pawloch00, line 185: please add xpk_print with explanation why exiting (generating available resources failed due to ...)

src/xpk/main.py
```diff
@@ -61,6 +61,7 @@
 
 xpk_print('Starting xpk', flush=True)
 main_args = parser.parse_args()
+main_args.enable_ray_cluster = False
```
pawloch00, line 64: we shouldn't have any args being set here

.github/workflows/nightly_tests.yaml
mbzomowski: how do I run the nightly tests? Also run which workflow?

src/xpk/core/ray.py
mbzomowski: Done.

src/xpk/core/ray.py
mbzomowski: Done.

src/xpk/core/ray.py
mbzomowski: Done.

src/xpk/core/ray.py
mbzomowski: Done.

.github/workflows/nightly_tests.yaml
pawloch00: In actions select nightly tests and run from branch

src/xpk/main.py
mbzomowski: Already discussed this w/ @Obliviour

pawloch00: Maybe try: https://github.com/AI-Hypercomputer/xpk/blob/e860597c408939ff7e5670f931fc35d4eb82f125/src/xpk/parser/storage.py#L81 with default=False ? But not sure, you will have to verify it.

Maybe the good solution would be to add create-ray, the same as we have create-pathways now? It would keep commands consistent. @Obliviour wdyt? . I am opting for this solution: create create-ray command, so we have similar patterns for different types of clusters. Sorry for review taking so long, but I think we are on good path

pawloch00:
Could you please add some exmaple of new command to Readme?

src/xpk/main.py
mbzomowski: crate-pathways still has a deprecated flag that is exposed to the user, in addition to the subcommand, which is how the subcommand works afaict: https://github.com/AI-Hypercomputer/xpk/blob/main/src/xpk/parser/cluster.py#L91

But I'll test your suggestion out

Obliviour (dismissed):

src/xpk/main.py
mbzomowski: Oh I misunderstood - your suggestion is still adding a flag, which we don't want

mbzomowski:
nightly tests pass on most recent commit: https://github.com/AI-Hypercomputer/xpk/actions/runs/12145314031

pawloch00 (approved):

PR #205, opened 2024-10-14
Use host networking for PW pods communication and remove cpu and mem limits from them

shauryagup:
These changes result in better perf at higher scale. More context in b/372327141

## Fixes / Features
- better perf at higher scale by upto 40%

## Testing / Documentation
Verified changes on large GKE cluster

- [ Y ] Tests pass
- [ NA] Appropriate changes to documentation are included in the PR

### DIFF
--- a/src/xpk/commands/workload.py
+++ b/src/xpk/commands/workload.py
@@ -218,6 +218,8 @@ spec:
               {machine_label}
               {autoprovisioning_args}
             priorityClassName: {args.priority}
+            hostNetwork: true
+            dnsPolicy: ClusterFirstWithHostNet
             volumes:
             - hostPath:
                 path: /tmp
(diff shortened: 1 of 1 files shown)


src/xpk/commands/workload.py
```diff
@@ -218,6 +218,8 @@
               {machine_label}
               {autoprovisioning_args}
             priorityClassName: {args.priority}
+            hostNetwork: true
```
Obliviour, line 221: for my background, did you find any benefit to hostNetwork: false. McJax also using hostNetwork: true and sees good performance.

Obliviour (approved):

src/xpk/commands/workload.py
shauryagup: Yes, we saw a significant improvement in step time after this change. Before this, PW was consistently 30-40% worse on large slices where McJAX step time was ~16 sec and PW was ~26 sec.

sadikneipp (approved):

Obliviour (approved):

PR #208, opened 2024-10-16
Implement batch command - basic version

pawloch00:
## Fixes / Features
- This command allows for execution of batch script in basic version, it does not mounts any volumes. Executed script should not rely on other files.
-

## Testing / Documentation
Testing details.

- [ y/n ] Tests pass


src/xpk/core/job_template.py
```diff
@@ -0,0 +1,67 @@
...
+      completions=JOB_TEMPLATE_DEFAULT_COMPLETIONS,
+      container=JOB_TEMPLATE_DEFAULT_CONT_NAME,
+      image=JOB_TEMPLATE_DEFAULT_IMG,
+  )
+
+  tmp = write_tmp_file(yml_string)
+  command = f"kubectl apply -f {str(tmp.file.name)}"
+  return_code = run_command_with_updates(command, "Creating JobTemplate", args)
+  if return_code != 0:
+    xpk_exit(return_code)
```
BluValor, line 67: No need for "if", just exit with the exit code.

src/xpk/commands/batch.py
```diff
@@ -0,0 +1,59 @@
...
+      f' {APP_PROFILE_TEMPLATE_DEFAULT_NAME} --'
+      f' {args.script}'
+  )
+  if len(args.script_args) != 0:
+    cmd += f'{args.script_args}'
+  return_code, _ = run_command_for_value(cmd, 'submit job', args)
+
+  if return_code != 0:
+    xpk_print(f'Running batch job returned ERROR {return_code}')
+    xpk_exit(return_code)
```
BluValor, line 59: I'm just wondering if we should use xpk_exit() in the indirect functions (functions that are not explicitly called by parser). It doesn't seem to be the case in the rest of the code - such functions usually print out some info and return an error code which is then processed by the upper most function.

I don't necessarily agree with the way it is done currently. The way error codes are processed is usually just a single "if" followed by exit function. Still it gives us a guarantee that the code won't randomly exit when calling some function. I would like to keep this guarantee for now, so please return the error code or 0 and check for it in the batch() function.

src/xpk/core/app_profile.py
```diff
@@ -0,0 +1,50 @@
...
+  yml_string = app_profile_yaml.format(
+      name=APP_PROFILE_TEMPLATE_DEFAULT_NAME,
+      template=JOB_TEMPLATE_DEFAULT_NAME,
+  )
+
+  tmp = write_tmp_file(yml_string)
+  command = f"kubectl apply -f {str(tmp.file.name)}"
+  return_code = run_command_with_updates(command, "Creating AppProfile", args)
+  if return_code != 0:
+    xpk_exit(return_code)
```
BluValor, line 50: Same as: https://github.com/AI-Hypercomputer/xpk/pull/208/files#r1804347717

src/xpk/core/job_template.py
BluValor, line 67: Same as: https://github.com/AI-Hypercomputer/xpk/pull/208/files#r1804347717

src/xpk/commands/batch.py
pawloch00: I think it makes neccessary flow of error codes. If cmd execution exists with error we should exit immediately.

BluValor: Alright, but let's keep in mind that we should also in the future change the existing code to this convention.

BluValor (dismissed):

BluValor (dismissed):

BluValor (approved):

mbobrovskyi:
Can we merge this PR?

mbobrovskyi:
Thanks!

PR #209, opened 2024-10-16
Add example of running batch job

pawloch00:
## Fixes / Features
Add readme example of running batch jobs
## Testing / Documentation
Testing details.


examples/batch.md
```diff
@@ -0,0 +1,29 @@
...
+### Execution
+
+To run xpk batch script run:
+```bash
+python3 xpk.py batch slurm_sample.sh
+```
+
+You may pass script arguments to executed script by running:
+```bash
+python3 xpk.py batch slurm_sample.sh arg1 arg2
```
BluValor, line 16: It might be helpful for beginner users to add here example --cluster, --zone and --project (I think these 3 are required, correct me if wrong). I had to think about it for a minute, so this info will be beneficial when trying the command first time.

pawloch00: only --cluster is required, but will add it

BluValor, line 16: Seems it is not possible, remove.

BluValor (approved):

sharabiani (approved):

PR #212, opened 2024-10-18
Create kjob profile in cluster create command

pawloch00:
## Fixes / Features
- Create kjob CRDs during cluster creatiion. Create AppProfile and JobTemplate during cluster creation

## Testing / Documentation
Testing details.

- [ y ] Tests pass


.github/workflows/build_tests.yaml
```diff
@@ -49,6 +49,12 @@ jobs:
       run: |
         gcloud config set compute/zone us-east4-a
         gcloud config get compute/zone
+    - name: Install kjob
+      run: |
+        git clone https://github.com/kubernetes-sigs/kueue.git
+        cd kueue/cmd/experimental/kjobctl
+        make kubectl-kjob
+        mv ./bin/kubectl-kjob /usr/local/bin/kubectl-kjob
```
BluValor, line 57: Does it work without calling `make install`?

.github/workflows/nightly_tests.yaml
```diff
@@ -51,6 +51,12 @@ jobs:
       run: |
         gcloud config set compute/zone us-east4-a
         gcloud config get compute/zone
+    - name: Install kjob
+      run: |
+        git clone https://github.com/kubernetes-sigs/kueue.git
+        cd kueue/cmd/experimental/kjobctl
+        make kubectl-kjob
+        mv ./bin/kubectl-kjob /usr/local/bin/kubectl-kjob
```
BluValor, line 59: Same as https://github.com/AI-Hypercomputer/xpk/pull/212/files#r1818795019.

.github/workflows/nightly_tests.yaml
```diff
@@ -105,6 +111,12 @@ jobs:
...
```
BluValor, line 119: Same as https://github.com/AI-Hypercomputer/xpk/pull/212#discussion_r1818795019.

.github/workflows/build_tests.yaml
pawloch00: [make install](https://github.com/kubernetes-sigs/kueue/blob/615a81ede077078a86d1f4881a2029b212a669a5/cmd/experimental/kjobctl/Makefile#L160)

consists of: ```(KUSTOMIZE) build config/crd | $(KUBECTL) apply --server-side -f -``` commands
which i copied and executed in L160 of kjob.py during cluster creation. We only apply make kubectl-kjob as specified in [instruction](https://github.com/kubernetes-sigs/kueue/blob/main/cmd/experimental/kjobctl/docs/installation.md), make install is applied on cluster during creation.

src/xpk/core/kjob.py
```diff
@@ -0,0 +1,196 @@
...
+  builds them with kustomize and then applies result on cluster.
+  It creates all neccessary kjob CRDs.
+
+  Args:
+    args - user provided arguments
+  Returns:
+    None
+  """
+  temp_dir = tempfile.mkdtemp()
+  mkdir(join(temp_dir, "bases"))
```
BluValor, line 173: Can we extract the bases path to a new variable and call these variables something like "kjob_kustomize_path" and "kjob_kustomize_bases_path"? 

It would be then obvious that we are creating a directory within a directory.
Using tmp in name obsures what it really is hurting readability.

src/xpk/core/kjob.py
```diff
@@ -0,0 +1,196 @@
...
+  This function downloads kjob CRDs files from kjob repo,
+  builds them with kustomize and then applies result on cluster.
+  It creates all neccessary kjob CRDs.
+
+  Args:
+    args - user provided arguments
+  Returns:
+    None
+  """
+  temp_dir = tempfile.mkdtemp()
```
BluValor, line 172: Is it really a temporary directory? From the documentation of tempfile (https://docs.python.org/3/library/tempfile.html#tempfile.mkdtemp):

_The user of [mkdtemp()](https://docs.python.org/3/library/tempfile.html#tempfile.mkdtemp) is responsible for deleting the temporary directory and its contents when done with it._

src/xpk/core/kjob.py
```diff
@@ -0,0 +1,196 @@
...
+
+# AppProfile defaults
+APP_PROFILE_TEMPLATE_DEFAULT_NAME = "xpk-def-app-profile"
+APP_PROFILE_TEMPLATE_MODE_NAME = "Slurm"
+
+# JobTemplate defaults
+JOB_TEMPLATE_DEFAULT_NAME = "xpk-def-batch"
+JOB_TEMPLATE_DEFAULT_PARALLELISM = 1
+JOB_TEMPLATE_DEFAULT_COMPLETIONS = 1
+JOB_TEMPLATE_DEFAULT_COMPLETION_MODE = "Indexed"
```
BluValor, line 38: Unused.

src/xpk/core/kjob.py
```diff
@@ -0,0 +1,196 @@
...
+def download_files_from_github_into_dir(
+    path: str, urls: list[(str, str)]
+) -> None:
+  for url, fn in urls:
+    target = join(path, fn)
+    try:
+      urllib.request.urlretrieve(url, target)
+    except ContentTooShortError as e:
+      xpk_print(f"downloading kjob CRD {fn} failed due to {e.content}")
+      xpk_exit(1)
```
BluValor: This file does not seem like a proper place for this function. This is the type of function that has it's place in a general "utils" file (create core/utils?).

src/xpk/core/kjob.py
BluValor: Resolving as I see it's done.

BluValor (approved):

src/xpk/core/kjob.py
```diff
@@ -0,0 +1,186 @@
...
+  This function downloads kjob CRDs files from kjob repo,
+  builds them with kustomize and then applies result on cluster.
+  It creates all neccessary kjob CRDs.
+
+  Args:
+    args - user provided arguments
+  Returns:
+    None
+  """
+  kjob_kustomize_path = tempfile.mkdtemp()
```
sharabiani, line 157: Isn't it better to download files one by one and apply. them one by one?
If there is any issue with any files we can report it better?

Anyways,I beleive this method can be splitted in smaller pieces with a descriptive names. Example:
"download_crds"
"download_kustomization"// or something better
"apply_crds"

src/xpk/utils.py
```diff
@@ -165,3 +167,15 @@ def directory_path_type(value):
         f'Directory path is invalid. User provided path was {value}'
     )
   return value
+
+
+def download_files_from_github_into_dir(
+    path: str, urls: list[tuple[str, str]]
+) -> None:
+  for url, fn in urls:
```
sharabiani, line 175: I would make a function to download one file and would use it here.

src/xpk/commands/cluster.py
```diff
@@ -146,6 +152,14 @@ def cluster_create(args) -> None:
   if install_kueue_on_cluster_code != 0:
     xpk_exit(install_kueue_on_cluster_code)
 
+  xpk_print('Verifying kjob installation')
+  verify_kjob_installed(args)
```
sharabiani, line 156: Would it be better if these methods returned some results and you give the control of decision to exit app here?

src/xpk/commands/cluster.py
```diff
@@ -186,6 +200,11 @@ def cluster_create(args) -> None:
   xpk_exit(0)
 
 
+def prepare_kjob(args) -> None:
```
sharabiani, line 203: this can be in kjob.py?

README.md
```diff
@@ -71,7 +71,7 @@ xpk uses many tool to provide all neccessary functionalities. User must install
 - gcloud (install from [here](https://cloud.google.com/sdk/gcloud#download_and_install_the))
 - kubectl (install from [here](https://kubernetes.io/docs/tasks/tools/))
 - kueuectl (install from [here](https://kueue.sigs.k8s.io/docs/reference/kubectl-kueue/installation/))
-
```
sharabiani, line 74: kjob installation?

src/xpk/core/utils.py
```diff
@@ -0,0 +1,32 @@
...
+limitations under the License.
+"""
+
+import urllib.request
+from urllib.error import ContentTooShortError
+import os
+from ..utils import xpk_print, xpk_exit
+
+
+def download_files_from_github_into_dir(
```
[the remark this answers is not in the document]
pawloch00: sory, forgot to rm file from xpk/utils

src/xpk/core/kjob.py
pawloch00: I think if downloading any of the files fail, we should faill whole installation.
Same with applying. What I did here, is I repeated make install commend from kjob: https://github.com/kubernetes-sigs/kueue/blob/5203ac9eb69226a1ae3b6ba6d361eaed7ef3dc8c/cmd/experimental/kjobctl/Makefile#L160C1-L162C69 I did not want  to copy whole repo here or Makefile and invoke make from our code, that is why kustomization is used.

src/xpk/commands/cluster.py
BluValor: There is a similar discussion here: https://github.com/AI-Hypercomputer/xpk/pull/208#discussion_r1804347717

I would like to add here that if the method is named verify I would expect it to exit the app upon failing, if it would be something like check then I would expect it to return the result.

In this case I believe @pawloch00 should leave the code as is.

sharabiani (dismissed):

BluValor (dismissed):

mbobrovskyi:
Can we merge this PR?

mbobrovskyi:
Thanks!

PR #218, opened 2024-10-24
Use filestore in workload

pawloch00:
## Fixes / Features
-
-

## Testing / Documentation
Testing details.

- [ y/n ] Tests pass


src/xpk/commands/workload.py
```diff
@@ -447,6 +451,13 @@ def workload_create(args) -> None:
   else:
     xpk_print('No gcsfuse Storages to add detected')
 
+  if len(gcpfilestore_storages) > 0:
+    xpk_print(
+        f'Detected gcp filestores instances to add: {gcpfilestore_storages}'
+    )
+  else:
+    xpk_print('No detected gcp filestore instances to add detected.')
```
BluValor, line 459: 'No detected gcp filestore instances to add detected.'
 ->
 'No gcp filestore instances to add detected.'

src/xpk/commands/cluster.py
```diff
@@ -506,6 +506,9 @@ def run_gke_cluster_create_command(
     addons_str = ','.join(addons)
     command += f' --addons={addons_str}'
 
+  if len(addons) > 0:
```
44past4, line 509: Why is this repeated?

src/xpk/commands/workload.py
```diff
@@ -521,6 +538,7 @@ def workload_create(args) -> None:
         storage_annotations=storage_annotations,
         service_account=service_account,
     )
+  print(yml_string)
```
44past4, line 541: Is this a leftover from some debugging? Should we use xpk_print here? The yaml can be quite big do we really want to write with each execution?

src/xpk/core/core.py
```diff
@@ -2623,6 +2624,11 @@ def get_volumes(args, system: SystemCharacteristics) -> str:
                   claimName: {storage.pvc}
                   readOnly: {storage.readonly}
               """
+    if storage.type == GCP_FILESTORE_TYPE:
+      volumes += f"""- name: {storage.pv}
+                persistentVolumeClaim:
+                  claimName: {storage.pvc}
```
44past4, line 2630: What about a readOnly flag? Does filestore supports it? If not we should probably validate it when adding storage.

pawloch00: It is supported by filestore. Current behavior is that readonly can be defined in Persistent Volume yaml file and it will be applied in volume mount of this storage. If we use --readonly=true/false it will overwrite readonly value from PersistentVolume definiton.

44past4 (approved):

PR #220, opened 2024-10-25
Allow set XLA_FLAGS with --debug_dump_gcs

PatriosTheGreat:
## Testing / Documentation
Testing details.

- [ y/n ] Tests pass

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,6 +1,1 @@
-* @rwitten
-* @Obliviour
-* @44past4
-* @sharabiani
-* @pawloch00
-* @BluValor
+* @Obliviour @44past4 @sharabiani @pawloch00 @BluValor
--- a/src/xpk/commands/workload.py
+++ b/src/xpk/commands/workload.py
@@ -218,6 +218,8 @@ spec:
               {machine_label}
               {autoprovisioning_args}
             priorityClassName: {args.priority}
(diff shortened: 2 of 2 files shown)


src/xpk/core/core.py
```diff
@@ -233,13 +233,16 @@ def parse_env_config(args, tensorboard_config, system: SystemCharacteristics):
...
-            'XLA_FLAGS.'
+            'Conflict: --xla_dump_to defined in both --debug_dump_gcs '
+            'and environment file  via XLA_FLAGS. Please choose one way to define it.'
         )
-      env['XLA_FLAGS'] = '--xla_dump_to=/tmp/xla_dump/'
+      other_flags = ''
+      if 'XLA_FLAGS' in env:
+        other_flags = env['XLA_FLAGS']
+
+      env['XLA_FLAGS'] = '--xla_dump_to=/tmp/xla_dump/ ' + other_flags
```
pawloch00, line 245: Not sure if /tmp should be hardcoded? Maybe os os.temp() or some function to get system tmp path?

PatriosTheGreat: This path should be the same as in line 2122. I'm not sure if we can  pass around os.temp directory path between those two invocations.

pawloch00:
could you please provide any description of this change? Why it is introduced?

PR #222, opened 2024-10-25
Private cluster feature implemented

sharabiani:
## Fixes / Features
- `--private` and `--authorized-networks` added to cluster creation. They create a private cluster and open access to Control Plane for _**user's current machine**_ and the authorized-networks if provided.
- Some refactoring was done to clean the utils module. 

#### Manual test:
Note: To perform the tests successfully, Could NATs were created for the VPC in the required regions.
    - A private cluster created
    - On Cloud Console verified the private_nodes, ip_alias, and master_authorized_networks are enabled.
    - Using kubectl verified all required components such as kueue controller manager are up and running.


pawloch00:
Could you please extract utils refactor to another PR? It would make this PR smaller

src/xpk/commands/cluster.py
```diff
@@ -456,23 +464,33 @@ def run_gke_cluster_create_command(
...
   )
 
+  enable_ip_alias = False
+
+  if args.private or args.authorized_networks is not None:
+    enable_ip_alias = True
+    command += ' --enable-master-authorized-networks --enable-private-nodes'
+
   if system.accelerator_type == AcceleratorType['GPU']:
+    enable_ip_alias = True
```
pawloch00, line 474: why are we setting this var to True here?

sharabiani:
> Could you please extract utils refactor to another PR? It would make this PR smaller

We still have a lot to refactor in this codebase. To do it gradually I believe it's a good practice to refactor the parts we touch in our changes. WDYT?
Anyways, I can split it in a different PR if makes confusion here.

.github/workflows/build_tests.yaml
```diff
@@ -55,8 +55,8 @@ jobs:
       run: |
         pip install .
         xpk --help
-    - name: Create a Pathways-enabled XPK Cluster with 2x v4-8 nodepools. Larger num-nodes to avoid master resizing.
-      run: python xpk.py cluster create-pathways --cluster $TPU_CLUSTER_NAME --tpu-type=v4-8  --num-slices=2 --zone=us-central2-b --default-pool-cpu-machine-type=n1-standard-16 --default-pool-cpu-num-nodes=16 --reservation='${{ secrets.GCP_TPU_V4_RESERVATION }}'
+    - name: Create a private Pathways-enabled XPK Cluster with 2x v4-8 nodepools. Larger num-nodes to avoid master resizing.
```
pawloch00, line 58: Can we check if private flag is applied ?

sharabiani: done

pawloch00 (dismissed):

src/xpk/core/cluster_private.py
```diff
@@ -0,0 +1,170 @@
...
+      xpk_print('Updating cluster authorized networks failed!')
+      return cluster_authorized_networks_update_code
+    xpk_print("Cluster's master authorized networks updated successfully.")
+  else:
+    xpk_print('Current machine is already in authorized networks.')
+
+  return 0
+
+
+def is_cluster_private(args) -> bool:
```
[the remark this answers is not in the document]
BluValor: I second this, and not only in this place but in other functions as well.

src/xpk/core/cluster_private.py
```diff
@@ -0,0 +1,170 @@
...
+      command,
+      'Check if Private Nodes is enabled in cluster.',
+      args,
+  )
+
+  if return_code != 0:
+    xpk_print('Checking if Private Nodes is enabled failed!')
+    xpk_exit(return_code)
+
+  if private_nodes_enabled.strip().lower() == 'true':
```
BluValor, line 108: This function would make a good util, something like: is_text_true

src/xpk/core/cluster_private.py
```diff
@@ -0,0 +1,170 @@
...
+      )
+      return 1
+
+  new_authorized_networks_needed = args.authorized_networks is not None
+
+  authorized_networks = (
+      args.authorized_networks
+      if new_authorized_networks_needed
+      else get_cluster_authorized_networks(args)
+  )
```
BluValor, line 50: Should we maybe add an additional check comparing the existing authorized networks with the ones provided by the user? It seems that this code will change the authorized networks even if the provided ones are the exact same as the ones currently existing.

src/xpk/core/cluster_private.py
```diff
@@ -0,0 +1,170 @@
...
+
+  if not new_authorized_networks_needed:
+    return_code, current_machine_already_authorized = (
+        is_current_machine_in_any_network(authorized_networks)
+    )
+    if return_code > 0:
+      xpk_print("Error checking current machine's IP address.")
+      return return_code
+    new_authorized_networks_needed = not current_machine_already_authorized
```
BluValor, line 59: This whole logic could be extracted to a separate function (like should_update_cluster_authorized_networks). This would make it clearer, moreover because there is variable reassigning happening here (new_authorized_networks_needed can change).

src/xpk/core/cluster_private.py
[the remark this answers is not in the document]
sharabiani: This will couple core module to a UI module. I would recommend to refactor other places in core to receive args as arguments if needed.

src/xpk/core/cluster_private.py
sharabiani: If users set this, it wouldn't be problem to update it.

src/xpk/core/cluster_private.py
sharabiani: done

src/xpk/core/cluster_private.py
sharabiani: done with a slightly difference

pawloch00 (approved):

pawloch00 (approved):

src/xpk/core/cluster_private.py
```diff
@@ -51,36 +80,16 @@ def authorize_private_cluster_access_if_necessary(args) -> int:
...
-    )
-    if cluster_authorized_networks_update_code > 0:
-      xpk_print('Updating cluster authorized networks failed!')
-      return cluster_authorized_networks_update_code
-    xpk_print("Cluster's master authorized networks updated successfully.")
-  else:
-    xpk_print('Current machine is already in authorized networks.')
-
-  return 0
+  return 0, authorized_networks if new_authorized_networks_needed else None
```
BluValor, line 92: I still don't like this piece of code as a whole - it's convoluted, some logic happens inside the get_authorized_networks_to_update function but still we have to handle the adding of the current network outside of it? My intention was - if this can't be any cleaner at least encapsulate it as a whole in a separate function. 

This should be a simple piece of logic:
- get networks if there are new, use the old ones otherwise
- check if the current network is in the desired networks, add it if not
- update or not the networks

Currently it doesn't really read like this. Can we divide the logic to these readable blocks?

Sorry about the nagging, I just had to think about this a whole lot just to understand it, and it shouldn't be this hard :(

BluValor (approved):

pawloch00 (approved):

BluValor (approved):

pawloch00 (approved):

PR #228, opened 2024-11-06
xpk shell command

BluValor:
## Fixes / Features
- add interactive pod template
- add interactive mode to app profile

Since the storage branches (development branch) are not yet merged with the main and they lack this type of workflow connecting the shell to existing storage will have to be done in followup task, after these features are usable at the same time.

## Testing / Documentation
Testing details.

- [ y/n ] Tests pass


src/xpk/commands/shell.py
```diff
@@ -0,0 +1,122 @@
...
+  pod_names = shell_name.strip().split('\n')
+  kjob_pod_names = [
+      name for name in pod_names if AppProfileDefaults.NAME.value in name
+  ]
+  shell_pod_names = [name for name in kjob_pod_names if 'interactive' in name]
+
+  return shell_pod_names[0] if shell_pod_names else None
+
+
+exit_instructions = 'To exit the shell input "exit".'
```
pawloch00, line 69: move it to upper part of file

src/xpk/commands/shell.py
```diff
@@ -0,0 +1,122 @@
...
+  Returns:
+    0 if successful and 1 otherwise.
+  """
+  exisitng_shell_pod_name = get_existing_shell_pod_name(args)
+
+  if exisitng_shell_pod_name is None:
+    xpk_print('There is no shell running to stop')
+    xpk_exit(0)
+
+  return_code = run_command_with_updates(
```
pawloch00, line 111: Are we sure we should delete shell on stop command? We can't run it again? Can this shell persist data between user sessions?

src/xpk/commands/shell.py
```diff
@@ -0,0 +1,122 @@
...
+      command=f'kubectl delete pod {exisitng_shell_pod_name}',
+      task='Deleting the existing shell.',
+      global_args=args,
+  )
+
+  if return_code != 0:
+    xpk_print(f'The command failed with code {return_code}.')
```
pawloch00, line 118: Which command? Too generic for log

BluValor: Actually the command is gonna log the task name, so the print is unnecessary here. Will remove it and only exit.

src/xpk/commands/shell.py
pawloch00: ok, but function is named shell stop and the docstring says stop the running interactive shell. Should't it have delete mentioned in name or docstring?

.github/workflows/nightly_tests.yaml
```diff
@@ -40,6 +40,8 @@ jobs:
     - uses: actions/setup-python@v5
       with:
         python-version: '3.10'
+    - name: Install expect package
```
pawloch00, line 43: Should this package be mentioned in readme as dependency? Should it be added to makefile? If it is not crucial I would prefer not to add it

BluValor: It's only for testing, it shouldn't be added as dependency.

PR #229, opened 2024-11-06
Wait for kueue to be available before install custom resources.

mbobrovskyi:
## Fixes / Features
- Wait for kueue to be available before install custom resources.

## Testing / Documentation
Testing details.

- [ y/n ] Tests pass

### DIFF
--- a/src/xpk/core/kueue.py
+++ b/src/xpk/core/kueue.py
@@ -186,6 +186,26 @@ def install_kueue_on_cluster(args) -> int:
   return return_code
 
 
+def wait_for_kueue_available(args: Namespace) -> int:
+  """Wait for Kueue to be fully available.
+
+  Args:
+    args: user provided arguments for running the command.
+
+  Returns:
+    0 if successful and 1 otherwise.
+  """
+  command = (
+      'kubectl wait deploy/kueue-controller-manager -nkueue-system'
+      ' --for=condition=available --timeout=5m'
+  )
+  task = 'Wait for Kueue to be available'
+  return_code = run_command_with_updates_retry(command, task, args)
+  if return_code != 0:
+    xpk_print(f'{task} returned ERROR {return_code}')
+  return return_code
+
+
 def install_kueue_crs(
     args,
     system: SystemCharacteristics,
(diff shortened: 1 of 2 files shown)


src/xpk/core/kueue.py
```diff
@@ -185,6 +185,23 @@ def install_kueue_on_cluster(args) -> int:
     xpk_print(f'{task} returned ERROR {return_code}')
   return return_code
 
+def wait_for_kueue_available(args) -> int:
```
pawloch00, line 188: I think it may be good place to raise an Exception instead of returning int. Generally we want to avoid returning int and raise Exceptions where possible

pawloch00: Please also note that new code should be type-hinted where possible

mbobrovskyi: > I think it may be good place to raise an Exception instead of returning int. Generally we want to avoid returning int and raise Exceptions where possible

Do we have some error handler to use this solution with exception?

mbobrovskyi: I can see some error handler and custom exception with error code. But not sure that it is a scope of current task. WDYT?

mbobrovskyi: > Please also note that new code should be type-hinted where possible

Done

mbobrovskyi:
cc @mwysokin

mbobrovskyi:
cc @IrvingMg

mbobrovskyi:
Anybody know why this error happens?

pawloch00:
This repo does not allow for forks, that's why it is failing

mbobrovskyi:
> This repo does not allow for forks, that's why it is failing. Please switch to local branch and than we will try to fix it.

I don't have permissions :(

pawloch00:
Please give it a try now

mbobrovskyi:
> Please give it a try now

Thanks!

pawloch00:
Please let me know if you can work from local branch now

mbobrovskyi:
> Please let me know if you can work from local branch now

Thank you! Now it works.

mbobrovskyi:
Due to switching to local branch.

https://github.com/AI-Hypercomputer/xpk/pull/235

mbobrovskyi:
Due to switching to local branch.

https://github.com/AI-Hypercomputer/xpk/pull/235

PR #231, opened 2024-11-07
Add flags on batch command.

mbobrovskyi:
## Fixes / Features
- Add flags on batch command.

## Testing / Documentation
Testing details.

- [ y/n ] Tests pass

### DIFF
--- a/src/xpk/commands/batch.py
+++ b/src/xpk/commands/batch.py
@@ -15,6 +15,8 @@ limitations under the License.
 """
 
 from argparse import Namespace
+
+from build.lib.xpk.core.kueue import LOCAL_QUEUE_NAME
 from ..utils import xpk_exit, xpk_print
 from .cluster import set_cluster_command
 from ..core.core import add_zone_and_project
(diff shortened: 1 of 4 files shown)


mbobrovskyi:
cc @IrvingMg

src/xpk/commands/batch.py
```diff
@@ -46,11 +48,66 @@ def batch(args: Namespace) -> None:
 
 def submit_job(args: Namespace) -> None:
```
IrvingMg, line 49: Maybe we're missing `--init-image` or don't we need it?

IrvingMg, line 49: Probably we also need to have `--skip-localqueue-validation` and `--skip-priority-validation` flags?

mbobrovskyi: Yeah, this is a question. Do we need it?

mbobrovskyi: WDYT? @44past4 @mwysokin

mbobrovskyi: If localqueue owned by us we don't need to check it. On the other hand user could remove it. So not sure about it to. 

@44past4 @mwysokin @pawloch00 WDYT?

mbobrovskyi: For about priority validation the same question. Do we need it?

@44past4 @mwysokin @pawloch00 WDYT?

pawloch00: There is only one localqueue defined during cluster creation that is used by xpk. I don't think we should allow user to perform any checks of changes on lq.

pawloch00: But lets have @44past4  for final decision

44past4: The main assumption in XPK is that this should be highly opinionated tool with limited number of options available for users. Therefore in this case I would assume that we own the local queue and priorities and users should not edit those in any way.

mbobrovskyi: Thanks!

44past4: I see limited value in the option to override this image when running a workload. Therefore for now I would skip this option.

mbobrovskyi: Thanks!

mbobrovskyi:
Due to switching to local branch.

https://github.com/AI-Hypercomputer/xpk/pull/236

mbobrovskyi:
Due to switching to local branch.

https://github.com/AI-Hypercomputer/xpk/pull/236

PR #233, opened 2024-11-07
Add Cloud Platform scope to NodePool Creation by default on Pathways + CPU

SujeethJinesh:
## Fixes / Features
- This PR adds the Cloud Platform access scope to keep consistency between pathways and McJAX NodePool creation. This will get rid of the warning below and make it easy for us to get goodput values by default. 

Screenshot 2024-11-07 at 2 47 22 PM

## Testing / Documentation
N/A

Testing details.

- [ y ] Tests pass


dipannita08 (approved):
Thanks!

pawloch00:
Please use cloud-platform instead of url

src/xpk/core/core.py
```diff
@@ -1500,7 +1500,7 @@ def run_gke_node_pool_create_command(
           f' --region={zone_to_region(args.zone)}'
           ' --num-nodes=1'
           f' --machine-type={args.pathways_gce_machine_type}'
-          ' --scopes=storage-full,gke-default'
+          ' --scopes=storage-full,gke-default,"https://www.googleapis.com/auth/cloud-platform"'
```
pawloch00, line 1503: Please use cloud-platform instead of url

SujeethJinesh: We currently use the full cloud-platform link in other parts of the same file. Should we keep it consistent or replace it (need to verify `cloud-platform` on its own also works).

E.g. 
https://github.com/AI-Hypercomputer/xpk/blob/main/src/xpk/core/core.py#L1449

and

https://github.com/AI-Hypercomputer/xpk/blob/main/src/xpk/core/core.py#L1460

shauryagup (approved):

src/xpk/core/core.py
SujeethJinesh: Done. Replaced with constant URL instead. Thanks!

pawloch00 (approved):

pawloch00:
Does warning you pasted in Readme occurs to every user using Pathways, or just in your case? (Sorry, I am not good with pathways yet)

SujeethJinesh:
> Does warning you pasted in Readme occurs to every user using Pathways, or just in your case? (Sorry, I am not good with pathways yet)

No worries! This issue would appear for anytime goodput is enabled and the cloud-platform access scope is not present, which is a great reminder and thanks @dipannita08 for pointing it out as well, that we should make this scope default for CPU accelerator types as well (added now). Thanks! I'll need another LGTM.

dipannita08 (approved):

pawloch00 (approved):

dipannita08 (approved):

PR #227, opened 2024-11-06
Add type checking to pre-commit and to build pipeline

pawloch00:
## Fixes / Features
- add pytype for type checking
- add modules are removed right now, new modules will be tests
- add pytype check as part of pre-commit 

## Testing / Documentation
Testing details.

- [ y/n ] Tests pass
- [ y/n ] Appropriate changes to documentation are included in the PR

PR #238, opened 2024-11-08
Fix pytype

pawloch00:
## Fixes / Features
-
-

## Testing / Documentation
Testing details.

- [ y/n ] Tests pass

### DIFF
--- a/.github/workflows/lint_and_format.yml
+++ b/.github/workflows/lint_and_format.yml
@@ -22,7 +22,9 @@ on:
     - synchronize
     branches:
     - main
-
+  push:
+    branches:
+      - main
 jobs:
   build-and-test:
     runs-on: ubuntu-latest
--- a/.github/workflows/lint_and_format.yml
+++ b/.github/workflows/lint_and_format.yml
@@ -36,6 +36,7 @@ jobs:
         python -m pip install pyink==24.3.0
         pip install pytype
         pip install pylint
+        pip install .
     - name: Check the code with pylint
       run: |
         pylint $(git ls-files '*.py')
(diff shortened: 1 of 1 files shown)


.github/workflows/lint_and_format.yml
```diff
@@ -36,6 +36,7 @@ jobs:
         python -m pip install pyink==24.3.0
         pip install pytype
         pip install pylint
+        pip install .
```
mbobrovskyi, line 39: Hmm. But why didn’t we catch this error in the original PR?

mbobrovskyi: I also can see linter errors on main branch.

pawloch00: Not sure, probably I tried it on file not importing urllib

pawloch00: What kind of errors? Aim of this change and previous PR is to add linter checks for new code. For existing code type hints will be corrected in separate PR

mbobrovskyi: I mean after push changes github action didn't show us this error.

mbobrovskyi: Probably we need to revert `push`. https://github.com/AI-Hypercomputer/xpk/pull/227/files#diff-9be398388c4a96355ba1dcbf7c41cee8567621b5e50c1693da04a5639394ce42L17

pawloch00: yep, aim of this pr is to remove error comming from missing urllib dependency

pawloch00: I see, probably we need to fix it, before we have each check run twice(first time with pull request trigger and second time with push trigger)

mbobrovskyi: Also, we need to check the trigger for merges to the main branch.

mbobrovskyi (approved):
LGTM. Thanks for fix!

PR #224, opened 2024-10-29
Making CloudDNS optional for Pathways-enabled clusters.

RoshaniN:
## Fixes / Features
- Makes Clouddns optional for Pathways XPK clusters. (Regular XPK clusters do not use clouddns at all.)

## Testing / Documentation
Testing details.

- [ y/n ] Tests pass
- [ y/n ] Appropriate changes to documentation are included in the PR

PR #241, opened 2024-11-08
Disabling CloudDNS upgrades, while ensuring backward compatibility.

RoshaniN:
## Fixes / Features
- Disabling CloudDNS upgrades until issues with the upgrade can be fixed.
- Maintaining backward compatibility by letting existing Pathways clusters using CloudDNS to function as expected.
- Will enable CloudDNS using a flag after b/350556475 is fixed via https://github.com/AI-Hypercomputer/xpk/pull/224

## Testing / Documentation
Testing details.

- [ y ] Tests pass


src/xpk/commands/workload.py
```diff
@@ -327,8 +327,7 @@ def workload_create(args) -> None:
 
   if args.headless and not is_cluster_using_clouddns(args):
     xpk_print(
-        'Please run xpk cluster create-pathways first, to upgrade and enable'
-        ' CloudDNS on your cluster.'
+        'Please use kubectl port forwarding to connect to the Pathways proxy.'
     )
     xpk_exit(1)
```
shauryagup, line 332: Any reason to continue to exit here? Can we just get rid of this check and have the port forwarding instructions printed out when starting headless workload?

```
kubectl port-forward pod/<proxy-pod-name> <proxy-port>:<proxy-port>
JAX_PLATFORMS=proxy JAX_BACKEND_TARGET=grpc://127.0.0.1:<proxy-port> python -c 'import pathwaysutils; import jax; print(jax.devices())'
```

shauryagup (commented):
I see some clouddns checks in pathways.py as well which we can remove in this PR

src/xpk/commands/workload.py
RoshaniN: Removed the stale exit and also added instructions for port forwarding.
Removed all clouddns checks (not needed for existing clouddns enabled clusters also.)
Thanks.

shauryagup (approved):

src/xpk/commands/cluster.py
```diff
@@ -500,12 +493,7 @@ def run_gke_cluster_create_command(
 
     if args.enable_pathways:
       enable_ip_alias = True
-      command += (
-          f' --create-subnetwork name={args.cluster}-subnetwork'
-          ' --cluster-dns=clouddns'
-          ' --cluster-dns-scope=vpc'
-          f' --cluster-dns-domain={args.cluster}-domain'
-      )
+      command += f' --create-subnetwork name={args.cluster}-subnetwork'
```
Obliviour, line 496: I wonder if this is still needed? do you remember what the subnetwork is for again?

Obliviour (approved):

src/xpk/commands/cluster.py
RoshaniN: The subnetwork was added to unblock Pathways workload from errors like "IP address exhausted in default network", etc. I will check separately, if it can be removed safely now and address that in another PR.

Obliviour: Ah that type of problem we should solve generally in xpk, it also occurs in the normal non-pathways case too! Maybe a new xpk argument --create-network?

Obliviour: Otherwise we get conflicts like https://buganizer.corp.google.com/issues/378732863 where we need a different network

PR #242, opened 2024-11-12
Integrate kind for local testing

IrvingMg:
## Fixes / Features
- Integrate kind for Local Testing

As part of the kjob integration in #212, we are developing new features where having a full cluster on GKE with GPUs, Vertex AI, etc., feels like overkill. To simplify the development and testing process, this PR implements a local testing environment using kind.

This PR adds a new command to xpk for managing local Kubernetes clusters with kind, as well as the --kind-cluster flag for the batch command. The command to create a local cluster looks like this:

```shell
python3 xpk.py kind create --cluster xpk-test
```

Once the local cluster is set up, you can use the --kind-cluster flag to run xpk commands against the local cluster instead of GKE. For example:

```shell
python3 xpk.py batch [other-options] --kind-cluster
```

This PR implements both the command for creating and managing local kind clusters and the --kind-cluster flag specifically for the batch command, which interacts with kjob and is well-suited for testing the local environment. Support for local testing may be extended to other xpk commands as needed.

For more details on how to set up and use the local testing environment, please refer to the updated README.

## Testing / Documentation
Testing details.

- [ y/n ] Tests pass

### DIFF
--- a/src/xpk/commands/batch.py
+++ b/src/xpk/commands/batch.py
@@ -21,10 +21,8 @@ from ..core.core import add_zone_and_project
 from ..core.job_template import create_job_template_instance
 from ..core.app_profile import create_app_profile_instance
 from ..core.app_profile import APP_PROFILE_TEMPLATE_DEFAULT_NAME
-from ..core.commands import (
-    run_command_for_value,
-    run_command_with_updates,
-)
+from ..core.commands import run_command_for_value
+from .kind import set_local_cluster_command
 
 
 def batch(args: Namespace) -> None:
@@ -61,37 +59,3 @@ def submit_job(args: Namespace) -> None:
   if return_code != 0:
     xpk_print(f'Running batch job returned ERROR {return_code}')
     xpk_exit(return_code)
-
-
-def set_local_cluster_command(args) -> int:
-  """Run local cluster configuration command to set the kubectl config.
-
-  Args:
-    args: user provided arguments for running the command.
-
-  Returns:
-    0 if successful and 1 otherwise.
-  """
-  if not args.cluster:
-    command = 'kubectl config current-context'
-    return_code, current_context = run_command_for_value(
-        command, 'get current-context', args
-    )
-    xpk_print(
-        'No local cluster name specified. Using current-context'
-        f' `{current_context.strip()}`'
-    )
-    return return_code
-
-  command = (
-      f'kubectl config use-context kind-{args.cluster} --namespace=default'
-  )
-  task = f'switch to cluster {args.cluster}'
-  return_code = run_command_with_updates(
-      command,
-      task,
-      args,
-  )
-  if return_code != 0:
-    xpk_print(f'{task} returned ERROR {return_code}')
-  return return_code
--- a/src/xpk/commands/batch.py
+++ b/src/xpk/commands/batch.py
@@ -23,6 +23,7 @@ from ..core.app_profile import create_app_profile_instance
 from ..core.app_profile import APP_PROFILE_TEMPLATE_DEFAULT_NAME
 from ..core.commands import (
     run_command_for_value,
+    run_command_with_updates,
 )
 
 
@@ -34,11 +35,14 @@ def batch(args: Namespace) -> None:
   Returns:
     None
   """
-  if not args.local_test:
+  if not args.kind_cluster:
     add_zone_and_project(args)
     set_cluster_command_code = set_cluster_command(args)
-    if set_cluster_command_code != 0:
-      xpk_exit(set_cluster_command_code)
+  else:
+    set_cluster_command_code = set_local_cluster_command(args)
+
+  if set_cluster_command_code != 0:
+    xpk_exit(set_cluster_command_code)
 
   create_job_template_instance(args)
   create_app_profile_instance(args)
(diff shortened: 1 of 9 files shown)


IrvingMg:
cc: @mbobrovskyi @mwysokin

src/xpk/commands/kind.py
mbobrovskyi, line 1: I'm also thinking creating a separate Python file to execute all the commands we have, plus kind. Is that possible?

pawloch00:
I doubt if we should merge it into main. Let me ask about it

src/xpk/commands/kind.py
IrvingMg: I think is a good idea to have a separate file like `xpk-test.py` to run commands only on local clusters but there might be some trade-offs to consider. 

It would be good to have the opinion of other teammates as well.

pawloch00:
In my opinion merging it to main with only batch command being executed localy is not worth it. If more commands can be added and kind can be used for ex. integration tests in github pipeline than I approve it

IrvingMg:
> In my opinion merging it to main with only batch command being executed localy is not worth it. If more commands can be added and kind can be used for ex. integration tests in github pipeline than I approve it

I agree it might seem not worth it as we're only adding the option of local testing for batch command. However, I think this may be a good opportunity to introduce local testing. To cover all the cases for local testing in a single PR seems to be a big effort, so I think we can add support for other commands bit by bit.

Besides that, currently, we have a good couple of examples - #236, #244 - where having the option of testing locally is useful. As these features don't require to have a cluster with special resources such as GPU, TPU, etc., I think we can avoid creating a cluster on GKE, saving time and costs, for testing by using a local cluster.

pawloch00:
OK, it would be nice to  create integration test using kind as followup task for this PR and use it in github pipeline where possible

pawloch00 (dismissed):

src/xpk/commands/batch.py
```diff
@@ -34,10 +34,11 @@ def batch(args: Namespace) -> None:
   Returns:
     None
   """
-  add_zone_and_project(args)
-  set_cluster_command_code = set_cluster_command(args)
-  if set_cluster_command_code != 0:
-    xpk_exit(set_cluster_command_code)
+  if not args.local_test:
```
mbobrovskyi, line 37: Can we also add it to other commands?

src/xpk/parser/batch.py
```diff
@@ -36,6 +38,13 @@ def set_batch_parser(batch_parser):
       default=None,
       help='Cluster to which command applies.',
   )
+  batch_optional_arguments.add_argument(
+      '--local-test',
```
mbobrovskyi, line 42:
```suggestion
      '--kind-cluster',
```
Maybe like this? And also name of cluster.

src/xpk/commands/batch.py
IrvingMg: I think adding it to other commands requires more changes in the base code. So I'd suggest keeping this PR only as the introduction of Kind and then, as a follow up we can add it to other commands.

src/xpk/parser/batch.py
IrvingMg: Done.

src/xpk/commands/batch.py
```diff
@@ -56,3 +61,37 @@ def submit_job(args: Namespace) -> None:
...
+  Returns:
+    0 if successful and 1 otherwise.
+  """
+  if not args.cluster:
+    command = 'kubectl config current-context'
+    return_code, current_context = run_command_for_value(
+        command, 'get current-context', args
+    )
+    xpk_print(
+        'No local cluster name specified. Using current-context'
```
mbobrovskyi, line 81:
```suggestion
        'No kind cluster name specified. Using current-context'
```

src/xpk/commands/batch.py
```diff
@@ -56,3 +61,37 @@ def submit_job(args: Namespace) -> None:
   if return_code != 0:
     xpk_print(f'Running batch job returned ERROR {return_code}')
     xpk_exit(return_code)
+
+
+def set_local_cluster_command(args) -> int:
```
mbobrovskyi, line 66:
```suggestion
def set_kind_cluster_command(args) -> int:
```

src/xpk/commands/batch.py
```diff
@@ -56,3 +61,37 @@ def submit_job(args: Namespace) -> None:
...
+  """Run local cluster configuration command to set the kubectl config.
```
mbobrovskyi, line 67:
```suggestion
  """Run kind cluster configuration command to set the kubectl config.
```

src/xpk/commands/batch.py
mbobrovskyi: Yeah, but for job I think we can add.

IrvingMg: On hold due to #265

IrvingMg: --kind-cluster flag added to `job` command.

mbobrovskyi (approved):

IrvingMg:
> OK, it would be nice to create integration test using kind as followup task for this PR and use it in github pipeline where possible

Opened issue for integration test #267

PR #244, opened 2024-11-13
Add job list command

IrvingMg:
## Fixes / Features
- Add job list command

## Testing / Documentation
Testing details.

- [ y/n ] Tests pass

### DIFF
--- a/.github/workflows/build_tests.yaml
+++ b/.github/workflows/build_tests.yaml
@@ -85,7 +85,7 @@ jobs:
     - name: Run a batch job on the cluster
       run: python3 xpk.py batch --cluster $TPU_CLUSTER_NAME --zone=us-central2-b test.sh
     - name: List out the Slurm jobs on the cluster
-      run: python3 xpk.py job ls
+      run: python3 xpk.py job ls | grep 'xpk-def-app-profile-slurm-'
     - name: Delete the cluster created
       if: always()
       run: python xpk.py cluster delete --cluster $TPU_CLUSTER_NAME --zone=us-central2-b
--- a/.github/workflows/build_tests.yaml
+++ b/.github/workflows/build_tests.yaml
@@ -82,6 +82,10 @@ jobs:
       run: python3 xpk.py workload delete --workload $WORKLOAD_NAME --cluster $TPU_CLUSTER_NAME --zone=us-central2-b
     - name: Delete the Pathways workload on the cluster
       run: python3 xpk.py workload delete --workload $PATHWAYS_WORKLOAD_NAME --cluster $TPU_CLUSTER_NAME --zone=us-central2-b
+    - name: Run a batch job on the cluster
+      run: python3 xpk.py batch --cluster $TPU_CLUSTER_NAME --zone=us-central2-b test.sh
+    - name: List out the Slurm jobs on the cluster
+      run: python3 xpk.py job ls
     - name: Delete the cluster created
       if: always()
       run: python xpk.py cluster delete --cluster $TPU_CLUSTER_NAME --zone=us-central2-b
--- a/.github/workflows/build_tests.yaml
+++ b/.github/workflows/build_tests.yaml
@@ -84,7 +84,7 @@ jobs:
       run: python3 xpk.py workload delete --workload $PATHWAYS_WORKLOAD_NAME --cluster $TPU_CLUSTER_NAME --zone=us-central2-b
     - name: Run a batch job on the cluster
       run: python3 xpk.py batch --cluster $TPU_CLUSTER_NAME --zone=us-central2-b test.sh
-    - name: List out the Slurm jobs on the cluster
+    - name: List out the jobs on the cluster
       run: python3 xpk.py job ls | grep 'xpk-def-app-profile-slurm-'
     - name: Delete the cluster created
       if: always()
--- a/README.md
+++ b/README.md
@@ -604,6 +604,24 @@ when creating the workload otherwise the workload will always finish with `Compl
     `125`: Workload finished but did not complete successfully.
     `1`: Other failure.
 
+## Job List
+
+*   Job List (see Slurm jobs):
+
+    ```shell
+    python3 xpk.py job ls 
+    ```
+
+* Example Job List Output:
+
+  ```
+    NAME                              PROFILE               LOCAL QUEUE   COMPLETIONS   DURATION   AGE
+    xpk-def-app-profile-slurm-74kbv   xpk-def-app-profile                 1/1           15s        17h
+    xpk-def-app-profile-slurm-brcsg   xpk-def-app-profile                 1/1           9s         3h56m
+    xpk-def-app-profile-slurm-kw99l   xpk-def-app-profile                 1/1           5s         3h54m
+    xpk-def-app-profile-slurm-x99nx   xpk-def-app-profile                 3/3           29s        17h
+  ```
+
 ## Inspector
 * Inspector provides debug info to understand cluster health, and why workloads are not running.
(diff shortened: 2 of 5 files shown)


src/xpk/parser/job.py
```diff
@@ -0,0 +1,47 @@
...
+  )
+
+  ### "job ls" command parser ###
+  job_list_parser = job_subcommands.add_parser('ls', help='List Slurm jobs.')
+
+  job_list_optional_arguments = job_list_parser.add_argument_group(
+      'Optional Arguments', 'Arguments optional for job list.'
+  )
+
+  job_list_optional_arguments.add_argument(
```
pawloch00, line 38: I don't think there is possiblity to definie custom namespace in xpk, but it needs to be checked.

IrvingMg: That's correct. Setting up the default namespace to something other than “default” will be provided in a separate task.

pawloch00: But this namespace should be created somewhere  and I don't think it is possible now. (I don't know about such plans too :P)

IrvingMg: @mwysokin would you mind confirming that namespace configuration is a future feature that will be required?

pawloch00: After consultation I --namespace option should be removed now

IrvingMg: Done.

IrvingMg:
> Can you please add tests to .github workflows including your change?

Done.

.github/workflows/build_tests.yaml
```diff
@@ -88,6 +88,10 @@ jobs:
       run: python3 xpk.py workload delete --workload $WORKLOAD_NAME --cluster $TPU_CLUSTER_NAME --zone=us-central2-b
     - name: Delete the Pathways workload on the cluster
       run: python3 xpk.py workload delete --workload $PATHWAYS_WORKLOAD_NAME --cluster $TPU_CLUSTER_NAME --zone=us-central2-b
+    - name: Run a batch job on the cluster
+      run: python3 xpk.py batch --cluster $TPU_CLUSTER_NAME --zone=us-central2-b test.sh
+    - name: List out the Slurm jobs on the cluster
+      run: python3 xpk.py job ls
```
pawloch00, line 94: please add an assert(grep) to check output sanity

src/xpk/commands/job.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+def job_list(args) -> None:
+  """Function around job list.
+
+  Args:
+    args: user provided arguments for running the command.
+
+  Returns:
+    0 if successful and 1 otherwise.
```
mbobrovskyi, line 32:
```suggestion
    None
```

src/xpk/commands/job.py
```diff
@@ -0,0 +1,50 @@
...
+  """
+  add_zone_and_project(args)
+  xpk_print(f'For project {args.project} and zone {args.zone}:', flush=True)
+  print(f'args: {args.namespace}')
```
mbobrovskyi, line 36: Is it for debug? Can we remove it?

src/xpk/commands/job.py
```diff
@@ -0,0 +1,50 @@
...
+
+def run_slurm_job_list_command(args) -> None:
+  cmd = (
+      f'kubectl-kjob list slurm  --profile {APP_PROFILE_TEMPLATE_DEFAULT_NAME}'
+  )
+
+  return_code = run_command_with_updates(cmd, 'list slurm jobs', args)
+  if return_code != 0:
+    xpk_print(f'Listing slurm jobs returned ERROR {return_code}')
+    xpk_exit(return_code)
```
mbobrovskyi, line 50:
```suggestion
  xpk_exit(return_code)
```

src/xpk/commands/job.py
mbobrovskyi, line 1: Do we need to add some info about this command on README?

IrvingMg: README updated

README.md
```diff
@@ -600,6 +600,24 @@ when creating the workload otherwise the workload will always finish with `Compl
     `125`: Workload finished but did not complete successfully.
     `1`: Other failure.
 
+## Job List
+
+*   Job List (see Slurm jobs):
```
mbobrovskyi, line 605: Where I can see? 🤔

IrvingMg: Maybe I misunderstood but the text in () is just a very short description of what Job List does.

IrvingMg: Sure. Should I remove it as well from the command description?

mbobrovskyi: I think so.

IrvingMg: Done.

mbobrovskyi: Thanks!

.github/workflows/build_tests.yaml
```diff
@@ -88,6 +88,10 @@ jobs:
       run: python3 xpk.py workload delete --workload $WORKLOAD_NAME --cluster $TPU_CLUSTER_NAME --zone=us-central2-b
     - name: Delete the Pathways workload on the cluster
       run: python3 xpk.py workload delete --workload $PATHWAYS_WORKLOAD_NAME --cluster $TPU_CLUSTER_NAME --zone=us-central2-b
+    - name: Run a batch job on the cluster
+      run: python3 xpk.py batch --cluster $TPU_CLUSTER_NAME --zone=us-central2-b test.sh
+    - name: List out the jobs on the cluster
+      run: python3 xpk.py job ls | grep 'xpk-def-app-profile-slurm-'
```
mbobrovskyi, line 94: I think we should fail test if something not correct.

IrvingMg: As I understand, grep will exit with code 1 if the pattern `xpk-def-app-profile-slurm-'` is not found which will make the test failing

mbobrovskyi: Ah I see. Thanks!

mbobrovskyi (dismissed):
LGTM. Thanks!

src/xpk/commands/job.py
```diff
@@ -0,0 +1,50 @@
...
+  """Function around job list.
+
+  Args:
+    args: user provided arguments for running the command.
+
+  Returns:
+    None
+  """
+  add_zone_and_project(args)
+  xpk_print(f'For project {args.project} and zone {args.zone}:', flush=True)
```
pawloch00, line 35: Listing jobs for project ...

mbobrovskyi (approved):

pawloch00 (approved):

IrvingMg:
Cluster deletion is failing in the pipeline. I think it might be because I need to remove the job created before deleting the cluster. The job deletion is done in #248. 

Maybe I should skip the test here and add both tests - ls and cancel - in #248? @pawloch00 WDYT?

pawloch00:
> Cluster deletion is failing in the pipeline. I think it might be because I need to remove the job created before deleting the cluster. The job deletion is done in #248.
> 
> Maybe I should skip the test here and add both tests - ls and cancel - in #248? @pawloch00 WDYT?

I think cluster is upgrading and it has nothing to do with your PR. Please wait and try to run check, I will try to delete cluster manually.

IrvingMg:
> > Cluster deletion is failing in the pipeline. I think it might be because I need to remove the job created before deleting the cluster. The job deletion is done in #248.
> > Maybe I should skip the test here and add both tests - ls and cancel - in #248? @pawloch00 WDYT?
> 
> I think cluster is upgrading and it has nothing to do with your PR. Please wait and try to run check, I will try to delete cluster manually.

Got it. Thanks!

PR #294, opened 2024-12-12
Print nodepools info when describing cluster

IrvingMg:
## Fixes / Features
- Add nodepools info to cluster describe command
-

## Testing / Documentation
Testing details.

- [ y/n ] Tests pass

### DIFF
--- a/.github/workflows/build_tests.yaml
+++ b/.github/workflows/build_tests.yaml
@@ -146,6 +146,8 @@ jobs:
       run: python xpk.py cluster create-pathways --cluster $TPU_CLUSTER_NAME --private --tpu-type=$TPU_TYPE  --num-slices=2 --zone=us-central2-b --default-pool-cpu-machine-type=n1-standard-16 --default-pool-cpu-num-nodes=16 --reservation='${{ secrets.GCP_TPU_V4_RESERVATION }}' --custom-cluster-arguments="${CLUSTER_ARGUMENTS}"
     - name: Verify the created cluster is private
       run: gcloud container clusters describe $TPU_CLUSTER_NAME --region=us-central2 --format="value(privateClusterConfig.enablePrivateNodes)" | grep 'True' || (echo 'The created cluster is not private.' && exit 1)
+    - name: List out the nodepools on the cluster
+      run: python xpk.py cluster describe --cluster $TPU_CLUSTER_NAME --zone=us-central2-b | grep -P '^(?=.*NODEPOOL_NAME)(?=.*SLICE)(?=.*TYPE)(?=.*EXPECTED_HEALTHY_NODES)(?=.*ACTUAL_HEALTHY_NODES)(?=.*TOTAL_NODES)'
     - name: Authenticate Docker
       run: gcloud auth configure-docker --quiet
     - name: Create test script to execute in workloads
--- a/src/xpk/commands/cluster.py
+++ b/src/xpk/commands/cluster.py
@@ -377,7 +377,67 @@ def nodepools_build_table(args) -> tuple[int, list[list]]:
       'TOTAL_NODES',
   ]]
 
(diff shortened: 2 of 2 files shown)


IrvingMg:
cc: @mbobrovskyi

mbobrovskyi (approved):
LGTM. Thanks!

.github/workflows/build_tests.yaml
```diff
@@ -123,6 +123,8 @@ jobs:
       run: python xpk.py cluster create-pathways --cluster $TPU_CLUSTER_NAME --private --tpu-type=v4-8  --num-slices=2 --zone=us-central2-b --default-pool-cpu-machine-type=n1-standard-16 --default-pool-cpu-num-nodes=16 --reservation='${{ secrets.GCP_TPU_V4_RESERVATION }}' --custom-cluster-arguments="${CLUSTER_ARGUMENTS}"
     - name: Verify the created cluster is private
       run: gcloud container clusters describe $TPU_CLUSTER_NAME --region=us-central2 --format="value(privateClusterConfig.enablePrivateNodes)" | grep 'True' || (echo 'The created cluster is not private.' && exit 1)
+    - name: List out the nodepools on the cluster
+      run: python xpk.py cluster describe --cluster $TPU_CLUSTER_NAME --zone=us-central2-b | grep -P '^(?=.*NODEPOOL_NAME)(?=.*SLICE)(?=.*TYPE)(?=.*EXPECTED_HEALTHY_NODES)(?=.*ACTUAL_HEALTHY_NODES)(?=.*TOTAL_NODES)'
```
sharabiani, line 127: the grep filter only prints out the headers.

sharabiani (commented):
- Could we reuse 'cluster inspect'? Maybee could provide some help description when they are not healthy? such as:
*Check the pantheon ui (link to it) to see if there are any errors or maybe the node pool is still being created
*Check the troubleshooting guide to unblock the cluster
- @Obliviour does the output meets your expectation?

.github/workflows/build_tests.yaml
IrvingMg: The idea is just to check the headers of the table to verify that we have data. I've updated the output to don't print the table when there is no info to show.

IrvingMg:
> * Could we reuse 'cluster inspect'? Maybee could provide some help description when they are not healthy? such as:
>   *Check the pantheon ui (link to it) to see if there are any errors or maybe the node pool is still being created
>   *Check the troubleshooting guide to unblock the cluster
> * @Obliviour does the output meets your expectation?

Regarding reusing `cluster inspect`, actually we're using some commands from there to get the info shown here but we have a requirement to replace the node pool list in `cluster describe` command with a table.

IrvingMg:
@sharabiani @pawloch00 any thoughts on this PR?

src/xpk/commands/cluster.py
```diff
@@ -362,6 +367,145 @@ def cluster_describe(args) -> None:
   xpk_exit(0)
 
 
+def nodepools_build_table(args) -> tuple[int, list[list]]:
```
pawloch00, line 370: please split this function into smaller ones.

src/xpk/commands/cluster.py
```diff
@@ -362,6 +367,182 @@ def cluster_describe(args) -> None:
...
+  )
+  if return_code != 0:
+    return [], return_code
+
+  return out.splitlines(), 0
+
+
+def get_actual_healthy_nodes(args) -> tuple[list[str], int]:
+  cmd_actual_healthy_nodes = (
+      'kubectl get node --no-headers=true -o'
```
pawloch00, line 507: can't we use kubernetes api for this queries? Maybe than we can do them in more elegant way than using awk?

pawloch00 (approved):

PR #297, opened 2024-12-16
A3-Mega cluster create and delete implemented

sharabiani:
## Fixes / Features
- Use Cluster Toolkit as backend for A3-Mega clusters using blueprinprint_generator, gcluster_manager, docker_manager
- Add support for reservation, default-node-pool-machine-type, default-node-pool-num-nodes
- refactor blueprinprint_generator, gcluster_manager, docker_manager

## Testing / Documentation
Testing details.

- [ y/n ] Tests pass: Manual tests passed. / Automated tests updated
- [ y/n ] To be done*


src/xpk/commands/cluster_gcluster.py
```diff
@@ -0,0 +1,123 @@
...
+
+def created_by_gcluster(args) -> bool:
+  prepare_directories()
+  unique_name = get_unique_name(args.project,zone_to_region(args.zone),args.cluster)
+  bpg = prepare_blueprint_generator()
+  return bpg.blueprint_exists(unique_name)
+
+
+def get_unique_name(project_id, region, cluster_name):
+  unique_string_hash = hash_string(input_string=f'{project_id}-{region}-{cluster_name}'.lower(), length=5)
```
pawloch00, line 81: what if cluster name contains _ ?

sharabiani: Does that make any problems?

pawloch00: won't it collide with charset in hashing function?

sharabiani: I wouldn't say so. In the hashstring fuction regardless what the input is, it will be first hashed, and the hashed integer will be convert to base36. that charset is used for converting to base36.

src/xpk/commands/cluster_gcluster.py
```diff
@@ -0,0 +1,146 @@
...
+      return bpg.generate_a3_mega_blueprint(
+          blueprint_name=blueprint_name,
+          cluster_name=args.cluster,
+          region=zone_to_region(args.zone),
+          project_id=args.project,
+          zone=args.zone,
+          auth_cidr=all_IPs_cidr,
+          num_nodes=args.num_nodes if not args.num_nodes is None else 2,
+          autoscaling_total_min_nodes=args.num_nodes
+          if not args.num_nodes is None
```
pawloch00, line 140: this line breaking is a little bit unnatural to me, maybe we can set in var before return?

src/xpk/core/blueprint/blueprint_generator.py
```diff
@@ -120,6 +127,10 @@ def generate_a3_mega_blueprint(
                 "display_name": "kubectl-access-network",
             }],
             "system_node_pool_machine_type": system_node_pool_machine_type,
+            "system_node_pool_node_count": {
+                "total_min_nodes": system_node_pool_min_node_count,
+                "total_max_nodes": 1000,
```
pawloch00, line 132: 1000 should be saced to global var

src/xpk/core/blueprint/blueprint_generator.py
```diff
@@ -133,48 +144,59 @@ def generate_a3_mega_blueprint(
...
             "initial_node_count": num_nodes,
             "zones": [zone],
             "host_maintenance_interval": "PERIODIC",
+            "reservation_affinity": reservation_affinity,
+            "run_workload_script": False,
         },
         outputs=["instructions"],
     )
-    num_chips = get_num_chips(num_nodes, a3_machine_type)
+    num_chips = num_nodes * system.chips_per_vm
```
pawloch00, line 175: what if we have more complex logic for having num_chips? Maybe we should leave it in function?

src/xpk/core/blueprint/blueprint_generator.py
```diff
@@ -133,48 +144,59 @@ def generate_a3_mega_blueprint(
...
     workload = DeploymentModule(
         id="workload_component_install",
         source="modules/management/kubectl-apply",
         use=["gke_cluster"],
         settings={
             "kueue": {
                 "install": True,
-                "config_path": f'$(ghpc_stage("{cluster_name}-a3-mega-xpk"))/kueue-xpk-configuration.yaml.tftpl',
+                "version": "v0.9.1",  # TAS feature-gates is enabled in CT
```
pawloch00, line 183: please save version is some global variable

src/xpk/core/docker_manager.py
```diff
@@ -93,6 +95,7 @@ class DockerManager(CommandRunner):
     - client (DockerClient) : docker client
     - nocache (bool) : wheter to use docker cache when building image
     - img_name (str) : name of docker image to create
+    - img_name (str) : name of docker image to create
```
pawloch00, line 98: duplicated line

pawloch00:
IIuc logs from docker container are now shown. Are they shown live? I think there is an option like docker logs --follow to show them live

src/xpk/core/docker_manager.py
```diff
@@ -163,11 +163,11 @@ def run_command(
       container = self.client.containers.run(
           image=self.img_name,
           entrypoint=cmd,
-          remove=self.rm_container_after,
-          name=self.container_name,
-          stdout=True,
-          stderr=True,
-          stream=True,
+          remove=True,
```
pawloch00, line 166: I think passing this as argument was a better way. We can use it for testing etc. Now it is hardcoded

src/xpk/core/docker_manager.py
```diff
@@ -205,19 +154,20 @@ def run_command(
...
-      self.client.containers.run(
+      container = self.client.containers.run(
           image=self.img_name,
           entrypoint=cmd,
-          remove=self.rm_container_after,
-          name=self.container_name,
-          stdout=True,
-          stderr=True,
+          remove=True,
+          name=(  # To allow multiple xpk commands run in one machine.
```
pawloch00, line 167: this can be moved to get_gcluster_container_name function

src/xpk/core/docker_manager.py
```diff
@@ -240,13 +193,13 @@ def run_command(
...
       xpk_print(f"Deploying cluster toolkit failed due to {e.explanation}")
       xpk_exit(DockerRunCommandExitCode)
-
-  def _make_upload_directory(self, name: str) -> str:
-    target_path = os.path.join(self.working_dir, upload_dir, name)
-    target_dir = os.path.join(self.working_dir, upload_dir)
-    if not os.path.exists(target_dir):
-      os.mkdir(target_dir)
-    return target_path
+    finally:
```
pawloch00, line 196: Why we don't trust docker api?

src/xpk/core/docker_manager.py
sharabiani: We remove it in finally block anyways. better to let the automatic stay there too and try manual delete if it wasn't removed.
What testing scenario may we need to keep the container? Manual test or automated?

src/xpk/core/docker_manager.py
sharabiani: when detached run, I saw it wasn't removed in time before the next run. it was for when we did not have unique name though

src/xpk/core/blueprint/blueprint_generator.py
sharabiani: I wouldn't make variables for template constant values.

src/xpk/core/blueprint/blueprint_generator.py
sharabiani: I would suggest to treat this as a template constant. We cannot simply allow kueue version change in other place as the config needs to become compatible, too. Generally we can argue to move all "kueue" block including the "Install", "config_path", "version" to a global variables. The same goes with Jobset. I don't think currently we want to make it dynamic now.

src/xpk/core/docker_manager.py
pawloch00: I thought about not having remove=True, but having remove=remove_container_after(it may be set to true by default).  It may be useful in automated tests.

src/xpk/core/docker_manager.py
pawloch00: ok, never saw this in tests.

sharabiani: Did you test detach?

src/xpk/commands/cluster_gcluster.py
```diff
@@ -0,0 +1,183 @@
...
+          autoscaling_total_min_nodes=num_nodes,
+          reservation=args.reservation if args.reservation else None,
+          spot=args.spot if args.spot else False,
+          system_node_pool_machine_type=args.default_pool_cpu_machine_type,
+          system_node_pool_min_node_count=args.default_pool_cpu_num_nodes,
+      )
+  return None
+
+
+def validate_consumtion_args(args):
```
pawloch00, line 169: consumption

src/xpk/commands/cluster_gcluster.py
```diff
@@ -0,0 +1,183 @@
...
+    args_set.append('--spot')
+  if not args.on_demand is None and args.on_demand:
+    args_set.append('--on-demand')
+
+  if len(args_set) > 1:
+    xpk_print(
+        f"Error: only one of {' or '.join(args_set)} can be set at the same"
+        ' time.'
+    )
+    xpk_exit(-1)
```
pawloch00, line 183: We don't use -1 exit code anywhere else

pawloch00 (dismissed):

pawloch00 (approved):

PR #304, opened 2024-12-23
Support building docker images from non arm architecture.

mbobrovskyi:
## Fixes / Features
- Support building docker images from non arm architecture.

## Testing / Documentation
Testing details.

- [ y/n ] Tests pass

### DIFF
--- a/src/xpk/core/core.py
+++ b/src/xpk/core/core.py
@@ -1662,7 +1662,8 @@ def build_docker_image_from_base_image(args, verbose=True) -> tuple[int, str]:
   )
   tmp = write_tmp_file(docker_file)
   docker_build_command = (
-      f'docker build -f {str(tmp.file.name)} -t {docker_name} {args.script_dir}'
+      f'docker buildx build --platform=linux/amd64 -f {str(tmp.file.name)} -t'
+      f' {docker_name} {args.script_dir}'
   )
   xpk_print(f'Building {args.script_dir} into docker image.')
   return_code = run_command_with_updates(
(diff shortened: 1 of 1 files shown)


src/xpk/core/core.py
```diff
@@ -1675,7 +1675,8 @@ def build_docker_image_from_base_image(args, verbose=True) -> tuple[int, str]:
   )
   tmp = write_tmp_file(docker_file)
   docker_build_command = (
-      f'docker build -f {str(tmp.file.name)} -t {docker_name} {args.script_dir}'
+      f'docker buildx build --platform=linux/amd64 -f {str(tmp.file.name)} -t'
```
pawloch00, line 1678: Please move value of platform to variable.

pawloch00: Can we have linux/arm64 set here?

mbobrovskyi: > Can we have linux/arm64 set here?

Should we build for both platforms, or should we add a flag to allow the user to configure this? WDYT?

pawloch00: I think we may add flag to allow user to configure it, with default value set to amd64. Please test how it works with this flag set to linux/arm64, if there are any problems than we should remove it.

mbobrovskyi: It's not working with current configuration. I think we also need to have node pool with linux/arm64 to run it.

pawloch00: ok, than leave it as it is

mbobrovskyi: > Please move value of platform to variable.

Done

pawloch00 (approved):

PR #308, opened 2025-01-03
Set up JobSet failure policy on workloads

IrvingMg:
## Fixes / Features
- Handling exit codes on Workloads with JobSet failure policy. 

## Testing / Documentation
Testing details.

- [ y/n ] Tests pass

### DIFF
--- a/src/xpk/commands/workload.py
+++ b/src/xpk/commands/workload.py
@@ -36,6 +36,7 @@ from ..core.core import (
     get_gpu_scheduler,
     get_gpu_tcp_volume,
     get_gpu_volume,
+    get_main_container_docker_image,
     get_user_workload_container,
     get_volumes,
     parse_env_config,
@@ -80,6 +81,7 @@ metadata:
 spec:
   ttlSecondsAfterFinished: {args.ttl_seconds_after_finished}
   failurePolicy:
+    {failure_policy_rules}
     maxRestarts: {args.max_restarts}
   replicatedJobs:
     - name: slice-job
@@ -89,6 +91,7 @@ spec:
           parallelism: {system.vms_per_slice}    # Equal to the number of VMs per slice
           completions: {system.vms_per_slice}    # Same as the above.
           backoffLimit: 0   # When any pod fails, the job is failed
+          {pod_failure_policy}
           template:
             metadata:
               labels:
@@ -122,6 +125,7 @@ metadata:
 spec:
   ttlSecondsAfterFinished: {args.ttl_seconds_after_finished}
   failurePolicy:
+    {failure_policy_rules}
     maxRestarts: {args.max_restarts}
   replicatedJobs:
     - name: slice-job
@@ -131,6 +135,7 @@ spec:
           parallelism: {args.num_nodes}
           completions: {args.num_nodes}
           backoffLimit: 0   # When any pod fails, the job is failed
+          {pod_failure_policy}
           template:
             metadata:
               labels:
@@ -180,6 +185,7 @@ metadata:
 spec:
   ttlSecondsAfterFinished: {args.ttl_seconds_after_finished}
   failurePolicy:
+    {failure_policy_rules}
     maxRestarts: {args.max_restarts}
   replicatedJobs:
     - name: slice-job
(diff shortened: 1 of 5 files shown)


src/xpk/commands/workload.py
```diff
@@ -445,6 +453,27 @@ def workload_create(args) -> None:
     if return_code != 0:
       xpk_exit(return_code)
 
+  failure_policy_rules = ''
+  pod_failure_policy = ''
+  if args.restart_on_user_code_failure:
```
Obliviour, line 458: This is great to use failure policy!! Looking forward to this change.

I named this argument poorly -- its intent is to allow for the workload to restart additionally on user code failures when set. The range of exit codes this should specify is anything that a user would specify like 1,2,... 

Some system level exit codes that should always restart the job are 42, 143. Probably should ask GKE folks what other system level exit codes exist. These should always restart the job and should be set / or assumed in the pod failure policy to restart by default.

---


For background some customers depend on xpk to restart the job in system failure cases by default and avoid user errors (like code typos) from restarting their workload. So this is the default case.

Obliviour: To be honest, I don't think anyone actually wants a user error to restart the job so the ideal choice in my opinion is to remove the "args.restart_on_user_code_failure" argument and just "restart the job in system failure cases by default and avoid user errors (like code typos) from restarting their workload".  Since now with the failure policy support we can do this.

src/xpk/core/core.py
```diff
@@ -2026,10 +2017,6 @@ def get_main_container(args, system, docker_image, resource_type) -> str:
                   echo EXIT_CODE=$EXIT_CODE;
                   {tpu_stacktrace_terminate_command}
                   {gpu_workload_terminate_command}
-                  if [ "$EXIT_CODE" = 143 ]; then
-                    exit $EXIT_CODE
```
Obliviour, line 2030: Sorry for another confusion line of code, can you keep `exit $EXIT_CODE`

Since failure policy will handle when the job will restart, we should always return the exit code of the program back.

src/xpk/commands/workload.py
IrvingMg: Thanks for the information. Then, to sum up:

1. Default behavior should be: restart on exit codes 42 and 143.
2. Remove `restart-on-user-code-failure` flag but allow users to pass additional exit codes for restarting workload.

Is that correct? And how is the user going to pass these additional exit codes, should we have another flag for that?

src/xpk/core/core.py
IrvingMg: Got it. Thanks for the explanation.

src/xpk/commands/workload.py
IrvingMg: @Obliviour can you confirm those are the changes needed?

IrvingMg:
cc @pawloch00

pawloch00:
LGTM for me

pawloch00 (approved):

src/xpk/commands/workload.py
Obliviour: > 1. Default behavior should be: restart on exit codes 42 and 143.

Apologies for the delay - this doc looks good to follow:

https://g3doc.corp.google.com/company/gfw/support/cloud/playbooks/container-engine/container-exit-codes.md?cl=head so basically **I think any exit code between 127 and 255 is a system level failure and should restart the workload + additionally exit code 42.**


127	Command not found: The command was not found in the system's PATH or the name/path is invalid.
128	Invalid argument to exit(): The exit() function was given an invalid argument.
128 + N	Fatal error [signal N](https://github.com/torvalds/linux/blob/98369dccd2f8e16bf4c6621053af7aa4821dcf8e/include/uapi/asm-generic/signal.h#L10): The process was forcefully killed by signal N (1 < N < 32)
137	Signal 9 ([SIGKILL](https://github.com/torvalds/linux/blob/98369dccd2f8e16bf4c6621053af7aa4821dcf8e/include/uapi/asm-generic/signal.h#L20)): The process was forcefully killed. In Kubernetes, this often happens when a container exceeds its resources.limits.memory and is terminated by OOMKiller.
255	Exit status out of range: The exit() function was given an argument outside the 0-255 range.

---

2. Remove restart-on-user-code-failure flag but allow users to pass additional exit codes for restarting workload.

I agree, another flag seems fine for this.

IrvingMg: Done.

IrvingMg:
cc: @Obliviour @pawloch00

IrvingMg:
@Obliviour, @pawloch00 any thoughts on this PR?

PR #309, opened 2025-01-07
Add url to the workload logs link directly

Obliviour:
## Fixes / Features
- Adds a new url that returns the url link to the workload logs directly for the first worker of the first slice
- This allows users to easily click for the logs and especially useful since this link will work after the workload is deleted. So if a user has access to xpk logs after deleting the workload, they can still follow this link.

## Testing / Documentation
Created an xpk workload on trillium and clicked the url generated to get to the workload link. This should work on TPUs and GPUs.

- [ y ] Tests pass


gobbleturk (approved):
Thank you! This will save me 50 clicks per day on average

pawloch00 (approved):

PR #320, opened 2025-01-10
Add ttlSecondsAfterFinished to Pathways workloads

Obliviour:
## Fixes / Features
- Adds automatic deletion of workloads for pathways workloads
- This should cover all workload types now.

## Testing / Documentation

- [ y ] Tests pass

### DIFF
--- a/src/xpk/commands/workload.py
+++ b/src/xpk/commands/workload.py
@@ -214,6 +214,7 @@ metadata:
     kueue.x-k8s.io/queue-name: {local_queue_name}  # Name of the LocalQueue
     xpk.google.com/workload: {args.workload}
 spec:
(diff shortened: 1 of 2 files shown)


pawloch00:
What if the user wants workloads live longer than 12 hours? Is it the case?

Obliviour:
> What if the user wants workloads live longer than 12 hours? Is it the case?

@pawloch00  the workload flag --ttl_seconds_after_finished can be set to be longer than 12 hours if someone needs it but for now I don't know of a use case since the cloud logs from the workload will last longer (depending on the rules of the logging bucket ~60 days).

pawloch00 (approved):

PR #322, opened 2025-01-13
Set min-master-version for A3-Ultra clusters

sharabiani:
## Fixes / Features
From GKE 1.31.4-gke.1072000, the DEFAULT version points to 550 (min required gpu driver version) for H200 GPU. In the lower versions the gpu driver will have an error and workloads cannot be scheduled.

## Testing / Documentation
Testing details.

- [ y/n ] Tests pass

### DIFF
--- a/src/xpk/core/blueprint/blueprint_generator.py
+++ b/src/xpk/core/blueprint/blueprint_generator.py
@@ -482,6 +482,7 @@ class BlueprintGenerator:
         use=[net_0_id],
         settings={
             "release_channel": "RAPID",
+            "min_master_version": "1.31.4-gke.1072000",
             "prefix_with_deployment_name": False,
             "name_suffix": cluster_name,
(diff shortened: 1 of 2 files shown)


src/xpk/core/blueprint/blueprint_generator.py
```diff
@@ -482,6 +482,7 @@ def generate_a3_ultra_blueprint(
         use=[net_0_id],
         settings={
             "release_channel": "RAPID",
+            "min_master_version": "1.31.4-gke.1072000",
```
pawloch00, line 485: Can we have it in variable?

pawloch00 (approved):

src/xpk/core/blueprint/blueprint_generator.py
sharabiani: We will remove this soon as the default will be set to this.

PR #346, opened 2025-01-27
Release v0.6.0

pawloch00:
## Fixes / Features
- Release v0.6.0 changes

## Testing / Documentation
Testing details.

- [ y/n ] Tests pass

### DIFF
--- a/.github/workflows/build_wheels.yaml
+++ b/.github/workflows/build_wheels.yaml
@@ -14,7 +14,6 @@
 
 name: PyPi releases
 
-
 on:
   workflow_dispatch:
     inputs:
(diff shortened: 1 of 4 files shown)


.github/workflows/build_wheels.yaml
```diff
@@ -15,7 +15,6 @@
 name: PyPi releases
 
 on: 
-  push:
```
44past4, line 18: Does it mean that we do not want for this being automatically on tag push?

44past4 (dismissed):

.github/workflows/build_wheels.yaml
pawloch00: Yes, I think this should be only manually triggered, as long as we don't have testpypi configured properly. Wdyt?

44past4: Makes sense for me.

pawloch00:
@sharabiani please approve

44past4 (dismissed):

PR #352, opened 2025-01-30
Implement config module

pawloch00:
## Fixes / Features
- Implement config module that supports setting predefined values in yaml file.

## Testing / Documentation
Testing details.

- [ y/n ] Tests pass

### DIFF
--- a/src/xpk/core/config.py
+++ b/src/xpk/core/config.py
@@ -43,14 +43,14 @@ class XpkConfig:
     self._config = config_file_path
     self._allowed_keys = default_keys
 
-  def _open_configs(self) -> dict:
+  def _open_configs(self) -> dict | None:
     dir_path = '/'.join(self._config.split('/')[:-1])
     file.ensure_directory_exists(dir_path)
 
     config_yaml = {'version': 'v1', CONFIGS_KEY: {}}
-
     if not os.path.exists(self._config):
-      return config_yaml
+      return None
+
     with open(self._config, encoding='utf-8', mode='r') as stream:
       config_yaml: dict = yaml.load(stream)
     return config_yaml
@@ -64,10 +64,9 @@ class XpkConfig:
       xpk_print(f'Key {key} is not an allowed xpk config key.')
       return
 
-    config_yaml = {'version': 'v1', CONFIGS_KEY: {}}
-    if os.path.exists(self._config):
-      config_yaml = self._open_configs()
-
+    config_yaml = self._open_configs()
+    if config_yaml is None:
+      config_yaml = {'version': 'v1', CONFIGS_KEY: {}}
     config_yaml[CONFIGS_KEY][key] = value
     self._save_configs(config_yaml)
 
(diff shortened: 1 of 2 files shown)


src/xpk/core/config.py
```diff
@@ -0,0 +1,89 @@
...
+  def _save_configs(self, config_yaml: dict) -> None:
+    with open(self._config, encoding='utf-8', mode='w') as stream:
+      yaml.dump(config_yaml, stream)
+
+  def set(self, key: str, value: str) -> None:
+    if key not in self._allowed_keys:
+      xpk_print(f'Key {key} is not an allowed xpk config key.')
+      return
+
+    dir_path = '/'.join(self._config.split('/')[:-1])
```
sharabiani, line 61: All the below lines can be moved to _open mehod:

```
    dir_path = '/'.join(self._config.split('/')[:-1])
    file.ensure_directory_exists(dir_path)

    config_yaml = {'version': 'v1', CONFIGS_KEY: {}}
    if os.path.exists(self._config):
```

src/xpk/core/config.py
```diff
@@ -0,0 +1,89 @@
...
+
+    config_yaml[CONFIGS_KEY][key] = value
+    self._save_configs(config_yaml)
+
+  def get(self, key: str) -> str | None:
+    if key not in self._allowed_keys:
+      xpk_print(f'Key {key} is not an allowed xpk config key.')
+      return None
+
+    if not os.path.exists(self._config):
```
sharabiani, line 76: no need for this. the _open method will handle it.

src/xpk/core/config.py
```diff
@@ -0,0 +1,89 @@
...
+      return None
+
+    config_yaml = self._open_configs()
+    vals: dict[str, str] = config_yaml[CONFIGS_KEY]
+    return vals[key]
+
+  def get_all(
+      self,
+  ) -> dict[str, dict[str, str] | str]:
+    if not os.path.exists(self._config):
```
sharabiani, line 86: no need for this. the _open method will handle it.

src/xpk/core/config.py
```diff
@@ -0,0 +1,89 @@
...
+      return {}
+    config_yaml = self._open_configs()
+    return config_yaml
```
sharabiani, line 89: `return config_yaml[CONFIGS_KEY]`

src/xpk/core/config.py
```diff
@@ -0,0 +1,90 @@
...
+    self._config = config_file_path
+    self._allowed_keys = default_keys
+
+  def _open_configs(self) -> dict | None:
+    dir_path = '/'.join(self._config.split('/')[:-1])
+    file.ensure_directory_exists(dir_path)
+
+    config_yaml = {'version': 'v1', CONFIGS_KEY: {}}
+    if not os.path.exists(self._config):
+      return None
```
sharabiani, line 52: shouldn't you return the default dict `config_yaml`?

sharabiani: this way you won't need `if config_yaml is None:` in other methods.

sharabiani (dismissed):

src/xpk/core/config.py
pawloch00: I can't open non existing file. And creating new files should only be done by write imo

sharabiani (approved):

PR #358, opened 2025-02-03
fix `xpk version` showing invalid git commit hash

### LINKED ISSUE #342: `xpk version` prints the wrong git commit
If I first install `xpk` such that it is available in the `$PATH`, and then run `xpk version` in my own git repo (such as MaxText), it ends up fetching the git commit of my git repo instead of the one where xpk was installed from.

gcie:
Add script in make install to save git commit hash in core.py when installing.

## Fixes / Features
Fixes #342

## Testing / Documentation
Run `xpk version` in different folders (outside of xpk repository) and check if git commit hash is always correctly shown

### DIFF
--- a/src/xpk/commands/version.py
+++ b/src/xpk/commands/version.py
@@ -15,25 +15,14 @@ limitations under the License.
 """
 
 from argparse import Namespace
-import os
 
-from ..core.commands import run_command_for_value
+from ..core.core import __git_commit_hash__
+from ..utils.console import xpk_print
 
 XPK_VERSION = 'v0.6.0'
-
-from ..utils.console import xpk_exit, xpk_print
 
 
 def version(args: Namespace) -> None:
   """Get version of xpk."""
   xpk_print('xpk_version:', XPK_VERSION)
-  if os.path.exists(os.path.join(os.getcwd(), '.git')):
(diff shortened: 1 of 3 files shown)


gcie:
> What if xpk is cloned inside another repo? Will it work?

I checked that and it will work.

In theory it is possible for `git rev-parse HEAD` to produce invalid hash, but it would require the user to have set the `$GIT_DIR` env variable to another repo (what would cause a lot of different issues when working with git). We can work around it by adding `--git-dir=.git`, but I think it's such a niche case that it's just not worth it.

src/xpk/commands/version.py
```diff
@@ -14,26 +14,13 @@
...
-    code, xpk_version = run_command_for_value(
-        'git rev-parse HEAD',
-        task='Get latest hash',
-        global_args=args,
-        quiet=True,
-    )
-    if code != 0:
-      xpk_exit(code)
-    xpk_print('git commit:', xpk_version.strip('\n'))
+  xpk_print('git commit hash:', __git_commit_hash__)
```
pawloch00, line 26: This git hash should be printed only if xpk is used from git clone. If xpk is installed from pip it should print only version.

pawloch00 (approved):

PR #373, opened 2025-02-17
Fixes for slurm-like commands

sharabiani:
## Fixes / Features
- Make updated to fix kjob and kueuectl installation
- Errors for commands fixed while testing

## Testing / Documentation
Testing details.

- [ y/n ] Tests pass


Makefile
```diff
@@ -45,6 +46,7 @@ mkdir-bin:
 install-kueuectl: mkdir-bin
 	curl -Lo $(BIN_PATH)/kubectl-kueue $(KUEUECTL_URL)
 	chmod +x $(BIN_PATH)/kubectl-kueue
+	sudo mv -f $(BIN_PATH)/kubectl-kueue $(USR_BIN_PATH)/kubectl-kueue
```
pawloch00, line 49: Are we not overwriting user installation of kueuectl by this mv?

Makefile
```diff
@@ -15,6 +15,7 @@ PROJECT_DIR := $(realpath $(shell dirname $(firstword $(MAKEFILE_LIST))))
 KJOB_DOCKER_IMG := xpk_kjob
 KJOB_DOCKER_CONTAINER := xpk_kjob_container
 BIN_PATH=$(PROJECT_DIR)/bin
+USR_BIN_PATH=/usr/local/bin
```
pawloch00, line 18: Is this true for macos?

src/xpk/core/kjob.py
```diff
@@ -112,6 +127,22 @@ def verify_kjob_installed(args: Namespace) -> int:
...
 
+def get_pod_template_interactive_command() -> str:
+  """Gets the interactive command for PodTemplate from config otherwise the default value.
+
+  Args:
+    args - user provided arguments
+  Returns:
+    str - PodTemplate's interactive command
+  """
+  config = XpkConfig()
```
pawloch00, line 138: Imho XpkConfig should be single object imported from config module, not initialized each time we want to get value.

Makefile
sharabiani: Aren't you addressing this in another PR as we discussed?

Makefile
sharabiani: yes

src/xpk/core/kjob.py
sharabiani: We can make a singleton class and replace it everywhere we used this module. It's out of this PR's scope.

Makefile
sharabiani: checks added

pawloch00 (approved):

PR #375, opened 2025-02-18
Add Workaround for Megascale Num Slices Issue on PW

SujeethJinesh:
## Fixes / Features
-
-

## Testing / Documentation
Testing details.

- [ y/n ] Tests pass


shauryagup (commented):
Not sure if we also need to override TPU_WORKER_HOSTNAMES env variable. I think it can be easily verified by running this on a multi host setup

shauryagup (approved):

SujeethJinesh:
> Not sure if we also need to override TPU_WORKER_HOSTNAMES env variable. I think it can be easily verified by running this on a multi host setup

Verified offline TPU_WORKER_HOSTNAMES is set correctly.

src/xpk/commands/workload.py
```diff
@@ -255,6 +255,28 @@
               volumeMounts:
               - mountPath: /tmp
                 name: shared-tmp
+              env:
+                # Workaround for v6e
```
RoshaniN, line 259: Hi Sujeeth,
Do you think we can inject this conditionally for v6e workloads on Pathways?
 i.e. check v6e type based on XPK's SystemCharacteristics and if that's true, add this section. Suggesting this especially since this is a workaround and specific to v6e.

src/xpk/commands/workload.py
```diff
@@ -255,6 +255,28 @@
...
+                - name: MEGASCALE_GRPC_ENABLE_XOR_TRACER
+                  value: "false"
```
RoshaniN, line 261: Hi Sujeeth,
Do you think we can inject this conditionally for v6e workloads on Pathways?
i.e. check v6e type based on XPK's SystemCharacteristics and if that's true, add this section. Suggesting this especially since this is a workaround and specific to v6e.

RoshaniN (commented):
Thanks Sujeeth

SujeethJinesh:
This PR depends on merging in https://github.com/AI-Hypercomputer/xpk/pull/379 first. After which we can switch to merging to `develop` rather than `main`.

44past4 (approved):

PR #403, opened 2025-03-03
Do not require manifest when attaching storage

gcie:
## Fixes / Features
- drop the requirement of providing manifest file when attaching FUSE or Filestore storage

## Testing / Documentation
- tests: `.github/workflows/fuse_tests.yaml` and `.github/workflows/filestore_tests.yaml`
- docs: `README.md`

### DIFF
--- a/.github/workflows/reusable_storage_tests.yaml
+++ b/.github/workflows/reusable_storage_tests.yaml
@@ -57,6 +57,7 @@ jobs:
       BUCKET_NAME: ${{secrets.BUCKET_NAME}}
       STORAGE_NAME: gcsfuse-test-${{inputs.run-id}}
   filestore-create-tests:
+    needs: [ fuse-attach-tests ]
     uses: ./.github/workflows/reusable_storage_tests_by_type.yaml
     concurrency: # We support one build or nightly test to run at a time currently.
       group: filestore-tests-${{inputs.run-id}}


sharabiani (commented):
A general comment. Will it have a conflict with Kjob Storage Support? If yes maybe better to make required changes after Kjob Storage PR is merged.

README.md
```diff
@@ -448,36 +448,35 @@ Currently XPK supports two types of storages: Cloud Storage FUSE and Google Clou
...
-    ```shell
-    python3 xpk.py storage attach test-storage --project=$PROJECT
-    --cluster=xpk-test --type=gcsfuse --auto-mount=false \
-    --mount-point='/test-mount-point' --readonly=false \
-    --manifest='examples/storage/gcsfuse-manifest.yaml'
-    ```
+- `--type` - the type of storage, currently xpk supports `gcsfuse` and `gcpfilestore` only.
+- `--auto-mount` - if set to true all workloads will have this storage mounted by default.
+- `--mount-point` - the path on which this storage should be mounted for a workload.
+- `--manifest` - the path to manifest which contains PersistentVolume and PersistentVolumeClaim definitions
```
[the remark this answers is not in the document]
gcie: Good question.

For the sake of backward compatibility I left this parameter as it was, so that when user provides the manifest, it works as before. It would simply ignore `--size` and `--bucket` params and use the manifest provided by the user.

This behavior is currently undocumented and if we want to leave it like so I think I should add some info in readme. But I also see the benefits of removing it to not overcomplicate things.

I'm for removal, but what do you think?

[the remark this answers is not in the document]
gcie: Also, `xpk storage attach` for `--type=gcpfilestore` still requires providing manifest, so I can't completely delete this parameter. I think this needs a proper rework, probably as a new task.

src/xpk/core/filestore.py
```diff
@@ -137,108 +154,112 @@ def create_filestore_instance(
...
+    ] = f"projects/{self.project}/global/networks/{network}"
     return data
 
-  def create_pv(self, vol: str, access_mode: str) -> dict:
-    abs_path = f"{os.path.dirname(__file__)}{FS_PV_PATH}"
-    with open(abs_path, "r", encoding="utf-8") as file:
+  def create_pv(self, name: str, vol: str, access_mode: str) -> dict:
+    """Create a yaml representing filestore PersistentVolume."""
+    if self.instance is None:
+      xpk_print(
```
pawloch00, line 194: What does it mean that filestore instance was not loaded?

src/xpk/core/filestore.py
```diff
@@ -137,108 +154,112 @@ def create_filestore_instance(
...
+          f"Filestore instance {self.name} does not exist or was not loaded."
+      )
+      xpk_exit(1)
+    template_path = os.path.dirname(__file__) + FS_PV_PATH
```
pawloch00, line 198: we should save those in xpkclusters directory I think

src/xpk/core/filestore.py
gcie: I want now the `FilestoreClient` to work in 2 cases: when there is no filestore instance and the user creates one, or there is a filestore instance, but it needs to be loaded.

Technically this error should never occur if the code is correct. I put it in just for testing purposes so that the error message would be more verbose.

src/xpk/core/filestore.py
gcie: We do. I moved the saving logic to `core.storage.create_storage_crds()`

src/xpk/core/filestore.py
pawloch00: I think we should stick to attach instead of load.

src/xpk/core/storage.py
```diff
@@ -487,28 +494,32 @@ def create_storage_crds(k8s_api_client: ApiClient, args: Namespace) -> None:
...
-  abs_path = f"{os.path.dirname(__file__)}{STORAGE_TEMPLATE_PATH}"
-  with open(abs_path, "r", encoding="utf-8") as file:
-    data = yaml.safe_load(file)
+  template_path = os.path.dirname(__file__) + STORAGE_TEMPLATE_PATH
+  with open(template_path, "r", encoding="utf-8") as file:
+    data = yaml.load(file)
+
+  ensure_directory_exists(MANIFESTS_PATH)
+  manifest_file_path = (
+      f"{MANIFESTS_PATH}/{args.cluster}-{args.zone}-{args.name}-manifest.yaml"
```
sharabiani, line 503: {project_id}-{region}-{cluster_name}-{args.name}-manifest.yaml
[example](https://github.com/AI-Hypercomputer/xpk/blob/05efeb55fca3bf174498e65ec6a522d07324771f/src/xpk/commands/cluster_gcluster.py#L164)

src/xpk/core/storage.py
```diff
@@ -474,7 +479,9 @@ def print_storages_for_cluster(storages: list[Storage]) -> None:
   )
 
 
-def create_storage_crds(k8s_api_client: ApiClient, args: Namespace) -> None:
+def create_storage_crds(
```
sharabiani, line 482: I think we can refactor this method a little bit and have something easier to read. For example (variable and functions names are just example):

```
def create_storage_crds():
  template = load_template()
  data = fill_template(template, args)
  store_manifest(data)
  apply_manifest(data)
```

src/xpk/core/filestore.py
```diff
@@ -137,108 +154,97 @@ def create_filestore_instance(
...
+    pvc = self.create_pvc(name, access_mode)
+    sc = self.create_sc(name, network)
+    return [pv, pvc, sc]
+
+  def load_location(self) -> str:
+    """Load and return filestore location"""
+    if self.location is not None:
+      return str(self.location)
+
+    if not self.check_filestore_instance_exists():
```
sharabiani, line 234: `check_filestore_instance_exists` now **'checks storage existance'** and **'sets some variables'**. I would suggest to split these responsibilities into different fuctions.

src/xpk/core/filestore.py
```diff
@@ -116,6 +131,8 @@ def create_filestore_instance(
   ) -> None:
     """Create new Filestore instance"""
 
+    self.location = self.zone if TIERS[tier].value == "Zonal" else self.region
```
sharabiani, line 134: I would use `Availability.Zonal.value` instead of hardcoded string.

src/xpk/core/filestore.py
```diff
@@ -137,108 +156,115 @@ def create_filestore_instance(
...
+    pvc = self.create_pvc(name, access_mode)
+    sc = self.create_sc(name, network)
+    return [pv, pvc, sc]
+
+  def load_location(self) -> str:
+    """Load and return Filestore's location"""
+    if self.location is not None:
+      return str(self.location)
+
+    parentZonal = self.get_parent(self.zone)
```
sharabiani, line 236: `load_location` function uses `get_parent` method and vice versa. Isn't there a loop here?

src/xpk/core/storage.py
```diff
@@ -487,39 +563,17 @@ def create_storage_crds(k8s_api_client: ApiClient, args: Namespace) -> None:
...
-  api_instance = k8s_client.CustomObjectsApi(k8s_api_client)
   try:
-    api_instance.create_cluster_custom_object(
-        group=XPK_API_GROUP_NAME,
-        version=XPK_API_GROUP_VERSION,
-        plural=STORAGE_CRD_PLURAL,
-        body=data,
-    )
-    xpk_print(f"Created {STORAGE_CRD_KIND} object: {data['metadata']['name']}")
+    template_path = os.path.dirname(__file__) + STORAGE_TEMPLATE_PATH
```
sharabiani, line 570: I see lines similar the below lines in multiple places. Maybe we can have a method to load data and pass just a filename/filepath?

```
    template_path = os.path.dirname(__file__) + STORAGE_TEMPLATE_PATH
    with open(template_path, "r", encoding="utf-8") as file:
      template = yaml.load(file)
```

pawloch00 (approved):

PR #406, opened 2025-03-05
Fix: file not found error when running outside xpk repository

gcie:
## Fixes / Features
Change the way blueprints are included in the library to follow [this guide](https://setuptools.pypa.io/en/stable/userguide/datafiles.html#subdirectory-for-data-files).
Currently running `xpk cluster create` command outside of the downloaded repo might end with file not found error when loading blueprints or yaml templates.


gcie:
It would have broken running xpk through `python3 xpk.py`. Closing for now.

PR #416, opened 2025-03-10
Update cluster-toolkit to 1.47.0

gcie:
## Fixes / Features
- update ctk to 1.47.0 (**Breaking change!**)

There was an network naming update in ctk for A3* clusters. This makes it so if this PR was merged, xpk would unable to run workloads on A3 Ultra and A3 Mega clusters created with ctk 1.45.1, but only on clusters created with the new version.

Also, this PR is required for A4 support.


src/xpk/core/blueprint/blueprint_generator.py
```diff
@@ -37,7 +37,7 @@
 }
 
 cluster_toolkit_url = "github.com/GoogleCloudPlatform/cluster-toolkit"
-cluster_toolkit_version = "v1.45.1"
+cluster_toolkit_version = "v1.47.0"
 
 
 def get_subnetworks_for_a3mega(cluster_name: str) -> list[str]:
```
sharabiani, line 43: the below codes can be deleted if not used anywhere else:

```def get_subnetworks_for_a3mega(cluster_name: str) -> list[str]:
  return [f"{cluster_name}-gpunet-{i}-subnet" for i in range(8)]
def get_subnetworks_for_a3ultra(cluster_name: str) -> list[str]:
  return [f"{cluster_name}-sub-1"] + [
      f"{cluster_name}-rdma-sub-{i}" for i in range(8)
  ]
  ```

File: src/xpk/core/blueprint/blueprint_generator.py [hunk 2]
```diff
@@ -157,6 +157,10 @@ def generate_a3_mega_blueprint(
                 "total_min_nodes": system_node_pool_min_node_count,
                 "total_max_nodes": 1000,
             },
+            "k8s_network_names": {
+                "gvnic_prefix": "vpc",
```
sharabiani, line 161: We can have `{cluster_name}-rdma-sub-` as rdma_prefix for a3ultra to keep it backward compatible, right?

sharabiani: I think with [this PR](https://github.com/GoogleCloudPlatform/cluster-toolkit/pull/3811) we can even have the A3Mega backward compatible and in result this upgrade won't break anything.

src/xpk/core/blueprint/blueprint_generator.py
```diff
@@ -490,6 +485,9 @@ def generate_a3_ultra_blueprint(
                 " alias_ip_range=[]}],"
                 f" {cluster_name}-rdma-net.subnetwork_interfaces_gke))"
             ),
+            "k8s_network_names": {
+                "rdma_prefix": f"{cluster_name}-rdma-sub-",
```
sharabiani, line 489: I would set `rdma_start_index: 0`, too; just in case CT may modify the default value in the future.

sharabiani (approved):

sharabiani (approved):

PR #418, opened 2025-03-13
Typing `args` (initial)

gcie:
## Fixes / Features
- initial solution to typing `args` parameter
- proof of concept on storage related commands


src/xpk/args/storage.py
```diff
@@ -0,0 +1,85 @@
...
+StorageAccessMode: TypeAlias = Literal[
+    'ReadWriteOnce', 'ReadOnlyMany', 'ReadWriteMany'
+]
+
+FilestoreTier: TypeAlias = Literal[
+    'BASIC_HDD', 'BASIC_SSD', 'ZONAL', 'REGIONAL', 'ENTERPRISE'
+]
+
+
+class StorageAttachArgs(ClusterConfig):
```
pawloch00, line 34: StorageAttachArgs and StorageCreateArgs share many fields. Can we move them to new, shared class?

src/xpk/args/utils.py
```diff
@@ -0,0 +1,35 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+
+import inspect
+from argparse import Namespace
+from typing import Any
+
+
+def apply_args(main_args: Namespace, annotation: Any) -> Any:
```
pawloch00, line 22: Bad naming imo. What does apply_args mean?

src/xpk/args/utils.py
```diff
@@ -0,0 +1,35 @@
...
+  args = annotation()
```
pawloch00, line 23: There is no information that annotation is callable. Please add docstring for this function

src/xpk/args/storage.py
```diff
@@ -0,0 +1,85 @@
...
+  bucket: Optional[str] = None
+  vol: Optional[str] = None
+  access_mode: StorageAccessMode = 'ReadWriteMany'
+  instance: Optional[str] = None
+  prefetch_metadata: bool = True
+  manifest: Optional[str] = None
+  mount_options: Optional[str] = 'implicit-dirs'
+
+
+class StorageCreateArgs(ClusterConfig):
```
pawloch00, line 52: Maybe it should be dataclass? If I want to test StorageCreate command in unit tests(in the future) and I want to pass StorageCreateArgs, is there any way to pass them via constructor?

src/xpk/args/gcloud_context.py
```diff
@@ -0,0 +1,128 @@
...
+    sys.exit(
+        'You must specify the project in the project flag or set it with'
+        " 'gcloud config set project <project>'"
+    )
+  return project_outputs[
+      -1
+  ]  # The project name lives on the last line of the output
+
+
+def get_zone():
```
pawloch00, line 48: This are general functions, not specific to args. Maybe they should be moved to utils?

src/xpk/args/storage.py
pawloch00: https://docs.python.org/3/library/dataclasses.html

PR #419, opened 2025-03-13
Add job priority support in Slurm mode

sharabiani:
## Fixes / Features
--priority added for xpk run and xpk batch command

## Testing / Documentation
Testing details.

- [ y/n ] Tests pass


src/xpk/core/kjob.py
```diff
@@ -78,6 +78,7 @@ class PodTemplateDefaults(Enum):
               workingDir: {working_directory}
               {resources}
           {node_selector}
+          priorityClassName: {priority}
```
44past4, line 81: It is not clear for me if priority should be set as priorityClassName on pod template or as kueue.x-k8s.io/priority-class label as indicated in Kueue documentation - https://kueue.sigs.k8s.io/docs/concepts/workload_priority_class/.

I would probably opt for the second option to avoid situations that Kubernetes Scheduler will be evicting some pods based on those priorities.

sharabiani: And we setup clusters with PriorityClass and not WorkloadPriorityClass:
https://github.com/AI-Hypercomputer/xpk/blob/c8f20956107c8a2b57064415d548555b11ec1413/src/xpk/blueprints/a3mega/kueue-xpk-configuration.yaml.tftpl#L47

44past4 (approved):

PR #420, opened 2025-03-14
`xpk storage detach`

gcie:
## Fixes / Features
- rename `xpk storage delete` to `xpk storage detach`
- create `xpk storage delete` command that deletes filestore instance


.github/workflows/reusable_storage_tests_by_type.yaml
```diff
@@ -185,12 +185,12 @@ jobs:
...
     - name: Delete the delete workload on the cluster
       run: python3 xpk.py workload delete --workload $FS_DELETE_WORKLOAD --cluster ${{inputs.cluster-name}} --zone=${{inputs.zone}}
-    - name: Delete storage
+    - name: Detach storage volumes
       if: always()
-      run: python3 xpk.py storage delete ${{secrets.STORAGE_NAME}} --zone=${{inputs.zone}} --cluster=${{inputs.cluster-name}}
+      run: python3 xpk.py storage detach ${{secrets.STORAGE_NAME}} --cluster=${{inputs.cluster-name}} --zone=${{inputs.zone}}
+    - name: Delete GCP Filestore Storage instance
+      if: always() && inputs.storage-command == 'create' && inputs.storage-type == 'gcpfilestore'
+      run: python3 xpk.py storage delete ${{secrets.STORAGE_NAME}} --cluster=${{inputs.cluster-name}} --zone=${{inputs.zone}}
```
sharabiani, line 193: Can we verify if the storage is deleted. It will be something like the step above `    - name: List and verify existing Storages`

src/xpk/commands/storage.py
```diff
@@ -85,6 +88,35 @@ def storage_create(args: Namespace) -> None:
...
+  filestore_instance_name = filestore_client.get_instance_fullname()
+
+  children = [
+      storage.name
+      for storage in storages
+      if storage.bucket.startswith(filestore_instance_name)
+  ]
+
+  if children and not args.force:
+    xpk_print(
```
sharabiani, line 110: I think we can ask for user input instead of stopping the operation: Something like [this](https://github.com/AI-Hypercomputer/xpk/blob/c8f20956107c8a2b57064415d548555b11ec1413/src/xpk/commands/workload.py#L620)

sharabiani: - And if they confirm the operation, we need to detach them first then delete.
- Are we going to only detach gcsfuse and not deleting them? @44past4 WDYT?

sharabiani: Should we detach an storage if it is deleted outside XPK and we only have the pv,pvc in the cluster?

gcie: > Are we going to only detach gcsfuse and not deleting them?

We do not manage automatic FUSE bucket creation (yet), so no deletion either.

gcie: > Should we detach an storage if it is deleted outside XPK and we only have the pv,pvc in the cluster?

If I understand this correctly, this would require us to check if storage exists before running every workload. I don't think we should do this.

If the user created and attached the storage with xpk and afterwards deleted it himself, he will (rightfully) get an error when running workloads. Detaching it with xpk will fix the error.

sharabiani: > If I understand this correctly, this would require us to check if storage exists before running every workload. I don't think we should do this.

Not with every workload. I meant when deleting only.
The edge case I was thinking about (maybe not so likely) is if a storage is attached to more than one cluster and it's possible they run `storage delete` for each cluster. First will delete and the other fails and requires `xpk detach`. maybe it's a good UX to detach a not existing storage if they call `storage delete`
Ignore it if it takes time to handle it.

gcie: I just added user prompt and automatic detaching. It should handle your edge case correctly, because detecting attached volumes does not take cluster name into the account, so the first `storage delete` will detach all volumes from all clusters (if the user confirms this).

Technically `xpk storage detach` does not need `--cluster` param, but changing this is tricky due to widely used `setup_k8s_env` that performs get-credentials (unnecessary in this case).

sharabiani: I don't quite understand how come detach doesn't get cluster name. it creates `k8s_api_client` which make sense only with a cluster name.
I can see `  add_cluster_arguments(req_args, required=True)` in the detach command parser.

src/xpk/commands/storage.py
```diff
@@ -85,6 +90,39 @@ def storage_create(args: Namespace) -> None:
...
+  children = [
+      storage
+      for storage in storages
+      if storage.bucket.startswith(filestore_instance_name)
+  ]
+
+  if children and not args.force:
+    children_names = ", ".join(child.name for child in children)
+    detach = get_user_input(
+        "Before deleting filestore instance it's recommended to detach all"
```
sharabiani, line 114:
```
        "Deleting a filestore storage will destroy your filestore instance and all its data will be lost."
        "Do you wish to delete the filestore instance {filestore instance name or full name}: y (yes) / n"
        " (no):\n'"
```

src/xpk/commands/storage.py
sharabiani: However,  I agree that detaching all the volumes from the cluster when deleting solves [my comment](https://github.com/AI-Hypercomputer/xpk/pull/420#discussion_r1998286886).

src/xpk/commands/storage.py
gcie: Do you want to display this prompt always, or only when there are PV/PVC/SC attached?

sharabiani: I think it would be a better UX to check if the storage is attached and exists before asking for confirmation.

gcie: User can attach a storage, then put data on it, then detach it. We have no way of checking if there is some data leftover in this case. We can only check if there are some storages attached.

I think the current prompt better explains what is happening.

.github/workflows/reusable_storage_tests_by_type.yaml
```diff
@@ -185,12 +185,16 @@ jobs:
...
       run: |
         ! kubectl get volumebundle | grep ${{secrets.STORAGE_NAME}}
-    - name: Delete the filestore instance
-      if: always() && inputs.storage-command == 'create'
-      run: gcloud filestore instances delete ${{secrets.STORAGE_NAME}} --zone=${{inputs.zone}} --force
+    - name: Delete GCP Filestore Storage instance
+      if: always() && inputs.storage-command == 'create' && inputs.storage-type == 'gcpfilestore'
+      run: python3 xpk.py storage delete ${{secrets.STORAGE_NAME}} --cluster=${{inputs.cluster-name}} --zone=${{inputs.zone}}
+    - name: Verify deletion of GCP Filestore Storage instance
+      if: inputs.storage-command == 'create' && inputs.storage-type == 'gcpfilestore'
```
sharabiani, line 198: `inputs.storage-command == 'create'`?

src/xpk/commands/storage.py
```diff
@@ -85,6 +90,37 @@ def storage_create(args: Namespace) -> None:
...
+      storage
+      for storage in storages
+      if storage.bucket.startswith(filestore_instance_name)
+  ]
+
+  if children and not args.force:
+    detach = get_user_input(
+        "Deleting a filestore storage will destroy your filestore instance and"
+        " all its data in all volumes will be lost. Do you wish to delete the"
+        f" filestore instance {filestore_instance_name}: y (yes) / n (no):\n'"
```
sharabiani, line 115: maybe a question mark instead of ":"

src/xpk/commands/storage.py
```diff
@@ -85,6 +90,37 @@ def storage_create(args: Namespace) -> None:
...
+    )
+    if detach:
```
sharabiani, line 117: if detach is false we need to stop the operation fully including ' filestore_client.delete_filestore_instance()'.

gcie: This way we will force the user to remove all attached storages when deleting filestore instance. He may want to only remove the instance, for example in order to clear data and recreate it later.

I know it's a rare case, but I need confirmation if we really want this.

.github/workflows/reusable_storage_tests_by_type.yaml
gcie: If we `create` then we `delete`, but if we `attach` then we `detach`.

src/xpk/commands/storage.py
sharabiani: I think we better look at the problem like this:
xpk storage create -> (create instance AND attach)
xpk storage delete -> (detach AND delete instance)

src/xpk/commands/storage.py
```diff
@@ -85,6 +90,39 @@ def storage_create(args: Namespace) -> None:
...
+      if storage.bucket.startswith(filestore_instance_name)
+  ]
+
+  if children and not args.force:
+    detach = get_user_input(
+        "Deleting a filestore storage will destroy your filestore instance and"
+        " all its data in all volumes will be lost. Do you wish to delete the"
+        f" filestore instance {filestore_instance_name}?\n y (yes) / n (no):\n'"
+    )
+    if not detach:
```
sharabiani, line 117: nit: I would add a message that "Deleting storage canceled."

sharabiani (approved):

PR #422, opened 2025-03-18
Update v6e-8 to use the correct system characteristics

Obliviour:
Based on https://cloud.google.com/tpu/docs/v6e, the type is ct6e-standard-8t and 1 vm per slice. 

## Fixes / Features
- Fixes v6e-8 type based on documentation
-

## Testing / Documentation
Testing details.

- [ y ] Tests pass


Obliviour:
not needed at the moment, v6e-8 works in the current way and this way. We can leave as is and change / add this config if someone nees it.

PR #423, opened 2025-03-18
Fix xpk version error when cloned from git

sharabiani:
## Fixes / Features
-
-

## Testing / Documentation
Testing details.

- [ y/n ] Tests pass

### DIFF
--- a/src/xpk/commands/version.py
+++ b/src/xpk/commands/version.py
@@ -15,7 +15,10 @@ limitations under the License.
 """
 
 from ..core.config import __version__
-from ..utils.console import xpk_print
+from ..utils.console import xpk_print, xpk_exit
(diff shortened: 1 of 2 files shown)


src/xpk/commands/version.py
```diff
@@ -24,6 +27,20 @@ def get_xpk_version() -> str:
...
+  xpk_version = __version__
+  git_hash = ''
+
+  if os.path.exists(os.path.join(os.getcwd(), '.git')):
+    code, git_hash = run_command_for_value(
+        'git rev-parse HEAD',
+        task='Get latest hash',
+        global_args=args,
+        quiet=True,
+    )
```
gcie, line 39: This would cause the problem from #342 that was fixed.

PR #426, opened 2025-03-19
fix yaml alignment when `remote-python-sidecar-image` is passed

sadikneipp:
## Fixes / Features
- Init Container section of sidecar was misaligned by two spaces, which would cause `kubectl apply` to fail when using `--remote-python-sidecar-image` in `workload create-pathways`

## Testing / Documentation
Tested that this can be deployed successfully in a v4-8 cluster.


RoshaniN (approved):
Thanks Sadi

PR #429, opened 2025-03-24
Bump go version to 1.24.1

mbobrovskyi:
## Fixes / Features
- Bump go version to 1.24.1

## Testing / Documentation
Testing details.

- [ y/n ] Tests pass

### DIFF
--- a/tools/Dockerfile-kjob
+++ b/tools/Dockerfile-kjob
@@ -12,39 +12,22 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-
-# Use a google-cloud-cli image as the base
-ARG BASE_IMAGE=gcr.io/google.com/cloudsdktool/google-cloud-cli:stable
-FROM ${BASE_IMAGE}
+FROM golang:1.24.1-alpine
 
 # Install necessary tools and libraries
-RUN apt-get update && \
-    apt-get install -y --no-install-recommends \
-        apt-transport-https \
-        ca-certificates \
-        curl \
-        gnupg \
-        lsb-release \
-        git \
-        make \
-        unzip \
-        wget
+RUN apk add git make bash
 
-# Install Go
-ARG GO_VERSION=1.23.0
-RUN wget -q "https://go.dev/dl/go${GO_VERSION}.linux-amd64.tar.gz" -O go.tar.gz && \
-    tar -C /usr/local -xzf go.tar.gz && \
-    rm go.tar.gz
-
-# Set GOPATH and add to PATH
-ENV GOPATH /go
-ENV PATH $PATH:/usr/local/go/bin:$GOPATH/bin
+ARG KJOB_BRANCH=main
+ARG KJOB_COMMIT_HASH=e2c92af44c047016cd5789995f6276b79f22663d
 
 # Clone the kjob repository
-ARG KJOB_BRANCH=main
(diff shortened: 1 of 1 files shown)


tools/Dockerfile-kjob
```diff
@@ -31,7 +31,7 @@ RUN apt-get update && \
         wget
 
 # Install Go
-ARG GO_VERSION=1.23.0
+ARG GO_VERSION=1.24.1
```
sharabiani, line 34: Doesn't it compile without this upgrade?
If it doesn't, it means xpk's current and older versions are broken because we clone the main branch of kjob and have `GO_VERSION=1.23.0` to compile it.

mbobrovskyi: > Doesn't it compile without this upgrade?

Yeah, we updated the Go version for kjob in the main branch.

https://github.com/kubernetes-sigs/kjob/blob/main/go.mod#L3

> If it doesn't, it means xpk's current and older versions are broken because we clone the main branch of kjob and have GO_VERSION=1.23.0 to compile it.

Oh, this isn't good. I think we should use the kjob release version for xpk releases.

mbobrovskyi: But I don't think it should break. We're using it only for tests.

sharabiani: > And the user installs it manually while following README.md
> 
> https://github.com/AI-Hypercomputer/xpk/blob/main/README.md?plain=1#L97

Not necessarily. We install it with [`make`](https://github.com/AI-Hypercomputer/xpk/blob/33369581705e7d27f367fd2fc7fd9af01669a798/Makefile#L51), too.

> Oh, this isn't good. I think we should use the kjob release version for xpk releases.

Can you please run `make` in your branch and make sure kjob installs with no issues?

mbobrovskyi: > Not necessarily. We install it with [make](https://github.com/AI-Hypercomputer/xpk/blob/33369581705e7d27f367fd2fc7fd9af01669a798/Makefile#L51), too.

Hmm, I thought we are using it only for tests. In that case, this solution won't work on macOS because Go in docker will build it only for Linux.

> Can you please run make in your branch and make sure kjob installs with no issues?

Yeah, you right, it's not working. I'm wondering why the tests didn't catch this error.

sharabiani: With new version yes. But the issue is the current version of xpk main branch doesn't work. In other words, with kjob's changes now current xpk's users' `make` is failing.

@44past4 Can you please advice?

mbobrovskyi: I propose releasing kjob v0.2.0 first and then creating a patch release for xpk v0.7.1 to avoid these problems in the future. WDYT?

44past4: We should lock kjob version in xpk by either using a released version or using a specific commit.

sharabiani: @44past4 Thank you. My main question is about how to handle the current users that already will face error if they run `make`.

@mbobrovskyi can you please pin xpk to kjob's current commit? We can update it to a release tag/branch when it's ready.

mbobrovskyi: Done

mbobrovskyi:
@sharabiani should I create cherry-pick to main?

sharabiani:
> Farhad Sharabiani should I create cherry-pick PR to main?

Can you please create cherry-pick PR to `staging` branch?

PR #435, opened 2025-03-27
Remove sudo requirement from make

sharabiani:
## Fixes / Features
- To avoid requirement for `export PATH=$PATH:$PWD/bin` we used to copy kjob and kueuectl binaries to user's bin directory. It needed `sudo`. Now sudo is removed because it made problems for the users scripts.
- Now it's required to run `export PATH=$PATH:$PWD/bin` 

## Testing / Documentation
Testing details.

- [ y/n ] Tests pass

### DIFF
--- a/.github/workflows/build_tests.yaml
+++ b/.github/workflows/build_tests.yaml
@@ -90,16 +90,7 @@ jobs:
     - uses: actions/setup-python@v5
       with:
         python-version: ${{ matrix.python-version }}
-    - name: Check if cache exists
-      id: check-cache
-      uses: actions/cache@v3
-      with:
-        path: |
-          usr/local/bin/
-          ~/.cache/pip
-          ${{env.pythonLocation}}
-        key: xpk-deps-${{ matrix.python-version }}-${{needs.set-variables.outputs.run-id}}
-        lookup-only: true
+
     - name: install dependencies
       if : steps.check-cache.outputs.cache-hit != 'true'
       run: make install-lint && make install-dev && cp ./bin/kubectl-kueue /usr/local/bin/kubectl-kueue && cp ./bin/kubectl-kjob /usr/local/bin/kubectl-kjob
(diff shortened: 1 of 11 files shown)


Makefile
```diff
@@ -46,7 +46,6 @@ mkdir-bin:
 install-kueuectl: mkdir-bin
 	curl -Lo $(BIN_PATH)/kubectl-kueue $(KUEUECTL_URL);
 	chmod +x $(BIN_PATH)/kubectl-kueue;
-	sudo mv -f $(BIN_PATH)/kubectl-kueue $(USR_BIN_PATH)/kubectl-kueue;
```
RoshaniN, line 49: Thanks Farhad!

Should "export PATH=$PATH:$PWD/bin" be added back into the README then?

RoshaniN: Synced offline, the readme instruction is still there.

RoshaniN (approved):

Makefile
sharabiani: It's already there.

.github/workflows/build_tests.yaml
```diff
@@ -98,11 +98,11 @@ jobs:
           usr/local/bin/
           ~/.cache/pip
           ${{env.pythonLocation}}
-        key: xpk-deps-${{ matrix.python-version }}-${{needs.set-variables.outputs.run-id}}
+        key: xpk-deps-${{ matrix.python-version }}-${{github.run_id	}}-${{github.run_attempt}}
```
gcie, line 101: Doesn't this make caching completely obsolete?

sharabiani: I don't quite get how it obsoletes the caching.
Previously the old cached files were used even when re-running the workflow. Now per each workflow run we will have a new cached files. They will be used in the other steps of the tests and avoid reinstalling the dependencies per each step.

gcie: Nevermind, I was thinking of caching across different runs. I agree, it's better this way 👍 
LGTM, but please remove the tabulation

sharabiani: done

PR #436, opened 2025-03-27
Implement Parallelstore and Hyperdisk storages attach 

sharabiani:
## Fixes / Features
- Attaching Parallelstore instance implemented.
- Attaching Block storage (PersistentDisk, Hyperdisk) instances implemented
- They are currently supported only by providing a PV/PVC manifest file.

## Testing / Documentation
Testing details.

- [ y/n ] Tests pass

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,10 +56,11 @@ and the following GPU types:
 and the following CPU types:
 * n2-standard-32
 
-xpk also supports Google Cloud Storage solutions:
+xpk also supports [Google Cloud Storage solutions](#storage):
 * [Cloud Storage FUSE](#fuse)
 * [Filestore](#filestore)
 * [Parallelstore](#parallelstore)
+* [Block storage (Persistent Disk, Hyperdisk)](#block-storage-persistent-disk-hyperdisk)
 
 # Permissions needed on Cloud Console:
 
@@ -444,10 +445,11 @@ Currently, the below flags/arguments are supported for A3-Mega and A3-Ultra mach
 
 
 ## Storage
-Currently XPK supports three types of storages:
-- Cloud Storage FUSE
-- Google Cloud Filestore
-- Google Cloud Parallelstore
+Currently XPK supports the below types of storages:
+- [Cloud Storage FUSE](#fuse)
+- [Google Cloud Filestore](#filestore)
+- [Google Cloud Parallelstore](#parallelstore)
+- [Google Cloud Block storages (Persistent Disk, Hyperdisk)](#block-storage-persistent-disk-hyperdisk)
 
 ### FUSE
 A FUSE adapter lets you mount and access Cloud Storage buckets as local file systems, so applications can read and write objects in your bucket using standard file system semantics.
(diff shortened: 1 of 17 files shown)


README.md
```diff
@@ -505,6 +511,54 @@ Commands `xpk storage create` and `xpk storage attach` with `--type=gcpfilestore
...
+- `--readonly` - if set to true, workload can only read from storage.
+- `--manifest` - path to the manifest file containing PersistentVolume and PresistentVolumeClaim definitions.
+
+### Block storage (Persistent Disk, Hyperdisk)
+
+A PersistentDisk adapter lets you mount and access Google Cloud Block storage solutions ([Persistent Disk](https://cloud.google.com/kubernetes-engine/docs/concepts/storage-overview#pd), [Hyperdisk](https://cloud.google.com/kubernetes-engine/docs/concepts/storage-overview#hyperdisk)) as local file systems, so applications can read and write files in your volumes using standard file system semantics.
+
+To use the GCE PersistentDisk with XPK you need to create a [disk in GCE](https://cloud.google.com/compute/docs/disks). Please consider that the disk type you are creating is [compatible with the VMs](https://cloud.google.com/compute/docs/machine-resource#machine_type_comparison) in the default and accelerator nodepools.
+
+Once it's ready you can use `xpk storage attach` with `--type=pd` command to attach a PersistentDisk instance to your cluster. Currently, attaching a PersistentDisk is supported only by providing a manifest file.
```
gcie, line 544: Does `--type=pd` originate from Persistent Disk? If yes, then it does not fit Hyperdisk. I suggest using one of `--type=disk` or `--type=block`

src/xpk/core/kjob.py
```diff
@@ -449,9 +449,16 @@ def create_volume_bundle_instance(
...
+
   gcsfuse_storages = get_auto_mount_gcsfuse_storages(k8s_api_client)
   if len(gcsfuse_storages) > 0:
-    return "gke-gcsfuse/volumes=true"
-  return None
+    annotations.append("gke-gcsfuse/volumes=true")
+
+  parallelstore_storages = get_auto_mount_parallelstore_storages(k8s_api_client)
+  if len(parallelstore_storages) > 0:
+    annotations.append("gke-parallelstore/volumes=true")
```
gcie, line 462: We need this annotation in `workload create` too.

I worked on supporting multiple annotations there (#449), I think we need to sync

README.md
sharabiani: it originates from the CSI Driver same as the other types . The driver name is`GcePersistentDiskCsiDriver` in gcloud or `pd.csi.storage.gke.io` in PV.

gcie: Still, some users might wonder why they have to set `--type=pd` when they use Hyperdisk instead of Persistent Disk (and might think it's a typo). The point of xpk is to simplify user experience as much as possible.

Although I'm fine with `pd` if you insist on it.

src/xpk/commands/workload.py
```diff
@@ -391,6 +393,12 @@ def workload_create(args) -> None:
     gcpfilestore_storages: list[Storage] = list(
         filter(lambda storage: storage.type == GCP_FILESTORE_TYPE, storages)
     )
+    parallelstore_storages: list[Storage] = list(
+        filter(lambda storage: storage.type == PARALLELSTORE_TYPE, storages)
+    )
+    pd_storages: list[Storage] = list(
+        filter(lambda storage: storage.type == GCE_PD_TYPE, storages)
+    )
```
gcie, line 401: Wouldn't it be clearer to use list comprehension here instead of filter+lambda?

Google's Python styleguide suggests so: go/pystyle#lambdas

src/xpk/commands/workload.py
```diff
@@ -391,6 +393,12 @@ def workload_create(args) -> None:
     gcpfilestore_storages: list[Storage] = list(
         filter(lambda storage: storage.type == GCP_FILESTORE_TYPE, storages)
     )
+    parallelstore_storages: list[Storage] = list(
```
gcie, line 396: The `gcs_fuse_storages`, `gcpfilestore_storages`, `parallelstore_storages` and `pd_storages` are not used outside of the scope of this giant `if` expression. We should consider refactoring it from here as it is getting quite large.

src/xpk/commands/workload.py
```diff
@@ -453,7 +482,13 @@ def workload_create(args) -> None:
...
         yml_string = rdma_decorator.decorate_jobset(yml_string, sub_networks)
 
-      if len(gcs_fuse_storages) + len(gcpfilestore_storages) > 0:
+      if (
+          len(gcs_fuse_storages)
+          + len(gcpfilestore_storages)
+          + len(parallelstore_storages)
+          + len(pd_storages)
+          > 0
+      ):
```
gcie, line 491: Why not `if all_storages:` instead?

src/xpk/core/workload_decorators/storage_decorator.py
```diff
@@ -45,6 +45,9 @@ def add_annotations(job_manifest, storages):
   gcs_present = any(storage.type == GCS_FUSE_TYPE for storage in storages)
   if gcs_present:
     annotations.update(GCS_FUSE_ANNOTATIONS)
+  gcs_present = any(storage.type == PARALLELSTORE_TYPE for storage in storages)
+  if gcs_present:
+    annotations.update(PARALLELSTORE_ANNOTATIONS)
```
gcie, line 50: `parallelstore_present`

src/xpk/commands/workload.py
sharabiani: That's my opinion too, but apparantly it was needed to be addressed separately: https://github.com/AI-Hypercomputer/xpk/pull/448#discussion_r2035182791

src/xpk/commands/workload.py
gcie: But you wrote this line and there is no need to add unnecessary work to the refactoring. And we should not repeat bad code for the sole sake of consistency.

sharabiani: I haven't written this line! I just added the two more storage option to the line.
And there is no repeat for the sake of consistency. Changing this line needs to refactoring the whole routine which is out of the scope of this task. Please create a task for the issue you see in this comment and the other 2-3 comments that are related to refactoring the storage task (it can include my comment [here](https://github.com/AI-Hypercomputer/xpk/pull/448#discussion_r2035182791) as well.

src/xpk/commands/workload.py
gcie: I was thinking of refactoring this storage-related logic, not pathways'. The only thing this code does is setting `all_storages` and `service_account`. I suggest moving it to a function that returns both.

pawloch00:
Can we have some tests added for this types of storage?

src/xpk/commands/workload.py
sharabiani: Task created for that. Feel free to add more description: b/409974713

src/xpk/commands/workload.py
sharabiani: b/409974713

sharabiani:
> Can we have some tests added for this types of storage?

Can't make it now. Task created: b/409975063

gcie (approved):

PR #446, opened 2025-04-04
Add the option to use Multi-tier checkpointing in workloads

abhinavclemson:
## Fixes / Features
Add the option to use Multi-tier checkpointing in workloads

## Testing details.

- [x] Manually tested on Cluster where MTC are deployed.

- [x] Appropriate changes to documentation are included in the PR


src/xpk/parser/workload.py
```diff
@@ -618,6 +618,15 @@ def add_shared_workload_create_optional_arguments(args_parsers):
             ' be used with the CSI driver provided by GKE.'
         ),
     )
```
RoshaniN, line 620: Thanks Abhinav!
Do you want to include this argument for Pathways (workload create-pathways) also? Adding it to shared args will do that. (Same about the ramdisk-directory above.)

Please consider moving them to workload create if this feature isn't ready / tested with Pathways!

RoshaniN (changes requested):
Thanks!

src/xpk/parser/workload.py
abhinavclemson: Currently, we are not testing/supporting MTC on pathways. Changes only comes into play if a ramdisk directory is provided. Above changes are intrinsic to McJax workload creation with ramdisk directory.

RoshaniN: Then move this arg along with other MTC args (such as ramdisk-directory) outside of this function to https://github.com/AI-Hypercomputer/xpk/blob/main/src/xpk/parser/workload.py#L113 where McJAX only args are present.

This function adds arguments to both `workload create` and `workload create-pathways` flows.
```
def add_shared_workload_create_optional_arguments(args_parsers):
  """Add shared optional arguments in workload create and Pathways workload create.
  ```

PR #448, opened 2025-04-04
Integrate PathwaysJob into XPK.

RoshaniN:
## Fixes / Features
- Using PathwaysJob CRD https://github.com/google/pathways-job/ instead of JobSet CRD for PathwaysJob workloads.
- - Installed PathwaysJob CRD as part of cluster create.
- Removing storage, failure policy rules and other unsupported , untested features from the `xpk workload create-pathways` flow.
- Adding custom components to Pathways workload flows conditionally.
- Modifying workload delete to also cater to PathwaysJob workloads.
- Adding support for elasticity in Pathways workloads in XPK, by introducing new args.
- Bug fix to correct default remote python image and other string default types from None to ""
- Removed cpu-rm, cpu-proxy. Retained cpu-user and changed its machine type to n2-standard-64 to allow Pathways workloads with huge pods named Pathways head.
- Updated `xpk cluster delete` flow to support deletion of PathwaysJobs as well.
- Ensuring backward compatibility and prompting users to upgrade to use PathwaysJob for pathways workloads.

## Testing / Documentation
Testing details.

- [ y ] Tests pass

### DIFF
--- a/.github/workflows/reusable_workload_tests.yaml
+++ b/.github/workflows/reusable_workload_tests.yaml
@@ -84,7 +84,7 @@ jobs:
     - name: List out the workloads on the cluster
       run: python3 xpk.py workload list --cluster ${{inputs.cluster-name}} --zone=${{inputs.zone}}
     - name: Run xpk info
-      run: python3 xpk.py info --cluster ${{inputs.cluster-name}} --zone=${{inputs.zone}} | tee output.txt | grep -P "^(?=.*QUEUE)(?=.*PENDING_WORKLOADS)(?=.*ADMITTED_WORKLOADS)(?=.*1x${{inputs.tpu-type}}:google.com/tpu)(?=.*cpu-rm:cpu)(?=.*cpu-rm:memory)(?=.*cpu-proxy:cpu)(?=.*cpu-proxy:memory)(?=.*cpu-user:cpu)(?=.*cpu-user:memory)" || (echo 'Invalid command output' && cat output.txt && exit 1)
+      run: python3 xpk.py info --cluster ${{inputs.cluster-name}} --zone=${{inputs.zone}} | tee output.txt | grep -P "^(?=.*QUEUE)(?=.*PENDING_WORKLOADS)(?=.*ADMITTED_WORKLOADS)(?=.*1x${{inputs.tpu-type}}:google.com/tpu)(?=.*cpu-user:cpu)(?=.*cpu-user:memory)" || (echo 'Invalid command output' && cat output.txt && exit 1)
     - name: Delete the workload on the cluster
       run: python3 xpk.py workload delete --workload $WORKLOAD_NAME --cluster ${{inputs.cluster-name}} --zone=${{inputs.zone}}
     - name: Delete the Pathways workload on the cluster
--- a/README.md
+++ b/README.md
@@ -221,6 +221,7 @@ all zones.
     --num-slices=4 --on-demand \
     --tpu-type=v5litepod-16
     ```
+    Note that Pathways clusters need a CPU nodepool of n2-standard-64 or higher.
 
 *   Cluster Create for Ray:
     A cluster with KubeRay enabled and a RayCluster can be created using `cluster create-ray`.
--- a/src/xpk/commands/cluster.py
+++ b/src/xpk/commands/cluster.py
@@ -22,6 +22,7 @@ from ..core.cluster import (
     get_cluster_credentials,
     install_nccl_on_cluster,
     set_jobset_on_cluster,
+    set_pathways_job_on_cluster,
     setup_k8s_env,
     update_cluster_with_gcsfuse_driver_if_necessary,
     update_cluster_with_workload_identity_if_necessary,
@@ -205,6 +206,10 @@ def cluster_create(args) -> None:
   set_jobset_on_cluster_code = set_jobset_on_cluster(args)
   if set_jobset_on_cluster_code != 0:
     xpk_exit(set_jobset_on_cluster_code)
+
+  set_pathways_job_on_cluster_code = set_pathways_job_on_cluster(args)
+  if set_pathways_job_on_cluster_code != 0:
+    xpk_exit(set_pathways_job_on_cluster_code)
 
   xpk_print('Enabling Kueue on the cluster')
   install_kueue_on_cluster_code = install_kueue_on_cluster(args)
@@ -792,10 +797,8 @@ def run_gke_cluster_create_command(
   addons = []
   if args.enable_gcsfuse_csi_driver:
     addons.append('GcsFuseCsiDriver')
-
   if args.enable_gcpfilestore_csi_driver:
     addons.append('GcpFilestoreCsiDriver')
-
   if len(addons) > 0:
     addons_str = ','.join(addons)
     command += f' --addons={addons_str}'
--- a/src/xpk/commands/workload.py
+++ b/src/xpk/commands/workload.py
@@ -36,19 +36,23 @@ from ..core.nap import (
     is_autoprovisioning_enabled,
 )
 from ..core.pathways import (
+    check_if_pathways_job_is_installed,
     ensure_pathways_workload_prerequisites,
-    get_pathways_proxy_args,
-    get_pathways_rm_args,
-    get_pathways_sidecar_container,
     get_pathways_unified_query_link,
-    get_pathways_worker_args,
+    append_custom_pathways_proxy_server,
+    append_custom_pathways_server,
+    append_custom_pathways_worker,
+    append_custom_colocated_python_sidecar,
     get_user_workload_for_pathways,
+    try_to_delete_pathwaysjob_first,
 )
 from ..core.resources import CLUSTER_METADATA_CONFIGMAP, get_cluster_configmap
 from ..core.scheduling import (
     check_if_workload_can_schedule,
     create_accelerator_label,
     create_machine_label,
+    create_tpu_topology,
+    create_tpu_machine_type,
     get_cpu_affinity,
     get_gpu_scheduler,
 )
@@ -57,8 +61,6 @@ from ..core.storage import (
     GCP_FILESTORE_TYPE,
     Storage,
     add_bucket_iam_members,
-    get_storage_volume_mounts_yaml,
-    get_storage_volumes_yaml,
     get_storages_to_mount,
     get_storage_volume_mounts_yaml_for_gpu,
     get_storage_volumes_yaml_for_gpu,
@@ -66,7 +68,6 @@ from ..core.storage import (
 )
 from ..core.system_characteristics import (
     AcceleratorType,
-    AcceleratorTypeToAcceleratorCharacteristics,
     get_system_characteristics,
 )
 from ..core.vertex import create_vertex_experiment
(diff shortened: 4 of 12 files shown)


.github/workflows/reusable_cluster_create.yaml
```diff
@@ -77,5 +77,5 @@ jobs:
     - name: Check xpk installation
       run: xpk --help
     - name: Create a Pathways-enabled XPK Cluster with 2x ${{inputs.tpu-type}} nodepools. Larger num-nodes to avoid master resizing.
-      run: python xpk.py cluster create-pathways --cluster ${{inputs.cluster-name}} --tpu-type=${{inputs.tpu-type}}  --num-slices=1 --zone=${{inputs.zone}} --default-pool-cpu-machine-type=n1-standard-16 --default-pool-cpu-num-nodes=2 --reservation='${{ secrets.GCP_TPU_V4_RESERVATION }}' --enable-gcpfilestore-csi-driver --enable-gcsfuse-csi-driver --custom-cluster-arguments="${CLUSTER_ARGUMENTS}"
```
SujeethJinesh, line 80: Is this change correct? I believe we'd at least need to add this back to pathways-gcsfuse support eventually. Same with below instances when it's removed.

cc @shauryagup

src/xpk/commands/workload.py
```diff
@@ -241,219 +237,41 @@
...
-                imagePullPolicy: Always
-                name: pathways-proxy
-                ports:
-                - containerPort: 29000
-              hostNetwork: true
-              dnsPolicy: ClusterFirstWithHostNet
-              nodeSelector:
-                cloud.google.com/gke-nodepool: cpu-proxy-np
-    {user_workload}
+# The indentation of PW_WORKLOAD_CREATE_YAML is intentional to allow reusing the user workload container YAML.
```
SujeethJinesh, line 240: Not needed for this, but it's a recurring issue that we may break YAMLs due to poor indentation. E.g. if we indent the receiving block, it could cause errors. I think we should find a way to fix this eventually.

@pawloch00  and XPK team.

src/xpk/core/pathways.py
```diff
@@ -223,58 +146,26 @@ def get_pathways_unified_query_link(args) -> str:
...
-      not second_line
-  ):  # to cover edge case if only one arg remains, we would have to look at the entire YAML in this case.
-    return yaml
-  # Calculate the indentation based on the second line of existing YAML.
-  indentation = ' ' * (len(second_line) - len(second_line.lstrip()))
-  custom_args = custom_args.split(' ')
-  for arg in custom_args:
-    yaml += '\n' + indentation + '- ' + arg
+  yaml = """"""
+  indentation = (
```
SujeethJinesh, line 156: Similar to above comment, fine for now though.

SujeethJinesh (commented):
Thanks Roshani! 

Minor nits and needed a bit better understanding of why we need to remove storage things in this PR.

.github/workflows/reusable_cluster_create.yaml
RoshaniN: Yes, we decided to remove the support for now, per @shauryagup 's suggestion.

src/xpk/commands/workload.py
RoshaniN: Since it's not relevant to this PR, will port this request to a bug for the XPK team to review indentation in general.

RoshaniN:
Tests failing for spurious errors - https://screenshot.googleplex.com/6qPS7LNLtnuj2kN workload already succeeded before next command could be run.

.github/workflows/reusable_cluster_create.yaml
shauryagup: Is there a reason we need to change the cluster create flow with pathways? I would prefer that the cluster stays fungible between McJAX and Pathways as much as possible

.github/workflows/reusable_workload_tests.yaml
```diff
@@ -78,7 +78,7 @@ jobs:
     - name: Wait for workload completion and confirm it succeeded
       run: python3 xpk.py workload list --cluster ${{inputs.cluster-name}} --zone=${{inputs.zone}} --wait-for-job-completion $WORKLOAD_NAME --timeout 300
     - name: Run a Pathways workload on Ubuntu base image
-      run: python xpk.py workload create-pathways --cluster ${{inputs.cluster-name}} --workload $PATHWAYS_WORKLOAD_NAME --docker-image='marketplace.gcr.io/google/ubuntu2004' --tpu-type=${{inputs.tpu-type}} --num-slices=1 --zone=${{inputs.zone}} --command "echo \"Hello world from a test script! \""
+      run: python xpk.py workload create-pathways --cluster ${{inputs.cluster-name}} --workload $PATHWAYS_WORKLOAD_NAME --docker-image='marketplace.gcr.io/google/ubuntu2004' --tpu-type=${{inputs.tpu-type}} --num-slices=1 --zone=${{inputs.zone}} --command "sleep 10 ; echo \"Hello world from a test script! \""
```
shauryagup, line 81: Just curious, any reason for adding the sleep?

src/xpk/commands/workload.py
```diff
@@ -241,219 +237,41 @@
...
-        spec:
-          backoffLimit: {backoff_limit}
-          completions: {system.vms_per_slice}
-          parallelism: {system.vms_per_slice}
-          template:
-            metadata:
-              annotations:
-                {storage_annotations}
-            spec:
-              terminationGracePeriodSeconds: {args.termination_grace_period_seconds}
```
shauryagup, line 279: I think maybe this is another gap but we are missing to propagate `terminationGracePeriodSeconds` on the workers which would be needed for suspend/resume. I would say this is a P1 (should be done before launch) as well.

src/xpk/commands/workload.py
```diff
@@ -241,219 +237,41 @@
...
-                      fieldRef:
-                        fieldPath: "metadata.labels['jobset.sigs.k8s.io/job-index']"
-                  - name: MEGASCALE_COORDINATOR_ADDRESS
-                    value: "$(JOBSET_NAME)-$(REPLICATED_JOB_NAME)-$(MEGASCALE_SLICE_ID)-0.$(JOBSET_NAME)"
-              {pathways_sidecar_container}
-              nodeSelector:
-                {accelerator_label}
-                {machine_label}
-                {autoprovisioning_args}
-              priorityClassName: {args.priority}
```
shauryagup, line 343: Kueue priority is also missing on PathwaysJob API. Till we get this, we would be limited to Pathways jobs through xpk stay at the default medium priority. I would say this is a P1 as well esp. since we call out HotSwap as a feature in Pathways which requires manipulating job priorities.

src/xpk/commands/workload.py
```diff
@@ -241,219 +237,41 @@
...
+      - componentType: worker
+        image: {args.server_image}
+        customFlags: {pathways_worker_args}
+      - componentType: colocated_python_sidecar
+        image: {args.remote_python_sidecar_image} # Provide colocated python sidecar image
+      workers:
+      - type: {machine_type}
+        topology: {topology}
+        numSlices: {args.num_slices}
+        maxSliceRestarts: {args.max_slice_restarts}
```
shauryagup, line 267: I didn't see where you are setting this. We should probably use the same value as xpk has right now which is 4 once `terminationGracePeriodSeconds` is supported!

src/xpk/commands/workload.py
```diff
@@ -655,29 +486,19 @@ def workload_create(args) -> None:
...
-            system.accelerator_type, system
+        topology=create_tpu_topology(system.accelerator_type, system),
+        machine_type=create_tpu_machine_type(system.accelerator_type, system),
+        pathways_worker_args=append_custom_pathways_args(
+            args.custom_pathways_worker_args
         ),
-        machine_label=create_machine_label(system.accelerator_type, system),
-        pathways_worker_args=get_pathways_worker_args(args),
-        pathways_proxy_args=get_pathways_proxy_args(args),
-        pathways_sidecar_container=get_pathways_sidecar_container(args),
```
shauryagup, line 664: Don't we need to pass the sidecar image?

src/xpk/core/nodepool.py
```diff
@@ -138,35 +138,36 @@ def run_gke_node_pool_create_command(
       else:
         node_pools_to_remain.append(node_pool_name)
 
-    # Workload Identity for existing nodepools
-    if args.enable_workload_identity or args.enable_gcsfuse_csi_driver:
```
shauryagup, line 142: I don't see a reason why we can't run this for pathways clusters as well

src/xpk/core/nodepool.py
```diff
@@ -298,8 +299,9 @@ def run_gke_node_pool_create_command(
           f' --scopes=storage-full,gke-default,{CLOUD_PLATFORM_AUTH_SCOPE_URL}'
       )
 
-    if args.enable_workload_identity or args.enable_gcsfuse_csi_driver:
-      command += ' --workload-metadata=GKE_METADATA'
+    if not args.enable_pathways:
```
shauryagup, line 302: Same as above

shauryagup (commented):
Thanks Roshani!

I think in general I would prefer that we align the `create cluster` flow between pathways and non pathways and eventually can deprecate the pathways specializations. I think after the PathwaysJob API integration, there is no reason to have separate pathways CPU nodepools and we should just be able to use the default pool albeit with a bigger default machine type. We can also address this in a separate PR

.github/workflows/reusable_workload_tests.yaml
RoshaniN: Tests failing for spurious errors - https://screenshot.googleplex.com/6qPS7LNLtnuj2kN workload already succeeded before next command could be run. However adding sleep also didn't work so I need to check why the workload is not accessible in the next test step.

src/xpk/core/nodepool.py
RoshaniN: We discussed removing storage support. Hence I removed enable_gcsfuse_csi_driver. "enable_workload_identity" was added in the same PR https://github.com/AI-Hypercomputer/xpk/commit/92df463f636887d04fc78823432673137084351f and I suppose it is to help with other storage features.

Did something change with regards to the decision to remove storage support for Pathways at the moment?

RoshaniN:
> Thanks Roshani!
> 
> I think in general I would prefer that we align the `create cluster` flow between pathways and non pathways and eventually can deprecate the pathways specializations. I think after the PathwaysJob API integration, there is no reason to have separate pathways CPU nodepools and we should just be able to use the default pool albeit with a bigger default machine type. We can also address this in a separate PR

Agree with you, will send a follow up PR next week.

src/xpk/commands/workload.py
RoshaniN: I am passing the sidecar image, but not the whole container (as that is now getting built in PathwaysJob). LMK if there are concerns!

src/xpk/commands/workload.py
RoshaniN: Yes, I am `terminationGracePeriodSeconds` imbibing this in my PathwaysJob PR.

src/xpk/commands/workload.py
RoshaniN: I missed calling out this arg as a missing feature and thanks for catching that! I could add it to workerSpec https://github.com/google/pathways-job/blob/main/api/v1/pathwaysjob_types.go#L123

WDYT?

src/xpk/commands/workload.py
RoshaniN: Sure, also to be added to workerSpec?

src/xpk/commands/workload.py
shauryagup: This is critical to get suspend/resume working

RoshaniN: Yes, I mentioned in a previous comment that I will add this to workerSpec.

RoshaniN: BTW JobSet default is 30 seconds.

// The grace period is the duration in seconds after the processes running in the pod are sent
	// a termination signal and the time when the processes are forcibly halted with a kill signal.
	// Set this value longer than the expected cleanup time for your process.
	// Defaults to 30 seconds.

src/xpk/commands/workload.py
```diff
@@ -241,219 +237,41 @@
...
+        image: {args.proxy_server_image}
+        customFlags: {pathways_proxy_args}
+      - componentType: pathways_server
+        image: {args.server_image}
+        customFlags: {pathways_rm_args}
+      - componentType: worker
+        image: {args.server_image}
+        customFlags: {pathways_worker_args}
+      - componentType: colocated_python_sidecar
+        image: {args.remote_python_sidecar_image} # Provide colocated python sidecar image
```
RoshaniN, line 262: @shauryagup setting the remote python image here.

src/xpk/commands/workload.py
```diff
@@ -241,219 +237,41 @@
...
+      name: {args.workload}
+      labels:
+        kueue.x-k8s.io/queue-name: {local_queue_name}  # Name of the LocalQueue
+        xpk.google.com/workload: {args.workload}
+    spec:
+      maxRestarts: {args.max_restarts}
+      customComponents:
+      - componentType: proxy_server
+        image: {args.proxy_server_image}
+        customFlags: {pathways_proxy_args}
```
shauryagup, line 254: Can we do this conditionally only if the user has provided an override? I would prefer that we control the default value of the server images only in the pathwaysjob API.

src/xpk/commands/workload.py
shauryagup: Thanks but I suspect this would be set to an empty string right now which is the default value. Can we only include this if the user has passed in a non empty `--remote_python_sidecar_image`?

src/xpk/commands/workload.py
shauryagup: 30 sec may be too small and we want to give users flexibility to change it.

src/xpk/commands/workload.py
RoshaniN: Addressed in https://github.com/google/pathways-job/pull/13

src/xpk/commands/workload.py
RoshaniN: Addressed in https://github.com/google/pathways-job/pull/13

src/xpk/commands/workload.py
RoshaniN: Yes, can do that.
Default = "" is a no-op essentially.

src/xpk/commands/workload.py
RoshaniN: Defaults are set in XPK to be the latest images, but I see what you mean. Will add conditionally.

src/xpk/core/nodepool.py
RoshaniN: Thanks, retained these features for cluster creation.  Storage features will be a no op during Pathways workload creation.

src/xpk/core/nodepool.py
RoshaniN: Reverted for cluster level features.

src/xpk/commands/workload.py
RoshaniN: Done, thanks!

shauryagup (approved):
Thanks Roshani! Just a couple minor comments

src/xpk/commands/cluster.py
```diff
@@ -206,6 +207,11 @@ def cluster_create(args) -> None:
   if set_jobset_on_cluster_code != 0:
     xpk_exit(set_jobset_on_cluster_code)
 
+  if args.enable_pathways:
```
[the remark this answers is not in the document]
RoshaniN: Do you mean adding Pathways job controller for the `xpk cluster create` flow, used by McJAX? 

Right now, I am adding it to the `xpk cluster create-pathways` flow with the reasoning that this is required only for a Pathways enabled cluster.

src/xpk/commands/workload.py
```diff
@@ -241,219 +240,38 @@
...
+      name: {args.workload}
+      labels:
+        kueue.x-k8s.io/queue-name: {local_queue_name}  # Name of the LocalQueue
+        xpk.google.com/workload: {args.workload}
+    spec:
+      maxRestarts: {args.max_restarts}
+      customComponents:
+      {custom_pathways_proxy_server}
+      {custom_pathways_worker}
+      {custom_pathways_worker}
```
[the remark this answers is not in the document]
RoshaniN: Thanks, great catch!

src/xpk/core/pathways.py
```diff
@@ -14,15 +14,21 @@
...
-from .cluster import XPK_SA
 from ..core.docker_container import get_user_workload_container
 from ..core.gcloud_context import zone_to_region
 from ..core.nodepool import get_all_nodepools_programmatic
 from ..utils.console import xpk_exit, xpk_print
 from .config import AcceleratorType
-from .storage import Storage, get_storage_volumes_yaml, GCS_FUSE_ANNOTATION
 from .system_characteristics import SystemCharacteristics
 
+DEFAULT_PATHWAYS_SERVER_WORKER_IMAGE = (
```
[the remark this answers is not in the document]
RoshaniN: XPK also has defaults at https://github.com/AI-Hypercomputer/xpk/blob/main/src/xpk/parser/workload.py#L245 . The goal for checking against default value is to find out whether the user has overridden this value, in which case XPK will generate custom component.

src/xpk/core/pathways.py
```diff
@@ -223,58 +154,123 @@ def get_pathways_unified_query_link(args) -> str:
...
+
+def append_custom_pathways_proxy_server(args) -> str:
+  """Append custom Pathways proxy server component using a YAML with proper indentation.
+
+  Returns:
+      yaml (string): yaml with custom proxy server appended.
+  """
+  yaml = """"""
+  if (
+      args.proxy_server_image != DEFAULT_PATHWAYS_PROXY_SERVER_IMAGE
```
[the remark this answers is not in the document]
RoshaniN: Same reasoning as above. Since XPK has defaults set, without this check, a custom component will always be generated, even if the user has not overridden this value.  PLMK if you have a different implementation in mind!

src/xpk/core/pathways.py
```diff
@@ -223,58 +154,123 @@ def get_pathways_unified_query_link(args) -> str:
   return f'https://console.cloud.google.com/logs/query;query={query_params}'
 
 
-def get_pathways_rm_args(args, system: SystemCharacteristics) -> str:
-  """Arguments for the Pathways resource manager.
-  Args:
-    args: user provided arguments for running the command.
+def append_custom_pathways_args(custom_args, prev_indentation=8) -> str:
```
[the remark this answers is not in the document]
RoshaniN: Sure, will rename.

src/xpk/commands/cluster.py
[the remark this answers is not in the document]
shauryagup: Yes, let us avoid separate logics. My assumption is that once you consolidate Pathways head to not need a dedicated CPU nodepool, we can just deprecate the `xpk cluster create-pathways` API.

src/xpk/core/pathways.py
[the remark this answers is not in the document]
shauryagup: Is that needed though? What if you always generate the `customComponent` spec whenever the user provides custom images?

sharabiani (changes requested):
@RoshaniN 

XPK workload flow is tightly coupled with Jobset and Introducing a new type such as pathways-job.pathways.domain/v1 will cause some commands behave unexpectedly  (workload list/inspect,...). Some methods such as `get_workload_list` , `check_if_workload_exists`, `wait_for_job_completion`, and etc are expecting a Jobset.  The failing test is also because of this issue.

@44past4 Can you please advice how to handle this?

src/xpk/core/pathways.py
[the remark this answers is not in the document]
RoshaniN: Exactly - this check is to determine whether user has provided / overridden images. At this point in the code, XPK cannot determine whether user has provided an image or XPK default is set.

We can remove passing the default image from xpk and allow PathwaysJob to handle it, in which case, we don't need to check against the default.

RoshaniN:
> @RoshaniN
> 
> XPK workload flow is tightly coupled with Jobset and Introducing a new type such as pathways-job.pathways.domain/v1 will cause some commands behave unexpectedly (workload list/inspect,...). Some methods such as `get_workload_list` , `check_if_workload_exists`, `wait_for_job_completion`, and etc are expecting a Jobset. The failing test is also because of this issue.
> 
> @44past4 Can you please advice how to handle this?

Thanks. https://github.com/google/pathways-job/ creates a JobSet under the hood. Workload list expects a kueueable workloads (they do not expect JobSet), this is done by setting kueue and workload labels in the JobSet.
Please look at https://b.corp.google.com/issues/368349820#comment15 for local test output where `check_if_workload_exists` is able to find the workloads generated by PathwaysJob.

ahg-g:
How production ready is the PathwaysJob API? What is the criteria to assess its readiness? For example, unless it is setup somewhere else, looking at https://github.com/google/pathways-job/tree/main/test, I noticed that it doesn't have proper integration test coverage, which is major gap.

src/xpk/core/pathways.py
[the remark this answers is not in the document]
shauryagup: A customComponent would only be generated if the user overrides the image. I am asking to remove the defaults from XPK. Is there any drawback to that?

RoshaniN:
> How production ready is the PathwaysJob API? What is the criteria to assess its readiness? For example, unless it is setup somewhere else, looking at https://github.com/google/pathways-job/tree/main/test, I noticed that it doesn't have proper integration test coverage, which is major gap.

I will cover tests today.

src/xpk/commands/cluster.py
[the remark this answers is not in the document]
RoshaniN: For now, there are still separate elements such as not allowing GPU / CPU type and some other checks, so let me handle deprecation of `xpk cluster create-pathways` in a future PR.

src/xpk/core/pathways.py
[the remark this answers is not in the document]
RoshaniN: Done, thanks.

src/xpk/core/pathways.py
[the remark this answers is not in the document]
RoshaniN: Done, I don't see drawbacks at the moment.

src/xpk/commands/cluster.py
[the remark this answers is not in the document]
shauryagup: Yes, that is fine. I just do not want to add more lines of divergence to that.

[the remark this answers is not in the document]
RoshaniN: Okay, done.

src/xpk/parser/workload.py
```diff
@@ -601,9 +633,9 @@ def add_shared_workload_create_optional_arguments(args_parsers):
         ),
     )
     custom_parser.add_argument(
-        '--remote-python-sidecar-image',
+        '--colocated-python-sidecar-image',
```
SujeethJinesh, line 636: We need to adjust the name in a few more places. If you're able to can you make the changes in this PR?

https://github.com/AI-Hypercomputer/xpk/pull/451/files

RoshaniN: The sidecar image function is not needed anymore as PathwaysJob builds the container. Please note.

SujeethJinesh: Got it, thank you! I'll also check in the equivalent change in MaxText once this PR is in

https://github.com/AI-Hypercomputer/maxtext/pull/1532

SujeethJinesh (approved):
Thanks Roshani!

src/xpk/commands/cluster.py
```diff
@@ -206,6 +207,10 @@ def cluster_create(args) -> None:
   if set_jobset_on_cluster_code != 0:
     xpk_exit(set_jobset_on_cluster_code)
 
+  set_pathways_job_on_cluster_code = set_pathways_job_on_cluster(args)
```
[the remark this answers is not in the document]
RoshaniN: Good question! This is similar to how customers upgrade JobSet . Re-running `cluster create` is sufficient before running Pathways workloads, to upgrade the XPK cluster and install it with PathwaysJob.

README.md
```diff
@@ -221,6 +221,7 @@ all zones.
     --num-slices=4 --on-demand \
     --tpu-type=v5litepod-16
     ```
+    Note that Pathways clusters need a CPU nodepool of n2-standard-64 or higher.
```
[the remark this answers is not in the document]
RoshaniN: Yes, this is a new requirement. The `cluster create-pathways` flow will take care of the upgrade.

The nodepool "cpu-user-np" will be upgraded to have n2-standard-64. "cpu-rm-np" and "cpu-proxy-np" will be removed as they are no longer needed.
The goal is to optimize 3 nodepools of smaller CPUs to one nodepool with larger CPU.

src/xpk/commands/workload.py
```diff
@@ -833,6 +650,10 @@ def workload_delete(args) -> None:
   elif not will_delete:
     xpk_print('Skipping delete command.')
   else:
+    # If PathwaysJob exists, delete it.
+    if try_to_delete_pathwaysjob_first(args, workloads):
```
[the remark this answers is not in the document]
RoshaniN: Yes, no restrictions or changes to how workloads can run in parallel.

1. Delete flow is filtered by workloads to be deleted here https://github.com/AI-Hypercomputer/xpk/pull/448/files#diff-07d28ff2ce7b01398c06eb2510bd4b48fe876c9b12aec763e863dba43b2e1b4dR644 based on the workload name.

2. If the workload name is not provided, all eligible workloads on the cluster are shown to user https://github.com/AI-Hypercomputer/xpk/pull/448/files#diff-07d28ff2ce7b01398c06eb2510bd4b48fe876c9b12aec763e863dba43b2e1b4dR630 for delete.

There are no changes to this logic, this is same as before.

The only extra step is to check whether any of the workloads marked for delete from 1 or 2 are PathwaysJob workloads. If so, they can be deleted using PathwaysJob command, else as before, they are deleted by JobSet command. So there are no errors here.

LMK if this answers your question @sharabiani ?

src/xpk/commands/workload.py
```diff
@@ -858,6 +679,40 @@ def workload_delete(args) -> None:
...
+  for workload in workloads:
+    args.workload = workload
+    command = f'kubectl delete pathwaysjob {workload} -n default'
+    task_name = f'PathwaysWorkloadDelete-{workload}'
+    commands.append(command)
+    task_names.append(task_name)
+
+  # Not batching deletion for single workload
+  if len(workloads) == 1:
+    return_code = run_command_with_updates(commands[0], 'Delete Workload', args)
```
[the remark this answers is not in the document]
RoshaniN: return_code=0 means successful delete. This is same as JobSet, no changes in the delete logic.

src/xpk/commands/workload.py
```diff
@@ -545,43 +362,64 @@ def workload_create(args) -> None:
...
-  autoprovisioning_enabled, return_code = is_autoprovisioning_enabled(
-      args, system
-  )
-  if return_code != 0:
-    xpk_exit(return_code)
-  if autoprovisioning_enabled:
-    # Determine NAP capacity type
-    autoprovisioning_args, return_code = (
-        get_autoprovisioning_node_selector_args(args)
+  if not args.use_pathways:
```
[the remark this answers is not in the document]
RoshaniN: I was able to implement this for get_autoprovisioning_args. Storage func headers would get changed everywhere as they do not take args as input but args.storage . Changing storage headers to accept args and then checking for use pathways could be addressed in a follow up PR, I think.

README.md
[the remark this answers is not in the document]
RoshaniN: I have updated nodepool creation deletion logic to be backward compatible now, new nodepool  will be created and older ones will be deleted. https://b.corp.google.com/issues/368349820#comment25

src/xpk/commands/cluster.py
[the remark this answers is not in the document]
RoshaniN: I have updated the `workload create-pathways` flow to indicate to the user to upgrade the cluster using `cluster create-pathways` flow. The new changes made XPK backward compatible.

sharabiani (approved):

PR #449, opened 2025-04-07
Remove sidecar limits for GCS fuse storages

gcie:
## Fixes / Features
- remove GCS fuse sidecar limits
- fixes error when creating workloads: `ValueError: dictionary update sequence element #0 has length 1; 2 is required`

### DIFF
--- a/src/xpk/commands/shell.py
+++ b/src/xpk/commands/shell.py
@@ -88,7 +88,9 @@ def connect_to_new_interactive_shell(args: Namespace) -> int:
       'kubectl-kjob create interactive --profile'
       f' {AppProfileDefaults.NAME.value} --pod-running-timeout 180s'
   )
-  cmd += get_gcsfuse_annotations(args)
+
+  for annotation in get_gcsfuse_annotations(args):
+    cmd += f' --pod-template-annotation {annotation}'
 
   return run_command_with_full_controls(
       command=cmd,
--- a/src/xpk/commands/shell.py
+++ b/src/xpk/commands/shell.py
@@ -20,7 +20,7 @@ from ..core.kjob import (
     AppProfileDefaults,
     prepare_kjob,
     get_pod_template_interactive_command,
-    get_gcsfuse_annotation,
+    get_gcsfuse_annotations,
 )
 
 exit_instructions = 'To exit the shell input "exit".'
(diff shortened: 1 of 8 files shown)


src/xpk/core/storage.py
```diff
@@ -46,7 +46,12 @@
 GCS_FUSE_TYPE = "gcsfuse"
 GCP_FILESTORE_TYPE = "gcpfilestore"
 MANIFESTS_PATH = os.path.abspath("xpkclusters/storage-manifests")
-GCS_FUSE_ANNOTATION = 'gke-gcsfuse/volumes: "true"'
+GCS_FUSE_ANNOTATIONS = {
+    "gke-gcsfuse/volumes": "true",
+    "gke-gcsfuse/cpu-limit": "0",
+    "gke-gcsfuse/memory-limit": "0",
+    "gke-gcsfuse/ephemeral-storage-limit": "0",
```
sharabiani, line 53: AFAIK, the goal is to set the recommended limits and not removing them.
@44past4 Please confirm

src/xpk/commands/workload.py
```diff
@@ -643,7 +641,7 @@ def workload_create(args) -> None:
           storage_volume_mounts=get_storage_volume_mounts_yaml_for_gpu(
               all_storages
           ),
-          storage_annotations=storage_annotations,
+          storage_annotations=get_storage_annotations_yaml(all_storages, 12),
```
sharabiani, line 644: I'm not sure if the offset. I understand that we have yaml generations in core modules (which are not good idea), however I would return the annotations and set the offset in the commands/workload module.

btw, I see there is [a work](https://github.com/AI-Hypercomputer/xpk/pull/448) that removes storage from pathway workloads

src/xpk/core/workload_decorators/storage_decorator.py
```diff
@@ -42,9 +42,9 @@ def decorate_jobset(jobset_manifest_str, storages) -> str:
 def add_annotations(job_manifest, storages):
```
sharabiani, line 42: Are we going to add these annotations for Slurm-mode JobTemplate, too? I mean for xpk run/batch/(shell?) commands.

gcie: Good catch, I think they should be there too

src/xpk/core/storage.py
gcie: From the doc:
>  To make it simple, we recommend removing limits on the sidecar CPU and memory consumption (i.e. enable it to use as much as is needed for your workload).

src/xpk/commands/shell.py
```diff
@@ -88,10 +88,7 @@ def connect_to_new_interactive_shell(args: Namespace) -> int:
       'kubectl-kjob create interactive --profile'
       f' {AppProfileDefaults.NAME.value} --pod-running-timeout 180s'
   )
-
-  gcsfuse_annotation = get_gcsfuse_annotation(args)
-  if gcsfuse_annotation is not None:
-    cmd += f' --pod-template-annotation {gcsfuse_annotation}'
+  cmd += get_gcsfuse_annotations(args)
```
sharabiani, line 91: I wouldn't move the command signature to the core module.

sharabiani (approved):

PR #450, opened 2025-04-07
Add `--mount-options` parameter to `xpk storage attach/create`

gcie:
## Fixes / Features
- add `--mount-options` parameter to `xpk storage attach/create` which allows the user to:
  - Remove `implicit-dirs` option
  - Increase Metadata Cache Values
  - Enable the File Cache + Parallel Downloads
  - Enable streaming writes
  - Set negative stat cache for checkpointing
  - Set `rename-dir-limit`
  - Enable list caching
  - Increase Kernel Read Ahead
  - Enable debug logs
  - Set RAM cache


src/xpk/parser/storage.py
```diff
@@ -153,6 +153,12 @@ def add_storage_attach_parser(
       type=str,
       help='Path to manifest file containing volume definitions',
   )
+  opt_args.add_argument(
```
pawloch00, line 156: Please add an example, or link to docs, explaining what are this options

pawloch00:
Can we have any test with this mount options in e2e tests?

pawloch00 (approved):

PR #452, opened 2025-04-09
Fix `xpk workload create` for L4 and A100

gcie:
## Fixes / Features
- `xpk workload create` for L4 and A100 machines


src/xpk/core/workload.py
```diff
@@ -282,6 +283,44 @@ def get_gpu_volume(system: SystemCharacteristics) -> str:
...
 
 
+def get_gpu_rxdm_container(
+    system: SystemCharacteristics, all_storages: list[Storage]
+) -> str:
+  gpu_rxdm_image = get_gpu_rxdm_image(system)
+  if not gpu_rxdm_image:
+    return ''
+
+  gpu_rxdm_container_template = """{gpu_rxdm_image}
```
pawloch00, line 293: can we have this yaml as sth else, but not string?

pawloch00:
Are we running tests for L4 gpus? they are pretty available imo

src/xpk/core/workload.py
gcie: Without refactoring the whole `commands/workload.py` to not use strings? Technically possible, but it would be better to do it altogether in a large, refactoring task.

gcie:
> Are we running tests for L4 gpus? they are pretty available imo

I don't think so. Should I make a task for it?

src/xpk/core/workload.py
pawloch00: cant't we have it in dictionary, and than have some function to convert it to string? just here

pawloch00:
> > Are we running tests for L4 gpus? they are pretty available imo
> 
> I don't think so. Should I make a task for it?

Imo, it would be nice to have it as part of this task, to be sure that L4 is supported and that the fix is indeed working.

gcie:
> > > Are we running tests for L4 gpus? they are pretty available imo
> > 
> > 
> > I don't think so. Should I make a task for it?
> 
> Imo, it would be nice to have it as part of this task, to be sure that L4 is supported and that the fix is indeed working.

Agreed, it would be nice. But it does not look so simple to implement, we would have to parallelize tests and I'm not sure I can squeeze it in here. The goal of this task was to simply fix the error. Let's do the test refactoring in another one.

src/xpk/core/workload.py
gcie: We would have to pad it properly in order to fit in workload yaml string. It's doable, but it would be less work to change it when refactoring it all.

pawloch00:
I think we can just add another device type on the list where you can choose devices and run it from there. No need to run it each time for now.

src/xpk/core/workload.py
pawloch00: Adding new code in bad-mannered fashion doesn't seem like a good idea for me. But I will left it up to you to decide

pawloch00 (approved):

PR #453, opened 2025-04-09
Pre-populate GCS fuse metadata when mounting volumes

gcie:
## Fixes / Features
- add metadata pre-population to GCS fuse volumes by default. User can disable it using flag `--no-prefetch-metadata` when running `xpk storage attach`.


src/xpk/core/workload.py
```diff
@@ -282,6 +283,44 @@ def get_gpu_volume(system: SystemCharacteristics) -> str:
...
 
 
+def get_gpu_rxdm_container(
+    system: SystemCharacteristics, all_storages: list[Storage]
+) -> str:
+  gpu_rxdm_image = get_gpu_rxdm_image(system)
+  if not gpu_rxdm_image:
+    return ''
+
+  gpu_rxdm_container_template = """{gpu_rxdm_image}
```
pawloch00, line 293: If this is a new yaml, can we have it as template, ditctionary or sth else? But not string :)

src/xpk/commands/storage.py
```diff
@@ -164,7 +164,10 @@ def storage_attach(args: Namespace) -> None:
         manifest = list(yaml.safe_load_all(f))
     else:
       manifest = gcsfuse.manifest(
-          name=args.name, bucket=args.bucket, size=args.size
+          name=args.name,
+          bucket=args.bucket,
+          size=args.size,
+          prefetch_metadata=args.prefetch_metadata,
```
pawloch00, line 170: Please add readme explanation of prefetch_metadata flag

pawloch00:
Can we have any test to verify this flag is working?

pawloch00 (approved):

Issue #462, opened 2025-04-21
The default "mount-options" for xpk breaks Filestore `storage attach`


ndebuhr:
The `implicit-dirs` default value for `mount-options` breaks Filestore attach commands, as this mount option is only valid for GCS Fuse storage. The workaround is to explicitly specify `--mount-options=""`. The documented README.md Filestore attach command does not work, since it doesn't contain this workaround.

PR #463, opened 2025-04-22
Fix error when `xpk storage attach --type=gcpfilestore` without `--mount-options`

gcie:
## Fixes / Features
- #462

### DIFF
--- a/.github/workflows/reusable_storage_tests.yaml
+++ b/.github/workflows/reusable_storage_tests.yaml
@@ -55,7 +55,6 @@ jobs:
     uses: ./.github/workflows/reusable_storage_delete.yaml
     concurrency: # We support one build or nightly test to run at a time currently.
       group: fuse-tests-${{inputs.run-id}}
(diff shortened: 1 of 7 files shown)


.github/workflows/reusable_storage_tests.yaml
```diff
@@ -28,52 +28,102 @@ on:
...
-      STORAGE_NAME: gcpfilestore-test-${{inputs.run-id}}
\ No newline at end of file
+
+  filestore-delete-tests:
+    if: always()
+    needs: [filestore-create-tests, filestore-attach-tests]
+    uses: ./.github/workflows/reusable_storage_delete.yaml
+    concurrency: # We support one build or nightly test to run at a time currently.
+      group: filestore-tests-${{inputs.run-id}}
+      cancel-in-progress: true
```
pawloch00, line 121: Does it mean that we will interrupt delete operation? I think it should be allowed to finish.

PR #466, opened 2025-04-29
`xpk cluster adapt`

gcie:
## Fixes / Features
- new command: `xpk cluster adapt` that prepares an existing cluster for XPK

### DIFF
--- a/src/xpk/commands/cluster.py
+++ b/src/xpk/commands/cluster.py
@@ -162,9 +162,10 @@ def cluster_adapt(args) -> None:
   if set_jobset_on_cluster_code != 0:
     xpk_exit(set_jobset_on_cluster_code)
 
-  set_pathways_job_on_cluster_code = set_pathways_job_on_cluster(args)
-  if set_pathways_job_on_cluster_code != 0:
-    xpk_exit(set_pathways_job_on_cluster_code)
+  # TODO: Uncomment when cluster_adapt will support TPU cluters
+  # set_pathways_job_on_cluster_code = set_pathways_job_on_cluster(args)
+  # if set_pathways_job_on_cluster_code != 0:
+  #   xpk_exit(set_pathways_job_on_cluster_code)
 
   install_kueue(args, system, autoprovisioning_config)
 


src/xpk/commands/cluster.py
```diff
@@ -74,14 +76,192 @@
...
+    err_code = disable_mglru_on_cluster(args)
+    if err_code > 0:
+      xpk_exit(err_code)
+
+  err_code = prepare_kjob(args)
+  if err_code > 0:
+    xpk_exit(err_code)
+
+  k8s_client = setup_k8s_env(args)
+  install_storage_crd(k8s_client)
```
pawloch00, line 219: Can we have this line next to installing storage csi-s ?

src/xpk/commands/cluster.py
```diff
@@ -74,14 +76,192 @@
...
+      xpk_print(
+          'Found unexpected number of nodes. Is the --device-type correct?'
+      )
+      xpk_exit(1)
+    else:
+      xpk_print(f'Using {args.num_nodes} nodes.')
+
+  # ToDo(roshanin@) - Re-enable CloudDNS on Pathways clusters conditionally.
+  # Enable WorkloadIdentity if not enabled already.
+  if args.enable_workload_identity or args.enable_gcsfuse_csi_driver:
```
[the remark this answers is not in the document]
gcie: We were supposed to remove this from `cluster_create` later and call `cluster_adapt` there

src/xpk/core/cluster.py
```diff
@@ -169,6 +199,49 @@ def install_nri_on_cluster(args) -> int:
...
+  """Get list of cluster's nodes descrition in yaml format
+
+  Args:
+    args: user provided arguments for running the command.
+
+  Returns:
+    List of nodes info yaml objects.
+  """
+  xpk_print("Getting cluster's info...")
+  command = 'kubectl get nodes -oyaml'
```
pawloch00, line 212: -o yaml ?

src/xpk/parser/cluster.py
```diff
@@ -454,6 +460,138 @@ def set_cluster_list_parser(cluster_list_parser: ArgumentParser):
   cluster_list_parser.set_defaults(func=cluster_list)
 
 
+def set_cluster_adapt_parser(cluster_adapt_parser: ArgumentParser):
+  ### Required arguments
```
pawloch00, line 464: Remove the comment

src/xpk/parser/cluster.py
```diff
@@ -454,6 +460,138 @@ def set_cluster_list_parser(cluster_list_parser: ArgumentParser):
...
+      '--device-type',
+      type=str,
+      default=None,
+      help=(
+          'The device type used on cluster (can be tpu or gpu or cpu), eg.'
+          ' h100-80gb-8, n2-standard-32-4 etc.'
+      ),
+  )
+
+  ### Optional arguments
```
pawloch00, line 498: Remove the comment

src/xpk/commands/cluster.py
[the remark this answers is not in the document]
pawloch00: I propose we create functions like enable_storage_csis, install_kueue, install_kjob, prepare_gpus and make them shared between cluster create and cluster adapt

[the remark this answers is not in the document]
pawloch00: I see, but this can be refactored into smaller, more specific functions that we will later remove from cluster create.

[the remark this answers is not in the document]
gcie: Creating small functions here in the long run serves the same purpose as simply grouping function calls. 

What I think should be done to tidy up is to replace this scheme:
```python
some_function_code = some_function(args)
if some_function_code != 0:
  xpk_exit(some_function_code)
```
Why are we trying so hard not to use exceptions?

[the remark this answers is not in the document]
pawloch00: You can easily write unit tests for small functions, while testing one long function can be harder. Using exceptions is planned for future work.

src/xpk/core/cluster.py
gcie: Doesn't really matter, both work fine.

src/xpk/parser/cluster.py
```diff
@@ -454,6 +460,138 @@ def set_cluster_list_parser(cluster_list_parser: ArgumentParser):
...
+  cluster_adapt_optional_arguments = cluster_adapt_parser.add_argument_group(
+      'Optional Arguments',
+      'Arguments optional for cluster adapt.',
+  )
+
+  cluster_adapt_optional_arguments.add_argument(
```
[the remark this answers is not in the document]
gcie: I did what I could, only 5 remain. I can't add shared_cluster_create_optional_arguments, because we do not use all of them in cluster adapt.

src/xpk/commands/cluster.py
```diff
@@ -311,34 +242,10 @@ def cluster_create(args) -> None:
...
-    if update_cluster_command_code != 0:
-      xpk_exit(update_cluster_command_code)
-
-  if args.enable_parallelstore_csi_driver:
-    update_cluster_command_code = (
-        update_cluster_with_parallelstore_driver_if_necessary(args)
-    )
-    if update_cluster_command_code != 0:
-      xpk_exit(update_cluster_command_code)
+  k8s_client = setup_k8s_env(args)
```
pawloch00, line 245: can we reuse k8s from lines 126

gcie: This is a scope of a different function.

pawloch00:
What about testing this? Is it possible to create a testing scenario with tpu involved? For example create cluster with gcloud command and run xpk adapt on it?

gcie:
> What about testing this? Is it possible to create a testing scenario with tpu involved? For example create cluster with gcloud command and run xpk adapt on it?

We will have coverage in github workflow after we refactor cluster_create to use cluster_adapt.

> What about the scenario where cluster is created with cluster toolkit, than xpk cluster adapt is done and than we want to remove cluster using xpk cluster delete?

I tested it manually and it works. If you want to cover it with tests, then we need to have integration tests for GPUs finished.

pawloch00:
But it should work for tpu too, thus making it eligible for testing now.

gcie:
> But it should work for tpu too, thus making it eligible for testing now.

We will have tests with TPU in github workflows after refactoring due to cluster_create depending on cluster_adapt. Also, the scope of this task was to make this command for A3 ultra only. We'll add TPU support in later PRs.

pawloch00:
> > But it should work for tpu too, thus making it eligible for testing now.
> 
> We will have tests with TPU in github workflows after refactoring due to cluster_create depending on cluster_adapt. Also, the scope of this task was to make this command for A3 ultra only. We'll add TPU support in later PRs.

Than it needs to be clearly specified in Readme that it is supported only for A3 ultra

src/xpk/commands/cluster.py
```diff
@@ -833,3 +881,87 @@ def run_gke_cluster_create_command(
...
+  if install_kueue_on_cluster_code != 0:
+    xpk_exit(install_kueue_on_cluster_code)
+
+  xpk_print('Wait for Kueue to be fully available')
+  wait_for_kueue_available_code = wait_for_kueue_available(args)
+  if wait_for_kueue_available_code != 0:
+    xpk_exit(wait_for_kueue_available_code)
+
+  xpk_print('Install Kueue Custom Resources')
+  enable_kueue_credentials_code = install_kueue_crs(
```
pawloch00, line 944: I think yaml string used in install kueue crs is missing TAS which is used for a3 ultra

src/xpk/commands/cluster.py
```diff
@@ -75,14 +77,122 @@
...
+  args.enable_pathways = False
+
+  system, return_code = get_system_characteristics(args)
+
+  if return_code > 0:
+    xpk_print('Fetching system characteristics failed!')
+    xpk_exit(return_code)
+
+  xpk_print(
+      f'Starting cluster adaptation for cluster {args.cluster}:', flush=True
```
pawloch00, line 95: why do we have flush=True here? Its not used in other places

src/xpk/commands/cluster.py
```diff
@@ -75,14 +77,122 @@
...
+  if args.enable_autoprovisioning:
+    xpk_print('Enabling Autoprovisioning')
+    autoprovisioning_config, return_code = enable_autoprovisioning_on_cluster(
+        args, system
+    )
+    if return_code != 0:
+      xpk_exit(return_code)
+
+  xpk_print('Creating ConfigMap for cluster')
+  create_cluster_configmaps_code = create_cluster_configmaps(
```
pawloch00, line 151: shouldn't we check if configmap already exists ?

gcie: The function `create_cluster_configmaps()` creates or updates the configmap. I don't see a reason why we shouldn't update it.

pawloch00 (approved):

src/xpk/core/cluster.py
```diff
@@ -169,6 +199,49 @@ def install_nri_on_cluster(args) -> int:
...
+  """Count cluster nodes by accelerator type"""
+  nodes_info = get_cluster_nodes_info(args)
+  accelerators = [
+      node['metadata']['labels']['cloud.google.com/gke-accelerator']
+      for node in nodes_info
+      if 'cloud.google.com/gke-accelerator' in node['metadata']['labels']
+  ]
+  accelerator = None
+  if system.device_type == H200_DEVICE_TYPE:
+    accelerator = 'nvidia-h200-141gb'
```
sharabiani, line 234: `accelerator = system.gke_accelerator` ?

src/xpk/core/cluster.py
```diff
@@ -169,6 +199,49 @@ def install_nri_on_cluster(args) -> int:
...
+  """Get list of cluster's nodes descrition in yaml format
+
+  Args:
+    args: user provided arguments for running the command.
+
+  Returns:
+    List of nodes info yaml objects.
+  """
+  xpk_print("Getting cluster's info...")
+  command = 'kubectl get nodes -o yaml'
```
sharabiani, line 212: Are we going to consider the nodes only in the default namespace?

src/xpk/commands/cluster.py
```diff
@@ -75,14 +77,122 @@
...
+
+  xpk_print(
+      'Enabling the jobset API on our cluster, to be deprecated when Jobset is'
+      ' globally available'
+  )
+  set_jobset_on_cluster_code = set_jobset_on_cluster(args)
+  if set_jobset_on_cluster_code != 0:
+    xpk_exit(set_jobset_on_cluster_code)
+
+  set_pathways_job_on_cluster_code = set_pathways_job_on_cluster(args)
```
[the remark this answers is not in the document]
sharabiani: Also shouldn't it be limited for TPUs?

sharabiani (commented):
And one more question about additional networks and subnetwork names for A3U machine. Are we assuming the subnetwork names are as we expect in the A3 workloads with RDMA annonations?

src/xpk/core/cluster.py
gcie: Nodes are not namespaced.

gcie:
> And one more question about additional networks and subnetwork names for A3U machine. Are we assuming the subnetwork names are as we expect in the A3 workloads with RDMA annonations?

Can you let me know why you think it would matter in this case? Btw, `xpk workload create` now automatically set subnetwork names on the fly to whatever was configured on cluster.

PR #470, opened 2025-05-09
Slurm commands guide

kzmyslona:
Guide detailing Slurm functionalities in XPK on finetuning Llama 3 8B example.


gcie:
Duplicate of #472

PR #475, opened 2025-05-14
Add Jobset controller patching for MTC cluster

abhinavclemson:
## Fixes / Features
Multi-tier checkpointing Cluster creation with Patched Jobset (This patch would enable scale testing), otherwise scale testing breaks at 1k VMs with MTC enabled.

The manager container in the jobset deployment should have the memory request increased to 1Gi and the limit to 2Gi, and the CPU request increased to 1.

We are basically want to update jobset memory and cpu limit for MTC Cluster with given below values, if they are lower: 
`"value": {"limits": {"memory": "2Gi"}, "requests": {"cpu": "1", "memory": "1Gi"}}}]'`

For high-scale training, our testing has shown that the jobset controller needs at least e2-standard-4 machines; the GKE default of e2-medium does not have enough memory.

Bug link: b/415019070

## Testing / Documentation
link: [Tested on v4 cluster creation
](https://pantheon.corp.google.com/kubernetes/deployment/us-central2/abhinavsing-in-mem/jobset-system/jobset-controller-manager/yaml/view?cloudshell=true&e=13802955&inv=1&invt=AbxVnw&jsmode=o&mods=allow_workbench_image_override&project=cloud-tpu-multipod-dev)

- [x] Tests pass
- [x] Appropriate changes to documentation are included in the PR

### DIFF
--- a/src/xpk/core/mtc.py
+++ b/src/xpk/core/mtc.py
@@ -14,10 +14,16 @@ See the License for the specific language governing permissions and
 limitations under the License.
 """
 
-from ..utils.console import xpk_exit, xpk_print
-from ..utils import templates
-from ..utils.kubectl import apply_kubectl_manifest
+import requests
+import yaml
+
+from ..core.cluster import JOBSET_VERSION
 from ..core.cluster import setup_k8s_env
+from ..utils import templates
+from ..utils.console import xpk_exit
+from ..utils.console import xpk_print
+from ..utils.kubectl import apply_kubectl_manifest
+
 
 MTC_CPC_PATH = "/../templates/mtc-cpc.yaml"
 
@@ -28,6 +34,17 @@ def create_mtc_cpc(
     mtc_toleration_key: str,
     mtc_ramdisk_size: str,
 ) -> dict:
+  """Create MTC Checkpoint Configuration.
+
+  Args:
+    mtc_gcs_bucket: GCS bucket for MTC
+    mtc_machine_type: Machine type for MTC
+    mtc_toleration_key: Toleration key for MTC
+    mtc_ramdisk_size: Ramdisk size for MTC
+
+  Returns:
+    MTC Checkpoint Configuration
+  """
   data = templates.load(MTC_CPC_PATH)
 
   data["spec"]["cloudStorageBucketName"] = mtc_gcs_bucket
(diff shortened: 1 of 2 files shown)


src/xpk/core/mtc.py
```diff
@@ -62,16 +68,111 @@ def install_mtc_on_cluster(args, system) -> int:
...
+
+  if manifest_content is None:
+    xpk_print("Manifest content not found.")
+    xpk_exit(1)
+
+  # Load all YAML documents from the manifest
+  yaml_data_list = list(yaml.safe_load_all(manifest_content))
+  # Iterate through the yaml_data to find the Deployment for
+  # jobset-controller-manager
+  update_manifest = 0
```
RoshaniN, line 120: Let us use bool update_manifest

src/xpk/core/mtc.py
```diff
@@ -62,16 +68,111 @@ def install_mtc_on_cluster(args, system) -> int:
...
+              container["resources"].get("requests", {}).get("memory", "0Mi")
+          )
+          current_memory_limit = (
+              container["resources"].get("limits", {}).get("memory", "0Mi")
+          )
+
+          # Define new values for comparison
+          new_cpu_request = "1000m"
+          new_memory_request = "1Gi"
+          new_memory_limit = "2Gi"
```
RoshaniN, line 147: Please add a readme comment on updated resource limits for the controller in an MTC cluster. https://github.com/AI-Hypercomputer/xpk/blob/develop/src/xpk/parser/cluster.py#L534 

If the user requests a really small CPU machine for the default pool, we want to make sure that jobset-controller can still be hosted.

src/xpk/core/mtc.py
```diff
@@ -62,16 +68,111 @@ def install_mtc_on_cluster(args, system) -> int:
   if args.mtc_toleration_key is None:
     args.mtc_toleration_key = "google.com/tpu"
 
+  k8s_api_client = setup_k8s_env(args)
```
RoshaniN, line 71: Could you please add details about MTC in the README and also talk a little bit about the requirements and the feature in general?

RoshaniN (changes requested):
Thank you, Abhinav!

src/xpk/core/mtc.py
RoshaniN: I don't see a README update for MTC yet - am I missing something?

abhinavclemson: 1). It is not GA'ed yet, customer can't use it for now.
2). README won't be helpful, this should be part of functionality, something like given machine is part of specific set of machine-types, otherwise fail the workload throwing an error. 

Currently, I have added the details in the description. Will raise a PR regarding #2.

RoshaniN: Okay, SG. Please ensure that you leave a README update about MTC cluster closer to availability.

PR #481, opened 2025-05-28
Increase the number of CPU nodes in the Pathways CPU nodepool.

RoshaniN:
## Fixes / Features
This feature only affects Pathways enabled clusters.
- Workloads are bottlenecked by the number of CPU nodes on clusters with Pathways enabled. 
- Increasing the default number of CPU nodes to 4 (which was 1 previously).
- Making this change backward compatible by resizing the Pathways CPU nodepool, if the cluster is already Pathways enabled.

## Testing / Documentation
Tested on a cluster with 19 v5litepod-32

- [ y/n ] Tests pass


src/xpk/core/nodepool.py
```diff
@@ -315,13 +315,20 @@ def run_gke_node_pool_create_command(
...
     for node_pool_name in desired_pw_cpu_node_pools:
       if node_pool_name in existing_node_pool_names:
-        continue
-      command = (
-          'gcloud beta container node-pools create'
-          f' {node_pool_name} --node-version={gke_node_pool_version} --cluster={args.cluster} --project={args.project} --node-locations={args.zone} --region={zone_to_region(args.zone)} --num-nodes=1'
-          f' --machine-type={args.pathways_gce_machine_type} --scopes=storage-full,gke-default,{CLOUD_PLATFORM_AUTH_SCOPE_URL} --enable-autoscaling'
-          ' --min-nodes=1 --max-nodes=20'
-      )
+        # Resize Pathways CPU nodepool to have at least 4 nodes
```
sharabiani, line 318: LGTM, however I believe it will downsize a node-pool if the number of nodes are already bigger than 4 nodes.

PR #483, opened 2025-06-01
Enable DNS Access to Prevent Connection Timeout Errors

SujeethJinesh:
## Fixes / Features
- Fixes a `Unable to connect to the server: dial tcp [34.105.61.132:443](http://34.105.61.132:443/): connect: connection timed out` issue we've been encountering (b/421683213). Any newly created cluster will have DNS enabled and use that.

## Testing / Documentation
Tested on an internal cluster.

- [ y ] Tests pass

### DIFF
--- a/src/xpk/commands/common.py
+++ b/src/xpk/commands/common.py
@@ -33,6 +33,7 @@ def set_cluster_command(args) -> int:
   command = (
       'gcloud container clusters get-credentials'
       f' {args.cluster} --region={zone_to_region(args.zone)}'
+      ' --dns-endpoint'
       f' --project={args.project} &&'
       ' kubectl config view && kubectl config set-context --current'
       ' --namespace=default'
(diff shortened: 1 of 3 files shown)


src/xpk/commands/common.py
```diff
@@ -33,6 +33,7 @@ def set_cluster_command(args) -> int:
   command = (
       'gcloud container clusters get-credentials'
       f' {args.cluster} --region={zone_to_region(args.zone)}'
+      ' --dns-endpoint'
```
RoshaniN, line 36: I have the same question as @pawloch00 - what happens when a cluster is already created, without ' --enable-dns-access' . Does getting credentials with ' --dns-endpoint' still work?

RoshaniN (commented):
Thanks Sujeeth!

src/xpk/commands/common.py
RoshaniN: Thanks!

(Perhaps best to find out how we can determine whether dns access is enabled on a cluster or not. However, I don't think the PR needs to be blocked on it, since you've verified that the change is backward compatible.)

PR #489, opened 2025-06-04
Tpu Slice Controller - Initial Commit

gabesaba:
Generated using `kubebuilder init --repo tpu-slice-controller`


mimowo:
As the starting point it lgtm
/lgtm

PR #494, opened 2025-06-10
Fix broken A3 High workloads

gcie:
## Fixes / Features
- fix broken A3 High workloads
- unify workload creation workflow of A3 High with other A* machines


src/xpk/commands/workload.py
```diff
@@ -455,7 +460,12 @@ def workload_create(args) -> None:
     if not is_TAS_possible(args):
       kueue_TAS_annotation = ''
 
-    if system.device_type in cluster_gcluster.supported_device_types:
+    if system.device_type in [
```
pawloch00, line 463: I thing it would be nicer to use cluster_gcluster.supported_device_types and append to it. When we add new gpu type supported by CT we have to add it manually here and remember about it. Using cluster_gcluster.supported_device_types seems to remove this problem.

src/xpk/core/storage.py
```diff
@@ -365,101 +365,6 @@ def get_storage_annotations(storages: list[Storage]) -> list[str]:
   return annotations
 
 
-def get_storage_volume_mounts_yaml(storages: list[Storage]) -> str:
```
pawloch00, line 368: I don't like we have so much dead code. We may want either to change our linters configuration or introduce an extra check to test for it.

PR #501, opened 2025-06-16
feat: Added an update to CoreDNS

DannyLiCom:
## Fixes / Features
Enhance DNS Scalability for Large-Scale Testing:
* Issue: During large-scale load testing, the existing Kube-DNS solution was found to be insufficient in supporting the demands of McJAX and Pathways TPU paths, leading to potential performance bottlenecks.
* Solution: Adjusted configurations to default to CoreDNS for McJAX and Pathways TPU paths.

## Testing / Documentation
When using the command ```python3 xpk.py cluster create-pathways``` to create a cluster, CoreDNS will be used by default. For the general command ```python3 xpk.py cluster create```, kube-dns is still the default.

- [ y/n ] Tests pass

### DIFF
--- a/src/xpk/commands/cluster.py
+++ b/src/xpk/commands/cluster.py
@@ -869,15 +869,6 @@ def cleanup_coredns_repo(coredns_repo_full_path: str):
 def update_coredns(args):
   """Updates and deploys CoreDNS within a cluster.
 
-  This function performs the following steps:
-  1. Installs 'jq'.
-  2. Clones the CoreDNS deployment repository from GitHub if it doesn't already exist.
-  3. Deploys CoreDNS to the cluster.
-  4. Scales down the 'kube-dns-autoscaler' and 'kube-dns' deployments.
-  5. Scales up the 'coredns' deployment to 15 replicas.
-  6. Waits for kube-dns to scale down and coredns to be ready using kubectl wait.
-  7. Cleans up the cloned repository.
-
   Args:
     args: user provided arguments for running the command.
 
--- a/src/xpk/commands/cluster.py
+++ b/src/xpk/commands/cluster.py
@@ -75,7 +75,7 @@ from ..utils.console import get_user_input, xpk_exit, xpk_print
 from ..utils.file import write_tmp_file
 from . import cluster_gcluster
 from .common import set_cluster_command
-
+import os
 
 def cluster_adapt(args) -> None:
   """Function that performs cluster adaptation.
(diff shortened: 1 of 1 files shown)


src/xpk/commands/cluster.py
```diff
@@ -696,6 +697,146 @@ def cluster_create_ray_cluster(args) -> None:
...
+  4. Deploys CoreDNS to the cluster.
+  5. Scales down the 'kube-dns-autoscaler' and 'kube-dns' deployments.
+  6. Scales up the 'coredns' deployment to 15 replicas.
+
+  Args:
+    args: user provided arguments for running the command.
+
+  Returns:
+    0 if successful and 1 otherwise.
+  """
```
SujeethJinesh, line 716: It would be good to add in an xpk_print statement saying that you are scaling down kubedns and scaling up coredns.

src/xpk/commands/cluster.py
```diff
@@ -696,6 +697,146 @@ def cluster_create_ray_cluster(args) -> None:
...
+  home_dir = os.path.expanduser("~")
```
SujeethJinesh, line 717: I think we should put this in the `/tmp/` directory, and also clean it up at the end once we're done with it (assuming we don't need the deployment package anymore).

src/xpk/commands/cluster.py
```diff
@@ -696,6 +697,146 @@ def cluster_create_ray_cluster(args) -> None:
...
+      f'git clone https://github.com/coredns/deployment.git {coredns_repo_full_path}'
+    )
+    xpk_print(f"[XPK] Task: 'Clone deployment ' in progress, Target directory:{coredns_repo_full_path}.")
+    return_code = run_command_with_updates(
+      command_git_clone, 'Clone deployment', args
+    )
+    if return_code != 0:
+      xpk_print(f'[XPK] Clone deployment error {return_code}')
+      xpk_exit(return_code)
+
```
SujeethJinesh, line 746: nit, extra spacing

src/xpk/commands/cluster.py
```diff
@@ -696,6 +697,146 @@ def cluster_create_ray_cluster(args) -> None:
...
+
+  cluster_name = os.environ.get('CLUSTER_NAME')
+  region = os.environ.get('REGION')
+  project = os.environ.get('PROJECT')
```
SujeethJinesh, line 750: These should be retrievable from the args. Also makes it so that we can get rid of the next checks I believe.

src/xpk/commands/cluster.py
```diff
@@ -696,6 +697,146 @@ def cluster_create_ray_cluster(args) -> None:
...
+    f'--region={region} --project={project} && kubectl config view '
+    f'&& kubectl config set-context --current --namespace=default'
+  )
+  xpk_print(f"[XPK] Task: 'Get cluster credentials' in progress.")
+  return_code = run_command_with_updates(
+    command_get_credentials, 'Get cluster credentials', args
+  )
+  if return_code != 0:
+    xpk_print(f'[XPK] Failed to get cluster credentials {return_code}')
+    xpk_exit(return_code)
```
SujeethJinesh, line 771: There's a command already for this you can use - https://github.com/AI-Hypercomputer/xpk/blob/59ed3e783bfc3a1bd601b57a970e5de75162ff92/src/xpk/core/cluster.py#L722

src/xpk/commands/cluster.py
```diff
@@ -696,6 +697,146 @@ def cluster_create_ray_cluster(args) -> None:
...
+
+  if not os.path.isdir(coredns_k8s_path):
+    xpk_print(f"[XPK] Error：CoreDNS Kubernetes path '{coredns_k8s_path}' does not exist. Has git clone been successful?")
+    xpk_exit(1)
+
+  # Remember the current directory so that you can restore it later
+  original_cwd = os.getcwd()
```
SujeethJinesh, line 778: Curious, can we execute the script from the current directory and just point to it rather than cd-ing to it? E.g.`./{path_to_core_dns}/deploy.sh | kubectl apply -f -`. If not, then this may be ok since the latter `kubectl apply -f -` may not be possible?

src/xpk/commands/cluster.py
```diff
@@ -696,6 +697,146 @@ def cluster_create_ray_cluster(args) -> None:
...
+    xpk_print(f"[XPK] Task: 'Deploy CoreDNS' in progress, Located at '{coredns_k8s_path}'")
+    
+    return_code = run_command_with_updates(
+        command_deploy_coredns, 'Deploy CoreDNS', args
+    )
+    if return_code != 0:
+      xpk_print(f'[XPK] Deploy CoreDNS error {return_code}')
+      xpk_exit(return_code)
+
+  finally:
```
SujeethJinesh, line 795: Just wanted to check, does this finally block run if we `xpk_exit`?

src/xpk/commands/cluster.py
```diff
@@ -696,6 +697,146 @@ def cluster_create_ray_cluster(args) -> None:
...
+    'kubectl scale deployment kube-dns-autoscaler --replicas=0 --namespace=kube-system'
+  )
+  xpk_print(f"[XPK] Task: 'Scaling down kube-dns-autoscaler' in progress")
+  return_code = run_command_with_updates(
+    command_autoscaler_scale_down, 'Scale down kube-dns-autoscaler', args
+  )
+  if return_code != 0:
+    xpk_print(f'[XPK] Scale down kube-dns-autoscaler error {return_code}')
+    xpk_exit(return_code)
+  xpk_print("\n[XPK] kube-dns-autoscaler has been scaled down.")
```
SujeethJinesh, line 810: I'd just add a comment here to note that the scale down command has started, but that scale down itself may take some time, and same with the one below.

src/xpk/commands/cluster.py
```diff
@@ -696,6 +697,146 @@ def cluster_create_ray_cluster(args) -> None:
...
+  return_code = run_command_with_updates(
+    command_dns_scale_down, 'Scale down kube-dns', args
+  )
+  if return_code != 0:
+    xpk_print(f'[XPK] Scale down kube-dns error {return_code}')
+    xpk_exit(return_code)
+  xpk_print("\n[XPK] kube-dns has been scaled down.")
+
+  command_coredns_scale = (
+    'kubectl scale deployment coredns --replicas=15 -n kube-system'
```
SujeethJinesh, line 826: (Non blocking comment) Hmm, I'm wondering if we will want to make the replicas scale with the number of nodes in the future. For now, this should be fine.

src/xpk/commands/cluster.py
```diff
@@ -696,6 +697,146 @@ def cluster_create_ray_cluster(args) -> None:
...
+  )
+  xpk_print(f"[XPK] Task: 'Scale CoreDNS' in progress")
+  return_code = run_command_with_updates(
+    command_coredns_scale, 'Scale CoreDNS', args
+  )
+  if return_code != 0:
+    xpk_print(f'[XPK] Scale CoreDNS error {return_code}')
+    xpk_exit(return_code)
+
+  xpk_print("\n[XPK] The CoreDNS setup process has been completed.")
```
SujeethJinesh, line 836: I'm wondering if we should block on the coreDNS scale up. If the scale up is not successful after maybe 2 minutes, then that could indicate a problem. I'm leaning towards yes, because we definitely want to avoid problems in the future where people may assume that coredns is up, but it failed and we weren't informed about it during cluster creation.

src/xpk/commands/cluster.py
```diff
@@ -890,6 +1031,9 @@ def run_gke_cluster_create_command(
   if return_code != 0:
     xpk_print(f'GKE Cluster Create request returned ERROR {return_code}')
     return 1
+  else:
+    if args.enable_pathways == True:
+      update_coredns(args)
```
SujeethJinesh, line 1036: We should probably make this an "update_coredns_if_necessary". We often recreate clusters and so it would be good to check if coredns is already up rather than go through the expensive procedure of recreating it. We can check to see if the coredns pods are up (and probably output how many are there as well).

src/xpk/commands/cluster.py
DannyLiCom: I've added a ```check_coredns_status``` function to verify if CoreDNS is Ready.

src/xpk/commands/cluster.py
DannyLiCom: I've added an ```update_coredns_if_necessary``` function that checks if CoreDNS is already installed; if so, it skips, otherwise, it proceeds with the setup.

src/xpk/commands/cluster.py
```diff
@@ -697,6 +704,280 @@ def cluster_create_ray_cluster(args) -> None:
...
+        xpk_print(f"Error: Deployment '{deployment_name}' not found in namespace '{namespace}'.")
+      else:
+        xpk_print(f"API error when reading deployment status: {e}")
+      return False
+    except Exception as e:
+      xpk_print(f"Unknown error occurred while checking deployment: {e}")
+      return False
+
+    # Checking the status of all CoreDNS Pods
+    try:
```
pawloch00, line 752: code in this block could be separate function

pawloch00:
Did testing scenarions included only TPU clusters? What with clusters created with GPUs, especially A3 Mega or ultra, that are created using cluster toolkit?

src/xpk/commands/cluster.py
```diff
@@ -697,6 +704,280 @@ def cluster_create_ray_cluster(args) -> None:
...
+    )
+    xpk_print(f"Task: 'Clone deployment ' in progress, Target directory:{coredns_repo_full_path}.")
+    return_code = run_command_with_updates(
+      command_git_clone, 'Clone deployment', args
+    )
+    if return_code != 0:
+      xpk_print(f'Clone deployment error {return_code}')
+      xpk_exit(return_code)
+
+  command_get_credentials = (
```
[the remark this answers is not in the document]
DannyLiCom: That is a bit strange. When I was testing, it also required get_credentials to be run again for it to execute.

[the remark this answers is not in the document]
DannyLiCom: @SujeethJinesh Since ```get_cluster_credentials()``` is already being used within ```update_coredns_if_necessary()```, I can remove it from its current location.

src/xpk/commands/cluster.py
```diff
@@ -697,6 +709,332 @@ def cluster_create_ray_cluster(args) -> None:
...
+    deployment_ok = _check_deployment_readiness(args, app_v1, namespace, deployment_name)
+    if not deployment_ok:
+      time.sleep(5) # Wait before retrying deployment check
+      continue
+
+    # 3. Check all CoreDNS Pods status
+    pods_ok = _check_coredns_pods_status(v1, namespace)
+    if not pods_ok:
+      time.sleep(5) # Wait before retrying pod check
+      continue
```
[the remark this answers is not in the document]
SujeethJinesh: Once this is resolved let me know so I can re-review this.

SujeethJinesh (approved):
LGTM from my perspective, but I'm not an XPK owner, so best to get verification from @Obliviour  or @pawloch00

wstcliyu (approved):

src/xpk/commands/cluster.py
```diff
@@ -245,6 +247,11 @@ def cluster_create(args) -> None:
 
   get_cluster_credentials(args)
 
+  if args.enable_pathways == True:
```
Obliviour, line 250: Don't we want coredns for non-pathways large scale clusters as well?

src/xpk/commands/cluster.py
```diff
@@ -696,6 +701,211 @@ def cluster_create_ray_cluster(args) -> None:
...
+  _install_jq(args)
+
+  # 2. Clone CoreDNS deployment repository
+  _clone_coredns_deployment_repo(args, coredns_repo_full_path)
+
+  # 3. Deploy CoreDNS to the cluster
+  _deploy_coredns_manifests(args, coredns_k8s_path)
+
+  # 4. Scale down kube-dns-autoscaler
+  _scale_down_deployment(args, "kube-dns-autoscaler")
```
pawloch00, line 846: we don't use _fn_name anywhere else in code. Please change it to fn_name

src/xpk/commands/cluster.py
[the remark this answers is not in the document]
pawloch00: Please remove this

src/xpk/commands/cluster.py
pawloch00: hardcoded value with no explanation why 15

src/xpk/commands/cluster.py
```diff
@@ -697,6 +703,268 @@ def cluster_create_ray_cluster(args) -> None:
...
+    shutil.rmtree(coredns_repo_full_path)
+    xpk_print(f'Successfully deleted directory: {coredns_repo_full_path}')
+  except OSError as e:
+    xpk_print(f'Error deleting directory {coredns_repo_full_path}: {e}')
+
+
+def update_coredns(args):
+  """Updates and deploys CoreDNS within a cluster.
+
+  This function performs the following steps:
```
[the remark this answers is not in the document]
DannyLiCom: @pawloch00 Deleted

pawloch00:
> Okay, Thanks!

Please accept collaborator invitation and rerun the tests

PR #510, opened 2025-06-26
[slice] Parse TAS Assignment into a NodeSelector

mszadkow:
## Fixes / Features
Parse TAS Assignment into a NodeSelector.


PBundyra:
Let's merge now and do some cleanups in the follow up

PR #514, opened 2025-06-30
Make the NodePools field optional.

mbobrovskyi:
## Fixes / Features
- Make the NodePools field optional.


IrvingMg (approved):
LGTM

mbobrovskyi:
cc: @gabesaba

slice/api/v1alpha1/slice_types.go
```diff
@@ -32,11 +32,13 @@ type SliceSpec struct {
 	// +kubebuilder:validation:Immutable
 	AcceleratorTopology string `json:"acceleratorTopology"`
 
-	// Required, set of nodePools to use to form slice
+	// Optional, set of nodePools to use to form slice
+	// +optional
 	NodePools []string `json:"nodePools"`
```
sgysherry, line 37: `NodePools` fields is no longer needed, and will be kept as `NodeSelector` only. 

The nodeSelector is actually a **replacement** of nodePool field.  Example usage: 
```
NodeSelector map[string][]string `json:"nodeSelector"`
```
For example, to select nodes in cubes cube-1 and cube-2, you could use
```
       {"cloud.google.com/gke-tpu-reservation-subblock": ["cube-1", "cube-2"]}
```
For example, to select nodes in nodepool nodepool-1 and nodepool-2, you could use
```
       {"cloud.google.com/gke-nodepool": ["nodepool-1", "nodepool-2"]}
```
Example CR:
```
apiVersion: slice.accelerator.gke.io/v1alpha1
kind: Slice
metadata:
  # Name of the Slice resource
  name: my-simple-slice
  # Namespace where the Slice resource will be created
  namespace: default
spec:
  # Specify the type of accelerator for this slice
  acceleratorType: "tpu-v4-podslice"
  # Define the desired topology for the accelerator slice
  acceleratorTopology: "2x2x1"
  # NodeSelector specifies a set of label-based selectors for nodes that can form the slice.
  nodeSelector:
    cloud.google.com/gke-nodepool:
      - "nodepool-1"
```

slice/api/v1alpha1/slice_types.go
```diff
@@ -32,11 +32,13 @@ type SliceSpec struct {
...
 
 	// Optional, set of nodeSelector to use to form slice
```
sgysherry, line 39: This will be a required field, and here is the comment explanation on design doc
```
        // Required, set of nodes to use to form a slice.
        // NodeSelector specifies a set of label-based selectors for nodes that can form the   
        // slice. The controller will select nodes where for each key-value pair in the map,              // the node's label value for that key is present in the corresponding string slice.          
        // This allows for a flexible "match any of these values for this label" selection.
        // The nodeSelector will follow an AND over the map entries but an OR within the list
        // items of the entry.
        // For example, to select nodes in cubes cube-1 and cube-2, you could use: 
        // {"cloud.google.com/gke-tpu-reservation-subblock": ["cube-1", "cube-2"]}
```

mbobrovskyi: Done.

PR #515, opened 2025-06-30
Don't create Slice if Workload doesn't have Topology Assignment.

mbobrovskyi:
## Fixes / Features
- Don't create Slice if Workload doesn't have Topology Assignment.


slice/internal/controller/workload_controller.go
```diff
@@ -110,18 +110,20 @@ func (r *WorkloadReconciler) newSlice(wl *kueue.Workload) (*v1alpha1.Slice, erro
...
-				}
-				if slice.Spec.NodeSelector[TPUReservationSubblockLabel] == nil {
-					slice.Spec.NodeSelector[TPUReservationSubblockLabel] = []string{}
-				}
-				// make sure there are no duplicates in the nodeSelector
-				for _, v := range domain.Values {
-					exists := slices.Contains(slice.Spec.NodeSelector[TPUReservationSubblockLabel], v)
-					if !exists {
-						slice.Spec.NodeSelector[TPUReservationSubblockLabel] = append(slice.Spec.NodeSelector[TPUReservationSubblockLabel], v)
+			if psa.TopologyAssignment != nil {
```
mbobrovskyi, line 113: It can be possible that Workload have PodSetAssignments but doesn't have TopologyAssignment.

slice/internal/controller/workload_controller.go
```diff
@@ -110,18 +110,20 @@ func (r *WorkloadReconciler) newSlice(wl *kueue.Workload) (*v1alpha1.Slice, erro
...
+				for _, domain := range psa.TopologyAssignment.Domains {
+					if slice.Spec.NodeSelector == nil {
+						slice.Spec.NodeSelector = make(map[string][]string)
+					}
+					if slice.Spec.NodeSelector[TPUReservationSubblockLabel] == nil {
+						slice.Spec.NodeSelector[TPUReservationSubblockLabel] = []string{}
+					}
```
PBundyra, line 120: Can we move this logic outside of the loop?

slice/internal/controller/workload_controller_test.go
PBundyra, line 1: Can we add a UT for a case where TopologyAssignment arrives after some delay?

mbobrovskyi: We have this test:

https://github.com/AI-Hypercomputer/xpk/blob/3975e2814722e939ecb987bdaa9e646a1bee4872/slice/internal/controller/workload_controller_test.go#L103-L153

slice/internal/controller/workload_controller.go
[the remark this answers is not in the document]
mbobrovskyi: Done

mbobrovskyi:
@PBundyra PTAL

PR #517, opened 2025-07-02
[slice] Remove Slice finalizer, once all Pods in the Workload have gracefully terminated.

mbobrovskyi:
## Fixes / Features
- Remove Slice finalizer, once all Pods in the Workload have gracefully terminated.

PR #520, opened 2025-07-07
[slice] Create Slice for each PodSet in Workload.

mbobrovskyi:
## Fixes / Features
- Create Slice for each PodSet in Workload.

### DIFF
--- a/slice/internal/controller/workload_controller.go
+++ b/slice/internal/controller/workload_controller.go
@@ -20,9 +20,10 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"sort"
+	"strings"
 	"time"
 
-	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
@@ -55,7 +56,7 @@ const (
 	SliceControllerName         = "accelerator.gke.io/slice"
 	TPUReservationSubblockLabel = "cloud.google.com/gke-tpu-reservation-subblock"
 
-	SliceCreatedEventType          = "SliceCreated"
+	SlicesCreatedEventType         = "SlicesCreated"
 	FailedCreateSliceEventType     = "FailedCreateSlice"
 	AdmissionCheckUpdatedEventType = "AdmissionCheckUpdated"
 )
@@ -106,8 +107,8 @@ func (r *WorkloadReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 
 	if finalize, reason := shouldFinalize(wl); finalize {
 		if controllerutil.ContainsFinalizer(wl, SliceControllerName) {
-			log.V(3).Info(fmt.Sprintf("Cleaning up the Slice and finalize the Workload because %s", reason))
-			cleanedUp, err := r.cleanupSlice(ctx, wl)
+			log.V(3).Info(fmt.Sprintf("Cleaning up the Slices and finalizing the Workload because %s", reason))
+			cleanedUp, err := r.cleanupSlices(ctx, wl)
 			if err != nil {
 				return ctrl.Result{}, err
 			}
@@ -148,19 +149,27 @@ func (r *WorkloadReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 		return ctrl.Result{}, nil
 	}
 
-	slice := v1alpha1.Slice{}
-	err = r.client.Get(ctx, core.SliceKeyFromWorkload(wl), &slice)
-	if apierrors.IsNotFound(err) {
-		// slice not found, create it and exit.
-		err = r.createSlice(ctx, log, wl, ac)
-		return ctrl.Result{}, err
-	} else if err != nil {
-		// error fetching slice
-		log.Error(err, "Failed to fetch the Slice")
+	slices, err := r.findWorkloadSlices(ctx, wl)
+	if err != nil {
+		log.Error(err, "Failed to list Slices")
 		return ctrl.Result{}, err
 	}
 
-	err = r.syncAdmissionCheckStatus(ctx, wl, ac, &slice)
+	deleted, _, _ := r.groupSlices(slices)
+	if len(deleted) > 0 {
+		log.V(3).Info(
+			"Waiting for deleted Slices to be cleaned up; skipping reconciliation for now",
+			"deletedSlices", klog.KObjSlice(deleted),
+		)
+		return ctrl.Result{}, err
+	}
+
+	changed, err := r.syncSlices(ctx, wl, ac, slices)
+	if err != nil || changed {
+		return ctrl.Result{}, err
+	}
+
+	err = r.syncAdmissionCheckStatus(ctx, wl, ac, slices)
 	return ctrl.Result{}, client.IgnoreNotFound(err)
 }
 
@@ -204,48 +213,98 @@ func isJobSetOwner(wl *kueue.Workload) bool {
 	return false
 }
 
-func (r *WorkloadReconciler) cleanupSlice(ctx context.Context, wl *kueue.Workload) (bool, error) {
-	slice := v1alpha1.Slice{}
-	sliceKey := core.SliceKeyFromWorkload(wl)
+func (r *WorkloadReconciler) cleanupSlices(ctx context.Context, wl *kueue.Workload) (bool, error) {
+	log := ctrl.LoggerFrom(ctx)
 
-	log := ctrl.LoggerFrom(ctx).WithValues("slice", klog.KRef(sliceKey.Namespace, sliceKey.Name))
-	ctrl.LoggerInto(ctx, log)
-
-	err := r.client.Get(ctx, sliceKey, &slice)
-	if apierrors.IsNotFound(err) {
-		// slice not found
-		return true, nil
-	} else if err != nil {
-		// error fetching slice
-		log.Error(err, "Failed to fetch the Slice")
+	slices, err := r.findWorkloadSlices(ctx, wl)
+	if err != nil {
+		log.Error(err, "Failed to fetch Slices")
 		return false, err
 	}
 
-	if !slice.DeletionTimestamp.IsZero() {
-		log.V(3).Info("Slice already deleted, finishing cleanup")
+	deleted, deformed, other := r.groupSlices(slices)
+
+	if len(deleted) == len(slices) {
+		log.V(3).Info("All slices already deleted; finishing cleanup")
 		return true, nil
 	}
 
-	if !core.Deformed(&slice) {
+	if len(deformed) > 0 {
+		log.V(3).Info("Found Slices in deformed state; cleaning them up", "deformedSlices", klog.KObjSlice(deformed))
+		// We still need to delete deformed Slices because requeueing causes a conflict error during Slice creation.
+		err = r.deleteSlices(ctx, deformed)
+		if err != nil {
+			return false, err
+		}
+	}
+
+	if len(other) > 0 {
 		terminated, err := r.ownerPodsFinished(ctx, wl)
 		if err != nil || !terminated {
 			return false, err
 		}
-	} else {
-		log.V(3).Info("Slice in deformed state")
-		// We still need to delete the Slice because requeueing causes a conflict error during Slice creation.
 	}
 
-	log.V(3).Info("Deleting the Slice")
-
-	err = r.client.Delete(ctx, &slice)
-	if apierrors.IsNotFound(err) {
-		return true, nil
-	} else if err != nil {
-		log.Error(err, "Failed to delete the Slice")
+	log.V(3).Info("Deleting Slices", "slices", klog.KObjSlice(other))
+	err = r.deleteSlices(ctx, other)
+	if err != nil {
+		return false, err
 	}
 
-	return true, err
+	return true, nil
+}
+
+func (r *WorkloadReconciler) findWorkloadSlices(ctx context.Context, wl *kueue.Workload) ([]v1alpha1.Slice, error) {
+	slices := &v1alpha1.SliceList{}
+	opts := []client.ListOption{
+		client.InNamespace(wl.Namespace),
+		client.MatchingFields{OwnerReferenceUID: string(wl.UID)},
+	}
+	if err := r.client.List(ctx, slices, opts...); err != nil {
+		return nil, err
+	}
+	return slices.Items, nil
+}
+
+// groupSlices categorizes a list of Slice objects into three groups based on their state.
+// It separates slices into deleted (marked for deletion), deformed (being torn down),
+// and other (active) slices.
+//
+// Parameters:
+//
+//	slices - A slice of v1alpha1.Slice objects to be categorized.
+//
+// Returns:
+//   - A slice containing deleted Slice objects (with non-zero DeletionTimestamp).
+//   - A slice containing deformed Slice objects (being torn down).
+//   - A slice containing other Slice objects (active/valid slices).
+func (r *WorkloadReconciler) groupSlices(slices []v1alpha1.Slice) ([]v1alpha1.Slice, []v1alpha1.Slice, []v1alpha1.Slice) {
+	var deleted, deformed, other []v1alpha1.Slice
+	for _, slice := range slices {
+		switch {
+		case !slice.DeletionTimestamp.IsZero():
+			deleted = append(deleted, slice)
+		case core.Deformed(&slice):
+			deformed = append(deformed, slice)
+		default:
+			other = append(other, slice)
+		}
+	}
+	return deleted, deformed, other
+}
+
+func (r *WorkloadReconciler) deleteSlices(ctx context.Context, slices []v1alpha1.Slice) error {
+	log := ctrl.LoggerFrom(ctx)
+	for _, slice := range slices {
+		log = log.WithValues("slice", klog.KObj(&slice))
+		log.V(3).Info("Deleting the Slice")
+		err := r.client.Delete(ctx, &slice)
+		if client.IgnoreNotFound(err) != nil {
+			log.Error(err, "Failed to delete the Slice")
+			return err
+		}
+	}
+	return nil
 }
 
 func (r *WorkloadReconciler) ownerPodsFinished(ctx context.Context, wl *kueue.Workload) (bool, error) {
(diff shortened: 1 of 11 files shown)


slice/internal/controller/workload_controller.go
```diff
@@ -85,75 +91,212 @@ func (r *WorkloadReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
...
 
-	return r.client.Create(ctx, slice)
+	psa := findPodSetAssignment(wl, podSetName)
+	if psa == nil {
+		return nil, errPodSetAssignmentNotFound
+	}
+
+	if psa.TopologyAssignment == nil {
+		return nil, errTopologyAssignmentNotFound
+	}
```
IrvingMg, line 234:
```suggestion
	psa := findPodSetAssignment(wl, podSetName)
	if psa == nil || psa.TopologyAssignment == nil {
		return nil, errPodSetAssignmentNotFound
	}
```

Maybe like this?

slice/internal/controller/workload_controller.go
```diff
@@ -85,75 +91,212 @@ func (r *WorkloadReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
...
+			NodeSelector:        make(map[string][]string),
+		},
+	}
+
+	for _, domain := range psa.TopologyAssignment.Domains {
+		if ps.Template.Spec.NodeSelector[TPUReservationSubBlockLabel] != "" {
+			subBlockDomains := sets.New[string]()
+			for _, v := range domain.Values {
+				subBlockDomains.Insert(v)
+			}
```
IrvingMg, line 253:
```suggestion
subBlockDomains.Insert(domain.Values...)
```

Can we do this?

slice/internal/controller/workload_controller.go
```diff
@@ -85,75 +91,212 @@ func (r *WorkloadReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
...
+			if subBlockDomains.Len() > 0 {
+				slice.Spec.NodeSelector[TPUReservationSubBlockLabel] = sets.List(subBlockDomains)
+			}
+		}
+		if ps.Template.Spec.NodeSelector[NodePoolLabel] != "" {
+			nodePoolDomains := sets.New[string]()
+			for _, v := range domain.Values {
+				nodePoolDomains.Insert(v)
+			}
```
IrvingMg, line 262:
```suggestion
nodePoolDomains.Insert(domain.Values...)
```
Same than above

slice/internal/controller/workload_controller.go
```diff
@@ -85,75 +91,212 @@ func (r *WorkloadReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
...
 	}
 
-	return slice, nil
+	for _, slice := range slicesToCreate {
+		err = r.client.Create(ctx, slice)
+		if err != nil {
+			log.Error(err, "Failed to create a Slice", "slice", klog.KObj(slice))
+			return err
+		}
+	}
```
IrvingMg, line 196: Not sure if there's an edge case where creating many slices causes a performance drop, but if so, maybe we could create them concurrently instead of sequentially? WDYT?

slice/internal/controller/workload_controller.go
mbobrovskyi: But here is different errors.

IrvingMg: Oh, sorry... I didn't see they're different errors

slice/internal/controller/workload_controller.go
```diff
@@ -85,75 +93,217 @@ func (r *WorkloadReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
...
 
-	if err := controllerutil.SetControllerReference(wl, slice, r.client.Scheme()); err != nil {
+	for _, slice := range slices {
+		err = r.client.Delete(ctx, &slice)
+		if client.IgnoreNotFound(err) != nil {
+			log.Error(err, "Failed to delete the Slice", "slice", klog.KObj(&slice))
+			return err
+		}
+	}
```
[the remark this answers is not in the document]
mbobrovskyi: Good catch! Thank you!

IrvingMg:
LGTM.

cc: @PBundyra @gabesaba

mbobrovskyi:
@gabesaba PTAL

gabesaba:
> @gabesaba PTAL

Let's put this PR on hold until the other two (#521, #517) merge. It will be more overhead given that they're also modifying workload controller, and the behavior of those two (especially #521) is more fundamental

mbobrovskyi:
> > @gabesaba PTAL
> 
> Let's put this PR on hold until the other two (#521, #517) merge. It will be more overhead given that they're also modifying workload controller, and the behavior of those two (especially #521) is more fundamental

https://github.com/AI-Hypercomputer/xpk/pull/521 is related for this changes :(

slice/internal/controller/workload_controller.go
```diff
@@ -85,75 +93,222 @@ func (r *WorkloadReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
...
+func GetSliceName(workloadName string, podSetName kueue.PodSetReference) string {
+	return fmt.Sprintf("%s-%s", workloadName, podSetName)
+}
 
-	err := r.client.Get(ctx, client.ObjectKeyFromObject(slice), slice)
-	if client.IgnoreNotFound(err) != nil {
-		return err
+func newSlice(wl *kueue.Workload, podSetName kueue.PodSetReference) (*v1alpha1.Slice, error) {
+	ps := findPodSet(wl, podSetName)
+	if findPodSet(wl, podSetName) == nil {
```
pajakd, line 233:
```suggestion
	if ps == nil {
```

slice/internal/controller/workload_controller.go
```diff
@@ -85,75 +93,222 @@ func (r *WorkloadReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
...
+	if err := r.client.List(ctx, slices, opts...); err != nil {
 		return nil, err
 	}
+	return slices.Items, nil
+}
+
+func (r *WorkloadReconciler) createSlicesIfNotExist(ctx context.Context, wl *kueue.Workload) error {
+	log := ctrl.LoggerFrom(ctx)
+
+	createdSlices, err := r.findWorkloadSlices(ctx, wl)
```
pajakd, line 157: nit: consider renaming to sth like `existingSlices`

slice/internal/controller/workload_controller.go

slice/internal/controller/workload_controller.go

slice/internal/controller/workload_controller.go
```diff
@@ -280,27 +379,63 @@ func (r *WorkloadReconciler) updateWorkloadAdmissionCheckStatus(ctx context.Cont
...
+
+	slicesByStatus := make(map[v1alpha1.SliceConditionType][]v1alpha1.Slice)
+	for _, slice := range slices {
+		if core.Forming(&slice) {
+			slicesByStatus[v1alpha1.Forming] = append(slicesByStatus[v1alpha1.Forming], slice)
+		}
+
+		if core.Ready(&slice) {
+			slicesByStatus[v1alpha1.Ready] = append(slicesByStatus[v1alpha1.Ready], slice)
+		}
```
pajakd, line 403:
```suggestion
		for _, status := range []v1alpha1.SliceConditionType{v1alpha1.Forming, v1alpha1.Ready, v1alpha1.Degraded, v1alpha1.Deformed, v1alpha1.Error} {
			if meta.IsStatusConditionTrue(slice.Status.Conditions, string(status)) {
				slicesByStatus[status] = append(slicesByStatus[status], slice)
			}
		}
```
Not sure if better but how about making a loop over all the possible states?

slice/internal/controller/workload_controller.go
```diff
@@ -280,27 +379,63 @@ func (r *WorkloadReconciler) updateWorkloadAdmissionCheckStatus(ctx context.Cont
...
-	case meta.IsStatusConditionTrue(slice.Status.Conditions, string(v1alpha1.Deformed)):
-		ac.State = kueue.CheckStateRejected
-		ac.Message = fmt.Sprintf("The Slice %q is being torn down", slice.Name)
-	case errCond != nil && errCond.Status == metav1.ConditionTrue:
+		ac.Message = fmt.Sprintf("The Slices %s are running with reduced capacity or performance", joinSliceNames(slicesByStatus[v1alpha1.Degraded]))
+	case len(slicesByStatus[v1alpha1.Deformed]) > 0:
 		ac.State = kueue.CheckStateRejected
-		ac.Message = fmt.Sprintf("The Slice %q is not operational due to an error: %s", slice.Name, errCond.Message)
+		ac.Message = fmt.Sprintf("The Slices %s are being torn down", joinSliceNames(slicesByStatus[v1alpha1.Deformed]))
+	}
```
pajakd, line 434: This is tricky if we can have multiple slices in different states now.

For example if one is `Ready` and another one is `Forming` we will return admission state `Ready` but I would assume that the admission check for the workload should transition to `Ready` only after all the slices are Ready, right?

I think the order of the `case` statements should be `Error`, `Deformed`, `Forming`, `Degraded`, `Ready`

slice/internal/controller/workload_controller.go
```diff
@@ -100,22 +100,19 @@ func (r *WorkloadReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
...
+
+		log.V(3).Info(fmt.Sprintf("Cleaning up the Slice and finalize the Workload because %s", reason))
+
+		err := r.cleanupSlices(ctx, wl)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+
+		err = r.finalizeWorkload(ctx, wl)
+		return ctrl.Result{}, client.IgnoreNotFound(err)
```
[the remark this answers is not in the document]
mbobrovskyi: Why this changes?

slice/internal/controller/workload_controller.go
```diff
@@ -280,27 +379,63 @@ func (r *WorkloadReconciler) updateWorkloadAdmissionCheckStatus(ctx context.Cont
...
 		ac.State = kueue.CheckStateReady
-		ac.Message = fmt.Sprintf("The Slice %q is running with reduced capacity or performance", slice.Name)
-	case meta.IsStatusConditionTrue(slice.Status.Conditions, string(v1alpha1.Deformed)):
-		ac.State = kueue.CheckStateRejected
-		ac.Message = fmt.Sprintf("The Slice %q is being torn down", slice.Name)
-	case errCond != nil && errCond.Status == metav1.ConditionTrue:
+		ac.Message = fmt.Sprintf("The Slices %s are running with reduced capacity or performance", joinSliceNames(slicesByStatus[v1alpha1.Degraded]))
+	case len(slicesByStatus[v1alpha1.Deformed]) > 0:
 		ac.State = kueue.CheckStateRejected
-		ac.Message = fmt.Sprintf("The Slice %q is not operational due to an error: %s", slice.Name, errCond.Message)
```
mbobrovskyi, line 303: I'm also not sure about this message. Previously, we used the error message from the condition, but now there could be multiple errors. We could join them together, but that might result in a very long message.

@pajakd @gabesaba WDYT?

slice/internal/controller/workload_controller.go
mbobrovskyi: Done.

slice/internal/controller/workload_controller.go

slice/internal/controller/workload_controller.go
pajakd: The problem is that we should aggregate info about all the slices in a single message (currently we only put there info about a single type).  

Would it be useful to have a message like "Slices are in states: 3 Forming, 5 Ready, 1 Error" (to eg. see the progress over time)? The error messages are something that would be useful to append to it, but if the message would get too long I think we can skip them (the errors should be logged somewhere anyway).

gabesaba:
> > @gabesaba PTAL
> 
> Let's put this PR on hold until the other two (#521, #517) merge. It will be more overhead given that they're also modifying workload controller, and the behavior of those two (especially #521) is more fundamental

/hold for 517

slice/internal/controller/workload_controller.go
```diff
@@ -204,48 +213,98 @@ func isJobSetOwner(wl *kueue.Workload) bool {
...
 	}
 
-	if !core.Deformed(&slice) {
+	if len(deformed) > 0 {
+		log.V(3).Info("Found Slices in deformed state; cleaning them up", "deformedSlices", klog.KObjSlice(deformed))
+		// We still need to delete deformed Slices because requeueing causes a conflict error during Slice creation.
+		err = r.deleteSlices(ctx, deformed)
+		if err != nil {
+			return false, err
+		}
```
pajakd, line 238: Is the fact that we failed to delete the `deformed` slices blocking us for some reason from trying to delete the `other` slices?

slice/internal/controller/workload_controller.go
```diff
@@ -204,48 +213,98 @@ func isJobSetOwner(wl *kueue.Workload) bool {
...
-	return true, err
+func (r *WorkloadReconciler) deleteSlices(ctx context.Context, slices []v1alpha1.Slice) error {
+	log := ctrl.LoggerFrom(ctx)
+	for _, slice := range slices {
+		log = log.WithValues("slice", klog.KObj(&slice))
+		log.V(3).Info("Deleting the Slice")
+		err := r.client.Delete(ctx, &slice)
+		if client.IgnoreNotFound(err) != nil {
+			log.Error(err, "Failed to delete the Slice")
+			return err
```
pajakd, line 304: Similarly as above. In theory we could try to delete all the slices -- is there any reason why we return upon the first error?

slice/internal/controller/workload_controller.go
```diff
@@ -361,46 +420,76 @@ func (r *WorkloadReconciler) sliceAC(ctx context.Context, wl *kueue.Workload) (*
...
+		if err != nil {
+			return false, err
+		}
+
+		slices = append(slices, *createdSlice)
+		changed = true
+	}
+
+	if changed {
+		msg := fmt.Sprintf("The Slices %v have been created", joinSliceNames(slices))
```
pajakd, line 449: Here `slices` contains both the preexisting slices and the newly created ones. I guess we should log only the names of the ones we create in this call.

slice/internal/controller/workload_controller.go
pajakd: OK

mbobrovskyi: >  is there any reason why we return upon the first error?

Any factors, such as a network issue.

slice/internal/controller/workload_controller.go
```diff
@@ -361,46 +420,77 @@ func (r *WorkloadReconciler) sliceAC(ctx context.Context, wl *kueue.Workload) (*
...
+	}
+
+	if changed {
+		msg := buildACStatusMessageForCreation(slices)
+		ctrl.LoggerFrom(ctx).V(3).Info(msg)
+		r.record.Event(wl, corev1.EventTypeNormal, SlicesCreatedEventType, msg)
+		ac.State = kueue.CheckStatePending
+		ac.Message = msg
+		return true, r.updateWorkloadAdmissionCheckStatus(ctx, wl, ac)
+	}
```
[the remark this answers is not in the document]

slice/internal/controller/workload_controller.go
```diff
@@ -413,36 +503,98 @@ func (r *WorkloadReconciler) updateWorkloadAdmissionCheckStatus(ctx context.Cont
...
+
+func buildACStatusMessageForSync(slices map[v1alpha1.SliceConditionType][]v1alpha1.Slice) string {
+	msg := fmt.Sprintf(
+		"Slices are in states: %d Forming, %d Degraded, %d Ready, %d Error, %d Deformed",
+		len(slices[v1alpha1.Forming]),
+		len(slices[v1alpha1.Degraded]),
+		len(slices[v1alpha1.Ready]),
+		len(slices[v1alpha1.Error]),
+		len(slices[v1alpha1.Deformed]),
+	)
```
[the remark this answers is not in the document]

slice/internal/controller/workload_controller.go
```diff
@@ -413,36 +503,98 @@ func (r *WorkloadReconciler) updateWorkloadAdmissionCheckStatus(ctx context.Cont
...
+	}
+
+	return nil
+}
+
+func groupSlicesByStatus(slices []v1alpha1.Slice) map[v1alpha1.SliceConditionType][]v1alpha1.Slice {
+	slicesByStatus := make(map[v1alpha1.SliceConditionType][]v1alpha1.Slice)
+	for _, slice := range slices {
+		for _, status := range []v1alpha1.SliceConditionType{
+			v1alpha1.Error, v1alpha1.Deformed, v1alpha1.Forming, v1alpha1.Degraded, v1alpha1.Ready,
```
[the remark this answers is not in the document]

slice/internal/controller/workload_controller.go
```diff
@@ -413,36 +503,98 @@ func (r *WorkloadReconciler) updateWorkloadAdmissionCheckStatus(ctx context.Cont
...
-		ac.Message = fmt.Sprintf("The Slice %q is running with reduced capacity or performance", slice.Name)
-	case core.Deformed(slice):
-		ac.State = kueue.CheckStateRejected
-		ac.Message = fmt.Sprintf("The Slice %q is being torn down", slice.Name)
-	case errCond != nil && errCond.Status == metav1.ConditionTrue:
+	case len(slicesByStatus[v1alpha1.Error]) > 0 || len(slicesByStatus[v1alpha1.Deformed]) > 0:
 		ac.State = kueue.CheckStateRejected
-		ac.Message = fmt.Sprintf("The Slice %q is not operational due to an error: %s", slice.Name, errCond.Message)
+	case len(slices) == len(slicesByStatus[v1alpha1.Degraded])+len(slicesByStatus[v1alpha1.Ready]):
+		ac.State = kueue.CheckStateReady
```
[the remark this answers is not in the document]
mbobrovskyi: Added test case.

slice/internal/controller/workload_controller.go
```diff
@@ -361,46 +420,77 @@ func (r *WorkloadReconciler) sliceAC(ctx context.Context, wl *kueue.Workload) (*
...
-func parseTopologyAssignmentIntoNodeSelector(slice *v1alpha1.Slice, wl *kueue.Workload) {
-	nodeSelectors := sets.New[string]()
+func (r *WorkloadReconciler) syncSlices(ctx context.Context, wl *kueue.Workload, ac *kueue.AdmissionCheckState, slices []v1alpha1.Slice) (bool, error) {
+	slicesByName := make(map[string]*v1alpha1.Slice, len(slices))
+	for _, slice := range slices {
+		slicesByName[slice.Name] = &slice
+	}
+
+	changed := false
+
```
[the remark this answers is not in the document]
mbobrovskyi: This is outdated.

slice/internal/controller/workload_controller.go

slice/internal/controller/workload_controller.go
mbobrovskyi: I considered using the client.DeleteAllOf() function, but it seems we can't use it in this case — we need to delete only the Deformed slices, while still waiting for the termination of the other slices.

pajakd: Ok, lets leave it as is.

slice/internal/controller/workload_controller.go
```diff
@@ -413,36 +504,94 @@ func (r *WorkloadReconciler) updateWorkloadAdmissionCheckStatus(ctx context.Cont
...
+	slicesByStatus := groupSlicesByStatus(slices)
+
+	switch {
+	case len(slicesByStatus[v1alpha1.Error]) > 0 || len(slicesByStatus[v1alpha1.Deformed]) > 0:
+		ac.State = kueue.CheckStateRejected
+	case len(slices) == len(slicesByStatus[v1alpha1.Degraded])+len(slicesByStatus[v1alpha1.Ready]):
+		ac.State = kueue.CheckStateReady
+	}
+
+	ac.Message = fmt.Sprintf("Slices are in states: %d Created", len(slices))
```
gabesaba, line 580: are we double counting? so it could be 3 Created, 3 Ready, when in total there are just 3?

slice/internal/controller/workload_controller.go
```diff
@@ -361,46 +420,78 @@ func (r *WorkloadReconciler) sliceAC(ctx context.Context, wl *kueue.Workload) (*
...
+		if err != nil {
+			return err
 		}
+
+		*slices = append(*slices, *createdSlice)
+		createdSlices = append(createdSlices, *createdSlice)
+	}
+
+	if len(createdSlices) > 0 {
+		msg := buildCreationEventMessage(*slices)
```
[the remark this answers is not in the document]

slice/internal/controller/workload_controller.go
mbobrovskyi: Fixed.

slice/internal/controller/workload_controller.go
```diff
@@ -361,46 +420,78 @@ func (r *WorkloadReconciler) sliceAC(ctx context.Context, wl *kueue.Workload) (*
...
+	}
+	createdSlices := make([]v1alpha1.Slice, 0, capacity)
+
 	for _, psa := range wl.Status.Admission.PodSetAssignments {
-		// we already validated that all assignments have a valid level,
-		// in validateRelevantWorkload.
-		subblockLevelIndex := topology.SubblockLevelIndex(&psa)
-		for _, domain := range psa.TopologyAssignment.Domains {
-			nodeSelectors.Insert(domain.Values[subblockLevelIndex])
+		sliceName := core.SliceName(wl.Name, psa.Name)
```
gabesaba, line 436: In follow-up PR:

Should we use `IsRelevantPodTemplateSpec` in this loop? Otherwise, we may try to create a slice for an invalid podset. Can you add a test for this to please?

gabesaba:
lgtm, thank you! Let's address the last comments in a follow-up PR.

PR #532, opened 2025-07-16
Add tpu resouce flavor to kueue

wstcliyu:
## Fixes / Features
- Add tpu resouce flavor to kueue config. AxLearn workloads request cpu and memory as well as tpu. So xpk needs to add cpu and memory to tpu flavor for AxLearn workload to be accepted by kueue.

## Testing / Documentation
Testing details.

- [ y ] Tests pass


SujeethJinesh:
This looks good to me, but I wanted to check one thing, would you be able to confirm that a CPU pod doesn't land on a TPU nodepool? May need to add tolerations for this in XPK, but not sure. You can test this by creating a simple cluster (or using an existing cluster) that does not have any CPU nodepools, and launch a cpu only jobset on it and confirm that it's unschedulable.

src/xpk/core/pathways.py
```diff
@@ -39,16 +39,16 @@ def add_pw_resource_flavors(args):
...
-    flavors:
-    - name: cpu-user
-      resources:
-      - name: "cpu"
-        nominalQuota: 480
-      - name: "memory"
-        nominalQuota: 2000G"""
-  if args.enable_pathways:
-    return resources_yaml
+  # resources_yaml = """- coveredResources: ["cpu", "memory"]
```
SujeethJinesh, line 42: I'd inline this function into the above.

PR #533, opened 2025-07-16
Parse cloud.google.com/gke-tpu-topology annotation for podSet size.

mbobrovskyi:
## Fixes / Features
- Parse cloud.google.com/gke-tpu-topology annotation for podSet size.


slice/internal/webhooks/jobset_webhook.go
```diff
@@ -51,26 +53,57 @@ var _ webhook.CustomDefaulter = &JobSetWebhook{}
...
+
+func (r *JobSetWebhook) podSetSliceSize(parallelism *int32, tpuTopology string) (int32, error) {
+	subBlockCount, err := r.subBlockCount(tpuTopology)
+	if err != nil {
+		return 0, err
+	}
+	return ptr.Deref(parallelism, 1) / subBlockCount, nil
+}
+
+func (r *JobSetWebhook) subBlockCount(tpuTopology string) (int32, error) {
```
mszadkow, line 99: nit: not sure there is a need for this to be a method

slice/internal/webhooks/jobset_webhook.go
```diff
@@ -51,26 +53,57 @@ var _ webhook.CustomDefaulter = &JobSetWebhook{}
...
+	subBlockCount := int32(1)
+	for _, part := range strings.Split(tpuTopology, "x") {
+		partInt, err := strconv.ParseInt(part, 10, 32)
+		if err != nil {
+			return 0, fmt.Errorf("invalid tpu topology %q: %w", tpuTopology, err)
+		}
+		subBlockCount *= int32(partInt)
+	}
+	return subBlockCount / 64, nil
```
mszadkow, line 108: each subBlockCount below 64 and this will return 0, right?

mbobrovskyi: @gabesaba WDYT?

gabesaba: We can assume that the smallest topology will be 4x4x4, so this will always be at least 64.

slice/internal/webhooks/jobset_webhook.go
```diff
@@ -55,22 +61,55 @@ func (r *JobSetWebhook) Default(ctx context.Context, obj runtime.Object) error {
...
 			rj.Template.Annotations = make(map[string]string)
 		}
-		rj.Template.Annotations[PodSetRequiredTopologyAnnotation] = AnnotationValueTBD
-		rj.Template.Annotations[PodSetSliceRequiredTopologyAnnotation] = AnnotationValueTBD
-		rj.Template.Annotations[PodSetSliceSizeAnnotation] = AnnotationValueTBD
+
+		rj.Template.Annotations[kueuealpha.PodSetRequiredTopologyAnnotation] = rj.Template.Spec.Template.Annotations[TPUBlockAnnotation]
+		rj.Template.Annotations[kueuealpha.PodSetSliceRequiredTopologyAnnotation] = rj.Template.Spec.Template.Annotations[TPUSubBlockAnnotation]
+
+		podSetSliceSize, err := podSetSliceSize(tpuTopology, rj.Template.Spec.Parallelism)
```
gabesaba, line 83: Can we validate the Parallelism before calling podSetSliceSize, so we don't have to pass a potential nil?

slice/internal/webhooks/jobset_webhook.go
```diff
@@ -55,22 +61,55 @@ func (r *JobSetWebhook) Default(ctx context.Context, obj runtime.Object) error {
...
 
-	// handle webhooks for Kueue related JobSets
-	cq, ok := jobSet.Labels[kueueconstants.QueueLabel]
-	if !ok || cq == "" {
+	if jobSet.Labels[kueueconstants.QueueLabel] == "" {
 		return nil
 	}
 
 	for i, rj := range jobSet.Spec.ReplicatedJobs {
+		tpuTopology := rj.Template.Spec.Template.Annotations[TPUTopologyAnnotation]
```
gabesaba, line 69: Pull out the inside of this for-loop to a function? something like `annotateReplicatedJobWithTopology`

slice/internal/webhooks/jobset_webhook.go
```diff
@@ -55,22 +61,55 @@ func (r *JobSetWebhook) Default(ctx context.Context, obj runtime.Object) error {
...
 
 		jobSet.Spec.ReplicatedJobs[i] = rj
 	}
 
 	return nil
 }
+
+func podSetSliceSize(tpuTopology string, parallelism *int32) (int32, error) {
+	dimensions := strings.Split(tpuTopology, "x")
+	if len(dimensions) < 2 || len(dimensions) > 3 {
```
gabesaba, line 97: len() != 3?

slice/internal/webhooks/jobset_webhook.go
```diff
@@ -55,22 +61,55 @@ func (r *JobSetWebhook) Default(ctx context.Context, obj runtime.Object) error {
...
+		return 0, fmt.Errorf("%w: invalid dimension count in %q", errInvalidTPUTopologyAnnotation, tpuTopology)
+	}
+
+	subBlockCount := int32(1)
```
gabesaba, line 101: subblock count is what we get after dividing by 64.

slice/internal/webhooks/jobset_webhook.go
```diff
@@ -55,22 +61,55 @@ func (r *JobSetWebhook) Default(ctx context.Context, obj runtime.Object) error {
...
 		return nil
 	}
 
 	for i, rj := range jobSet.Spec.ReplicatedJobs {
+		tpuTopology := rj.Template.Spec.Template.Annotations[TPUTopologyAnnotation]
+		tpuAccelerator := rj.Template.Spec.Template.Spec.NodeSelector[TPUAcceleratorLabel]
+
+		if tpuTopology == "" || tpuAccelerator == "" {
+			continue
+		}
```
gabesaba, line 74: may consider using regex to simplify the validation (then you can delete other validation logic later, and just cast the string to int). we should also skip if it is not `tpu-v7x`
```suggestion
                validTopology, _ := MatchString("[0-9]+x[0-9]+x[0-9]+", tpuTopology)
		if  !validTopology || tpuAccelerator != "tpu-v7x" {
			continue
		}
```

slice/internal/webhooks/jobset_webhook_test.go
```diff
@@ -21,108 +21,242 @@ import (
...
-			},
+		"no queue label": {
+			jobSet: testingjobjobset.MakeJobSet(baseJobSetName, utils.DefaultNamespace).
+				ReplicatedJobs(testingjobjobset.ReplicatedJobRequirements{
+					Name:        "rj1",
+					Parallelism: 12,
+					PodAnnotations: map[string]string{
+						TPUTopologyAnnotation: "4x4x12",
+						TPUBlockAnnotation:    "cloud.google.com/topology-rack",
+						TPUSubBlockAnnotation: "cloud.google.com/topology-host",
```
gabesaba, line 50: for the test cases, please use literals instead of variables (here and elsewhere). This makes it easier to see that the test is doing the right thing, without having to scroll, reference the variable

gabesaba: also, why are these annotations on the pods? I would only expect TPUTopologyAnnotation and in NodeSelector, TPUAcceleratorLabel

slice/internal/webhooks/jobset_webhook.go
mbobrovskyi: I think it is possible to set 2D dimension (2x2), no?

slice/internal/webhooks/jobset_webhook_test.go
mbobrovskyi: > also, why are these annotations on the pods? I would only expect TPUTopologyAnnotation and in NodeSelector, TPUAcceleratorLabel

In result i need to have 3 labels:

```
kueuealpha.PodSetRequiredTopologyAnnotation:      "cloud.google.com/topology-rack",
kueuealpha.PodSetSliceRequiredTopologyAnnotation: "cloud.google.com/topology-host",
kueuealpha.PodSetSliceSizeAnnotation:             "4",
```

IIUC we can get it from:

```
cloud.google.com/gke-tpu-block:    "cloud.google.com/topology-rack",
cloud.google.com/gke-tpu-subblock: "cloud.google.com/topology-host",
```

slice/internal/webhooks/jobset_webhook.go
mbobrovskyi: Done

slice/internal/webhooks/jobset_webhook.go
```diff
@@ -55,22 +58,58 @@ func (r *JobSetWebhook) Default(ctx context.Context, obj runtime.Object) error {
...
+		return nil
+	}
 
-		jobSet.Spec.ReplicatedJobs[i] = rj
+	if rj.Template.Annotations == nil {
+		rj.Template.Annotations = make(map[string]string)
 	}
 
+	rj.Template.Annotations[kueuealpha.PodSetRequiredTopologyAnnotation] = rj.Template.Spec.Template.Annotations[TPUBlockAnnotation]
+	rj.Template.Annotations[kueuealpha.PodSetSliceRequiredTopologyAnnotation] = rj.Template.Spec.Template.Annotations[TPUSubBlockAnnotation]
```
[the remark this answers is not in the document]
gabesaba: Also, I think that the left-hand side should be Template.Spec.Template.ObjectMeta.Annotations

See https://github.com/kubernetes-sigs/kueue/blob/b656089c0e4e91e5af834da1f9b572971bf348e5/pkg/controller/jobs/jobset/jobset_controller.go#L124, and how it is used https://github.com/kubernetes-sigs/kueue/blob/b656089c0e4e91e5af834da1f9b572971bf348e5/pkg/controller/jobframework/tas.go#L55-L60

slice/test/e2e/jobset_test.go
gabesaba, line 1: Can we add a test which demonstrates this new behavior working E2E with Kueue - that the Topology assignment would differ after we set `PodSetSliceRequiredTopologyAnnotation` to a real value (rather than TBD). Something that would also catch the Annotation being on Jobset, vs template.spec.annotations (one of my comments above)

slice/internal/webhooks/jobset_webhook_test.go
gabesaba: We want something like this:
```
kueue.x-k8s.io/podset-required-topology:  cloud.google.com/gke-tpu-block
kueue.x-k8s.io/podset-slice-required-topology: cloud.google.com/gke-tpu-subblock
kueue.x-k8s.io/podset-slice-size: 16
```

slice/test/e2e/jobset_test.go
gabesaba: An easier starting point would be to validate topology request

https://github.com/kubernetes-sigs/kueue/blob/b656089c0e4e91e5af834da1f9b572971bf348e5/pkg/controller/jobs/jobset/jobset_controller.go#L131

slice/internal/webhooks/jobset_webhook_test.go
mbobrovskyi: Yes, exactly, that what we have.

mbobrovskyi: Done

slice/internal/webhooks/jobset_webhook.go
```diff
@@ -55,22 +56,58 @@ func (r *JobSetWebhook) Default(ctx context.Context, obj runtime.Object) error {
...
+		return err
+	}
+	rj.Template.Spec.Template.Annotations[kueuealpha.PodSetSliceSizeAnnotation] = fmt.Sprint(size)
+
 	return nil
 }
+
+func (r *JobSetWebhook) podSetSliceSize(tpuTopology string, parallelism int32) (int32, error) {
+	dimensions := strings.Split(tpuTopology, "x")
+	totalChips := int32(1)
```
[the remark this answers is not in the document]
mbobrovskyi: This conversion to int32 is necessary because parallelism is of type int32.

slice/internal/webhooks/jobset_webhook.go
```diff
@@ -55,22 +56,58 @@ func (r *JobSetWebhook) Default(ctx context.Context, obj runtime.Object) error {
...
+
+	for _, dim := range dimensions {
+		partInt, err := strconv.ParseInt(dim, 10, 32)
```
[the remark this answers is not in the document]
mbobrovskyi: Done

slice/internal/webhooks/jobset_webhook.go
```diff
@@ -55,22 +56,58 @@ func (r *JobSetWebhook) Default(ctx context.Context, obj runtime.Object) error {
...
+	size, err := r.podSetSliceSize(tpuTopology, ptr.Deref(rj.Template.Spec.Parallelism, 1))
+	if err != nil {
+		return err
+	}
+	rj.Template.Spec.Template.Annotations[kueuealpha.PodSetSliceSizeAnnotation] = fmt.Sprint(size)
+
 	return nil
 }
+
+func (r *JobSetWebhook) podSetSliceSize(tpuTopology string, parallelism int32) (int32, error) {
```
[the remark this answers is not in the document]
mbobrovskyi: Done

slice/internal/webhooks/jobset_webhook.go
```diff
@@ -55,22 +56,58 @@ func (r *JobSetWebhook) Default(ctx context.Context, obj runtime.Object) error {
...
+		return nil
+	}
 
-		jobSet.Spec.ReplicatedJobs[i] = rj
+	if rj.Template.Spec.Template.Annotations == nil {
+		rj.Template.Spec.Template.Annotations = make(map[string]string)
 	}
 
+	rj.Template.Spec.Template.Annotations[kueuealpha.PodSetRequiredTopologyAnnotation] = "cloud.google.com/gke-tpu-block"
+	rj.Template.Spec.Template.Annotations[kueuealpha.PodSetSliceRequiredTopologyAnnotation] = "cloud.google.com/gke-tpu-subblock"
```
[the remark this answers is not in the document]
mbobrovskyi: Done

slice/test/e2e/jobset_test.go
```diff
@@ -85,42 +98,52 @@ var _ = ginkgo.Describe("JobSet", func() {
...
 					},
 				).
-				RequestAndLimit("rj1", "cpu", "200m").
-				RequestAndLimit("rj2", "cpu", "200m").
+				RequestAndLimit("rj1", extraResource, "1").
+				RequestAndLimit("rj2", extraResource, "1").
 				Obj()
 
-			ginkgo.By("Creating the JobSet", func() {
+			ginkgo.By("Creating a JobSet", func() {
```
gabesaba, line 130: Can we update this test to span 2 sub-blocks? 

To do this: update the topology to 4x4x8
a jobset with single replicated job, single replica, with 128 parallelism, each requesting 1 tpu. we can validate that the subblock slice size comes out properly, to 64

slice/internal/webhooks/jobset_webhook_test.go
```diff
@@ -21,105 +21,193 @@ import (
...
+						kueuealpha.PodSetRequiredTopologyAnnotation:      TPUBlockLabel,
+						kueuealpha.PodSetSliceRequiredTopologyAnnotation: TPUSubBlockLabel,
+						kueuealpha.PodSetSliceSizeAnnotation:             "4",
+					},
+					NodeSelector: map[string]string{
+						TPUAcceleratorLabel: "tpu-v7x",
+					},
+				}).
+				Obj(),
+		},
```
gabesaba, line 141: Can we run the following cases? Perhaps a new structured test, so that these values are easy to specify
4x4x4, parallelism: 16 => 16 (expected size)
4x4x4, parallelism: 64 => 64
4x4x12, parallelism: 48 => 16
4x4x12, parallelism: 96 => 32

slice/internal/webhooks/jobset_webhook_test.go
```diff
@@ -21,105 +21,193 @@ import (
...
 	jobset "sigs.k8s.io/jobset/api/jobset/v1alpha2"
-	kueueconstants "sigs.k8s.io/kueue/pkg/controller/constants"
+	kueuealpha "sigs.k8s.io/kueue/apis/kueue/v1alpha1"
 
-	utiltesting "tpu-slice-controller/internal/util/testing"
+	testingjobjobset "tpu-slice-controller/internal/util/testingjobs/jobset"
+	"tpu-slice-controller/test/utils"
 )
 
 func TestDefault(t *testing.T) {
```
gabesaba, line 31: Is default right name of test anymore? This was just for when we were adding a placeholder, right?

gabesaba (commented):
looks good. just a few comments on tests

slice/internal/webhooks/jobset_webhook.go
```diff
@@ -55,22 +57,58 @@ func (r *JobSetWebhook) Default(ctx context.Context, obj runtime.Object) error {
...
+	return nil
+}
+
+func (r *JobSetWebhook) annotateReplicatedJobWithTopology(rj *v1alpha2.ReplicatedJob) error {
+	tpuTopology := rj.Template.Spec.Template.Annotations[TPUTopologyAnnotation]
+	tpuAccelerator := rj.Template.Spec.Template.Spec.NodeSelector[TPUAcceleratorLabel]
+
+	validTopology, _ := regexp.MatchString("[0-9]+x[0-9]+x[0-9]+", tpuTopology)
+	if !validTopology || tpuAccelerator != "tpu-v7x" {
+		return nil
```
sgysherry, line 80: Do we want to return an error here especially if the topology is invalid?

slice/internal/webhooks/jobset_webhook.go
```diff
@@ -55,22 +57,58 @@ func (r *JobSetWebhook) Default(ctx context.Context, obj runtime.Object) error {
...
+	}
 
-		jobSet.Spec.ReplicatedJobs[i] = rj
+	if rj.Template.Spec.Template.Annotations == nil {
+		rj.Template.Spec.Template.Annotations = make(map[string]string)
 	}
 
+	rj.Template.Spec.Template.Annotations[kueuealpha.PodSetRequiredTopologyAnnotation] = TPUBlockLabel
```
sgysherry, line 87: `rj.Template.Spec.Template.Annotations` or `rj.Template.Spec.Template.Spec.Annotations`?

gabesaba:
lgtm. Thank you :)

PR #557, opened 2025-07-25
Use kueue manifest from CT url

pawloch00:
## Fixes / Features
-
-

## Testing / Documentation
Testing details.

- [ y/n ] Tests pass


pawloch00:
@SujeethJinesh we are removing string with patch to kueue manager config. If you want to change limits in kueue config please come up with another way to do this. The way it is now is error prone and creates two sources of same kueue config.

sharabiani (approved):

PR #571, opened 2025-07-30
Update CPU limit for medium to large scale clusters

SujeethJinesh:
## Fixes / Features
- Update the CPU limit for clusters with medium to large number of nodes. See https://screenshot.googleplex.com/3XQpts9RicMnFHQ for details on the reason we saw the crashing on jobset manager.
- Ideally we should also scale this CPU to the number of nodes rather than hardcoding it.

## Testing / Documentation
Testing details. Updated manually on scaling cluster

- [ y ] Tests pass


RoshaniN (approved):
Thanks!

SidneyShen (approved):

PR #573, opened 2025-08-01
[Slice] Create Slices only for relevant PodSets.

mbobrovskyi:
## Fixes / Features
- Create Slices only for relevant PodSets.


slice/internal/controller/workload_controller.go
```diff
@@ -433,6 +434,11 @@ func (r *WorkloadReconciler) syncSlices(ctx context.Context, wl *kueue.Workload,
 	createdSlices := make([]v1alpha1.Slice, 0, capacity)
 
 	for _, psa := range wl.Status.Admission.PodSetAssignments {
+		podSet := podset.FindPodSetByName(wl.Spec.PodSets, psa.Name)
+		if !core.IsRelevantPodTemplateSpec(podSet.Template) {
```
gabesaba, line 438: I think that we miss the case where the topology assignment is invalid (missing subblock level)

Can we define new function, `shouldCreateSliceForPodsetAssignment`
Then, in there we can check:
* IsRelevantPodTemplateSpec
* IsAssignmentValid(PodSetAssignment) (new function, in topology/validation.go)?

gabesaba (commented):
lgtm, thank you!

PR #574, opened 2025-08-03
feat: Pathways use single resource group

samos123:
Jobs requesting TPU resources may also have requests for CPU and memory. However when pathways is enabled, Kueue will not be able to admit such jobs since there is no cpu and memory quota.

This fix adds a very high number of CPU and memory for TPU/GPU resources and merges the pathways resource group with the accelerator resource group.

This also allows us to run AXLearn jobs without having to make changes manually.

## Testing / Documentation
Testing details.

- [ y/n ] Tests pass

### DIFF
--- a/src/xpk/core/kueue.py
+++ b/src/xpk/core/kueue.py
@@ -28,7 +28,7 @@ from .commands import (
     run_command_with_updates,
     run_command_with_updates_retry,
 )
-from .pathways import add_pw_resource_flavors, add_pw_resources_to_kueue
+from .pathways import add_pw_resource_flavors
 from .resources import AutoprovisioningConfig
 from .scheduling import (
     create_accelerator_label,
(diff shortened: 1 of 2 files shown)


samos123:
@SujeethJinesh @Obliviour i didn't have time to test yet but this is basically what's needed for axlearn.

src/xpk/core/kueue.py
```diff
@@ -480,30 +479,50 @@ def install_kueue_crs(
...
     total_chips: total number of chips for the specific resource type.
+    enable_pathways: if pathways is enabled.
 
   Returns:
     A string of Kueue covered resources configuration.
   """
+  pathways_resources = ''
+  if enable_pathways:
+    pathways_resources = """
+    - name: cpu-user
```
RoshaniN, line 498: I don't think you can have two resource flavors with the same covered resources, IIUC. Please test before submitting this change.

RoshaniN (changes requested):
Thanks Sam!

samos123:
Closing this PR because it was made using a fork. This is the new PR: https://github.com/AI-Hypercomputer/xpk/pull/600

PR #575, opened 2025-08-05
[Slice] Cleanup unit tests

mbobrovskyi:
## Fixes / Features
- Cleanup unit tests

### DIFF
--- a/slice/internal/controller/workload_controller_test.go
+++ b/slice/internal/controller/workload_controller_test.go
@@ -58,8 +58,15 @@ var (
 	errTest = errors.New("test error")
 )
 
+var (
+	jobSetGVK   = jobset.SchemeGroupVersion.WithKind("JobSet")
+	jobGVK      = batchv1.SchemeGroupVersion.WithKind("Job")
+	workloadGVK = kueue.SchemeGroupVersion.WithKind("Workload")
+)
+
 func TestWorkloadReconciler(t *testing.T) {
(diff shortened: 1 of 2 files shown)


slice/internal/controller/workload_controller_test.go
```diff
@@ -58,8 +58,15 @@ var (
 	errTest = errors.New("test error")
 )
 
+var (
```
PBundyra, line 61:
```suggestion
const (
```

PBundyra:
I wonder if `.PodSets(basePodSets...).(baseAdmission, now)` shouldn't be a part of `baseWorkload` as they occur in almost every scenario. WDYT @gabesaba

PBundyra:
Also fields: `interceptorFuncsCreate` and `err` appear in only one scenario, so maybe we could have a separate test suite for that and remove them from `TestWorkloadReconciler`

slice/internal/controller/workload_controller_test.go
mbobrovskyi: We can't. This is calculation value.

mbobrovskyi:
> Also fields: interceptorFuncsCreate and wantErr appear in only one scenario, so maybe we could have a separate test suite for that and remove them from TestWorkloadReconciler

This field is not set in other tests. I'm not sure if creating a separate test is the better approach.

PR #600, opened 2025-08-20
feat: Pathways use single resource group

samos123:
Jobs requesting TPU resources may also have requests for CPU and memory. However when pathways is enabled, Kueue will not be able to admit such jobs since there is no cpu and memory quota.

This fix adds a very high number of CPU and memory for TPU/GPU resources and merges the pathways resource group with the accelerator resource group.

This also allows us to run AXLearn jobs without having to make changes manually.

Follow up from: https://github.com/AI-Hypercomputer/xpk/pull/574
this time with a branch within xpk repo.


samos123:
Seems @lukebaumann encountered an issue when not using `create-pathways`. A potential fix is to remove the `create-pathways` command all together since it doesn't seem needed. We may be able to get rid of cpu resource flavor which would also unblock AXLearn jobs.

samos123:
This PR should solve NAP and AXLearn support as well. Would prefer to get this merged and will check with Luke on why it wasn't working for him.

samos123:
It's still needed for us to be able to run AXLearn on xpk clusters.

PR #606, opened 2025-08-27
Fix nodepool creation

scaliby:
## Fixes
When cluster is re-created but nodepool numbers are not consecutive numbers and desired node pool size is smaller than the original one, XPK removes nodepools that are above the specified limit.

For instance, when user creates a cluster with `--num-slices=3`, it creates a following node pools:
```diff
+ 'cluster-np-0',
+ 'cluster-np-1',
+ 'cluster-np-2',
```

Then, a user manually deletes a node pool on the UI or through `gcloud` command so the list of node pools is as follows:
```diff
'cluster-np-0',
- 'cluster-np-1',
'cluster-np-2',
```

Then, as a user re-runs cluster create with `--num-slices=2`, XPK will do propose following changes:
```diff
'cluster-np-0',
+ 'cluster-np-1'
- 'cluster-np-2'
```
Where no changes are expected.

After this change XPK will propose no changes to node pools in the mentioned case.

## Testing / Documentation
Change has been covered with unit tests.

- [y] Tests pass
- [y] Appropriate changes to documentation are included in the PR

### DIFF
--- a/src/xpk/core/tests/unit/test_nodepool.py
+++ b/src/xpk/core/tests/unit/test_nodepool.py
@@ -70,9 +70,9 @@ def test_compute_desired_node_pool_names_with_consecutive_numbers_missing_and_de
 def test_compute_desired_node_pool_names_with_unknown_node_pools():
   result = get_desired_node_pool_names(
       existing_node_pool_names=[
+          "unknown-node-pool",
           node_pool_name(0),
           node_pool_name(3),
-          "unknown-node-pool",
       ],
       cluster_name=CLUSTER_NAME,
       desired_node_pool_count=2,


src/xpk/core/tests/unit/test_nodepool.py
```diff
@@ -0,0 +1,90 @@
...
+    ]
+    assert set(result) == set(expected_result)
+
+
+def test_compute_desired_node_pool_names_with_unknown_node_pools():
+    result = get_desired_node_pool_names(
+        existing_node_pool_names=[
+            node_pool_name(0),
+            node_pool_name(3),
+            "unknown-node-pool"
```
jamOne-, line 80: nit: could you put this non-standard name as first in the list. It doesn't matter for the current code, but I feel it might have caught some potential bugs in the future.

PR #629, opened 2025-09-16
Fix most type annotations and remove all inline type suppressions.

SikaGrr:
This change should have almost no effect on runtime (just type annotations).

### DIFF
--- a/src/xpk/commands/cluster.py
+++ b/src/xpk/commands/cluster.py
@@ -92,7 +92,7 @@ def cluster_adapt(args) -> None:
 
   system, return_code = get_system_characteristics(args)
 
-  if return_code > 0:
+  if return_code > 0 or system is None:
     xpk_print('Fetching system characteristics failed!')
     xpk_exit(return_code)
 
@@ -199,7 +199,7 @@ def cluster_create(args) -> None:
   """
   system, return_code = get_system_characteristics(args)
 
-  if return_code > 0:
+  if return_code > 0 or system is None:
     xpk_print('Fetching system characteristics failed!')
     xpk_exit(return_code)
 
@@ -215,13 +215,13 @@ def cluster_create(args) -> None:
     xpk_exit(0)
 
   return_code, gke_server_config = get_gke_server_config(args)
-  if return_code != 0:
+  if return_code != 0 or gke_server_config is None:
     xpk_exit(return_code)
 
   return_code, gke_control_plane_version = get_gke_control_plane_version(
       args, gke_server_config
   )
-  if return_code != 0:
+  if return_code != 0 or gke_control_plane_version is None:
     xpk_exit(return_code)
 
   create_cluster_command_code = create_cluster_if_necessary(
@@ -396,7 +396,7 @@ def cluster_cacheimage(args) -> None:
   get_cluster_credentials(args)
   system, return_code = get_system_characteristics(args)
 
-  if return_code > 0:
+  if return_code > 0 or system is None:
     xpk_print('Fetching system characteristics failed!')
     xpk_exit(return_code)
 
@@ -813,7 +813,7 @@ def check_deployment_exists(args, deployment_name: str, namespace: str) -> bool:
   result = run_command_with_updates(
       command, 'Waiting for kubeDNS to be checked.', args
   )
-  return result
+  return result != 0
 
 
 def verify_coredns_readiness(
@@ -872,7 +872,7 @@ def cleanup_coredns_repo(coredns_repo_full_path: str):
     xpk_print(f'Error deleting directory {coredns_repo_full_path}: {e}')
 
 
-def update_coredns(args):
+def update_coredns(args) -> int:
   """Updates and deploys CoreDNS within a cluster.
 
   Args:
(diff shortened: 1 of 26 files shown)


src/xpk/commands/cluster.py
```diff
@@ -92,7 +92,7 @@ def cluster_adapt(args) -> None:
 
   system, return_code = get_system_characteristics(args)
 
-  if return_code > 0:
+  if return_code > 0 or system is None:
```
jamOne-, line 95: Alternatively we could add `assert(system)` below the `if`. But this approach LGTM too.

src/xpk/commands/cluster.py
```diff
@@ -813,7 +813,7 @@ def check_deployment_exists(args, deployment_name: str, namespace: str) -> bool:
   result = run_command_with_updates(
       command, 'Waiting for kubeDNS to be checked.', args
   )
-  return result
+  return result != 0
```
jamOne-, line 816: Could you please add a TODO here, to increase the visibility of this bug, if by any chance one of us uses this function to be aware that it's wrong

src/xpk/utils/console.py
```diff
@@ -29,7 +30,7 @@ def xpk_print(*args, **kwargs):
   sys.stdout.flush()
 
 
-def xpk_exit(error_code):
+def xpk_exit(error_code) -> NoReturn:
```
jamOne-, line 33: That's perfect! TIL

src/xpk/commands/info.py
```diff
@@ -51,19 +51,19 @@ def info(args: Namespace) -> None:
...
 
-  if lq:
+  if lq and lqs is not None:
     print_formatted_lqs(lqs, quotas)
 
   if cq:
     print_formatted_cqs(cqs, quotas)
 
 
-def get_nominal_quotas(cqs: list[dict]) -> dict[str, dict[str, str]]:
```
jamOne-, line 61: wow

src/xpk/core/nap.py
```diff
@@ -336,11 +337,13 @@ def get_autoprovisioning_node_selector_args(args) -> tuple[str, int]:
       )
       return node_selector_args, 1
 
-    return_code, capacity_type_str = get_value_from_map(
+    return_code, optional_capacity_type_str = get_value_from_map(
         CAPACITY_TYPE_CONFIG_KEY, cluster_config_map
     )
     if return_code != 0:
       return node_selector_args, return_code
+    # return_code==0 implies capacity_type is defined
```
jamOne-, line 345: optional:
```
return_code, capacity_type_str = get_value_from_map( ... )
if return_code != 0:
  return node_selector_args, return_code
assert(capacity_type_str)
```

but it's less explicit

src/xpk/core/system_characteristics.py
```diff
@@ -55,7 +55,7 @@ class SystemCharacteristics:
   gke_accelerator: str
   gce_machine_type: str
   chips_per_vm: int
-  accelerator_type: AcceleratorType  # type: ignore
+  accelerator_type: int
```
jamOne-, line 58: Why can't it be `AcceleratorType`? It'd be nice.

BTW is this how we define enums in Python?
```
AcceleratorType = {'TPU': 1, 'GPU': 2, 'CPU': 3}
```

src/xpk/parser/storage.py
```diff
@@ -28,6 +28,13 @@
     add_kind_cluster_arguments,
     add_shared_arguments,
 )
+from typing import Protocol, Any
+
+
+class Subcommands(Protocol):
+
+  def add_parser(self, *args, **kwargs) -> Any:
+    ...
```
jamOne-, line 37: What does it do?

jamOne- (commented):
Thank you Ivan, great work! You found so many wrong types it's shocking!

src/xpk/parser/storage.py
SikaGrr: duck typing check for presence of add_parser function, since we don't have a common base type (or even visible type):
https://mypy.readthedocs.io/en/stable/duck_type_compatibility.html
https://mypy.readthedocs.io/en/stable/protocols.html#protocol-types

src/xpk/core/nap.py
SikaGrr: Here I prefer this approach because if somehow we get a None we don't want to crash (in other places we would crash anyway).

src/xpk/commands/cluster.py
SikaGrr: Done.

src/xpk/commands/cluster.py
SikaGrr: I'll have a followup to this PR with remaining changes, I prefer to minimize any runtime differences in this PR.

jamOne-: Do you know if the `asserts` are being removed from the final code that we release, or do they stay?

nitpicking, but `or system is None` is also a runtime difference

src/xpk/parser/storage.py
jamOne-: I'll read that, thanks!

jamOne- (approved):

PR #633, opened 2025-09-18
Use kubernetes.io/hostname as third level.

### LINKED ISSUE #614: [slice] e2e tests still passing after removing health label
We annotate each replicated job in JobSet here: https://github.com/AI-Hypercomputer/xpk/blob/1d2ac6b4d1e896cbd0ffde3781e0067a6ffd2812/slice/internal/webhooks/jobset_webhook.go#L66

However, after deletion/modification of the label used in e2e test, the tests still pass. Remove this line: https://github.com/AI-Hypercomputer/xpk/blob/1d2ac6b4d1e896cbd0ffde3781e0067a6ffd2812/slice/hack/kind-cluster.yaml#L30

And observe e2e tests still pass. Kueue should not return topology assignments on such nodes for these replicated jobs, which require this NodeLabel

**Additional Info**
Modifying the [loop](https://github.com/kubernetes-sigs/kueue/blob/b44cf15a52b5c97d46d172dcf2432228bd03403e/pkg/util/testingjobs/jobset/wrappers.go#L79) which creates each replicated job to add a node selector:
```
jt.Spec.Template.Spec.NodeSelector = map[string]string{"hello": "world"}
```
Results in Kueue's e2e tests failing. So, it seems that Kueue/TAS is acting upon these node selectors in some way. This makes me think that this is more likely a Slice Controller issue.

mbobrovskyi:
## Fixes / Features
- Use kubernetes.io/hostname as third level.

FIxes #614


slice/internal/topology/topology.go
```diff
@@ -16,15 +16,12 @@ package topology
...
 
-// SubblockLevelIndex returns the index of the TPUSubBlock topology
-// assignment, or -1 if it doesn't exist.
-func SubblockLevelIndex(topologyAssignment *kueue.TopologyAssignment) int {
-	for i, level := range topologyAssignment.Levels {
-		if level == core.TPUSubBlockLabel {
+// LevelIndex returns the index of the specific level topology assignment, or -1 if it doesn't exist.
+func LevelIndex(topologyAssignment *kueue.TopologyAssignment, level string) int {
+	for i, currentLevel := range topologyAssignment.Levels {
+		if level == currentLevel {
```
gabesaba, line 24: After comment above:
```
func HostnameLevelIndex(topologyAssignment)
```

slice/internal/topology/validation.go
```diff
@@ -29,18 +34,48 @@ func AnyAssignment(admission *kueue.Admission) bool {
...
+
+	hostnameLevelIndex := LevelIndex(psa.TopologyAssignment, corev1.LabelHostname)
+	if hostnameLevelIndex == -1 {
+		return false
+	}
+
+	for _, domain := range psa.TopologyAssignment.Domains {
+		if hostnameLevelIndex >= len(domain.Values) {
+			return false
+		}
```
gabesaba, line 66: when would this happen? would this have to be bug in TAS?

slice/internal/topology/validation.go
```diff
@@ -29,18 +34,48 @@ func AnyAssignment(admission *kueue.Admission) bool {
...
+		if GetTPUSubBlockLabelValue(domain, hostnameLevelIndex, nodes) == "" {
+			return false
+		}
```
gabesaba, line 69: What do you think of this for readability?

```
nodeName := domain.Values[hostnameLevelIndex]
GetTPUSubBlockLabelValue(nodes, nodeName)
```

slice/test/e2e/jobset_test.go
```diff
@@ -186,7 +187,7 @@ var _ = ginkgo.Describe("JobSet", func() {
 					gomega.Expect(createdWorkload.Status.Admission.PodSetAssignments).Should(gomega.HaveLen(1))
 					gomega.Expect(createdWorkload.Status.Admission.PodSetAssignments[0].TopologyAssignment).Should(gomega.BeComparableTo(
 						&kueue.TopologyAssignment{
-							Levels:  []string{"cloud.google.com/gce-topology-block", "cloud.google.com/gke-tpu-slice-4x4x4-id"},
+							Levels:  []string{corev1.LabelHostname},
```
gabesaba, line 190: nit: please use string literal

slice/test/e2e/jobset_test.go
```diff
@@ -63,7 +63,7 @@ var _ = ginkgo.Describe("JobSet", func() {
 		utils.MustCreate(ctx, k8sClient, ns)
 
 		topology = testing.MakeTopology("topology").
-			Levels("cloud.google.com/gce-topology-block", "cloud.google.com/gke-tpu-slice-4x4x4-id").
+			Levels("cloud.google.com/gce-topology-block", "cloud.google.com/gke-tpu-slice-4x4x4-id", corev1.LabelHostname).
```
gabesaba, line 66: nit: string literal for LabelHostName

slice/internal/controller/workload_controller.go
```diff
@@ -457,38 +482,43 @@ func (r *WorkloadReconciler) syncSlices(ctx context.Context, wl *kueue.Workload,
...
 	nodeSelectors := sets.New[string]()
 	// we already validated that all assignments have a valid level,
 	// in validateRelevantWorkload.
-	subblockLevelIndex := topology.SubblockLevelIndex(topologyAssignment)
-	for _, domain := range topologyAssignment.Domains {
-		nodeSelectors.Insert(domain.Values[subblockLevelIndex])
+	if subblockLevelIndex := topology.LevelIndex(topologyAssignment, core.TPUSubBlockLabel); subblockLevelIndex != -1 {
+		for _, domain := range topologyAssignment.Domains {
+			nodeSelectors.Insert(domain.Values[subblockLevelIndex])
+		}
```
gabesaba, line 499: Let's delete this branch, as I don't think we need it any more (and if I'm mistaken, we can restore it). I prefer to have just one well-lit path (then we won't have to update tests to cover both)

mbobrovskyi: However, deleting this branch would break the two-level topology. Do you think that's acceptable?

slice/internal/topology/validation.go
mbobrovskyi: It shouldn't happen, but this ensures we prevent a panic here.

slice/internal/controller/workload_controller.go
```diff
@@ -201,6 +207,19 @@ func shouldFinalize(wl *kueue.Workload) (bool, string) {
 	return false, ""
 }
 
+func (r *WorkloadReconciler) getNodes(ctx context.Context) (map[string]corev1.Node, error) {
+	nodes := &corev1.NodeList{}
+	err := r.client.List(ctx, nodes)
+	if err != nil {
+		return nil, err
+	}
+	mapNodes := make(map[string]corev1.Node)
```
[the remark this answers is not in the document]

slice/internal/topology/validation.go
mbobrovskyi: I would like to keep it just to ensure we never get a panic.

slice/internal/topology/validation.go
mbobrovskyi: Good point!

slice/test/e2e/jobset_test.go
mbobrovskyi: Done

slice/internal/controller/workload_controller.go
gabesaba: Yes, because the 2-level topology doesn't work with the health labels properly (which is an important feature)

slice/internal/topology/validation.go
gabesaba: Since this is an invariant which shouldn't happen, I'm okay with the program panicking here. If there were some major problem with TAS, it would be better to catch it here than to silently fail

As an alternative we could add a log line, but I'm not against the panic.

slice/internal/controller/workload_controller.go
mbobrovskyi: Make sense.

gabesaba (commented):
lgtm. nice work.

PR #676, opened 2025-10-08
Stop using GPP if WP is used

FIoannides:
Workload policy and group placement policy are mutually exclusive and since for TPUv7x we create WP we should avoid enabling GPP as well


FIoannides:
> Should we now have at least one golden for an older tpu version?

Cluster private is using v5 and it retains this flag

FIoannides:
> Have you checked golden affected workflows without dry-run?

Not yet, I am trying now on staging, but if GKE team confirms this is enough we can merge even without a real run IMO

jamOne- (approved):

PR #693, opened 2025-10-20
Workload commands granular dependencies validation

scaliby:
## Fixes / Features
Verify only required dependencies in workload commands

## Testing / Documentation
Checked all commands manually.

- [X] Tests pass

### DIFF
--- a/src/xpk/utils/validation.py
+++ b/src/xpk/utils/validation.py
@@ -91,10 +91,10 @@ def validate_dependencies():
 def validate_dependencies_list(dependencies: list[SystemDependency]):
   """Validates a list of system dependencies and returns none or exits with error."""
   for dependency in dependencies:
-    validate_dependency(dependency)
+    _validate_dependency(dependency)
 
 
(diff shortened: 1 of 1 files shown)


jamOne- (changes requested):

src/xpk/commands/workload.py
```diff
@@ -308,7 +308,9 @@ def workload_create(args) -> None:
     0 if successful and 1 otherwise.
   """
   if should_validate_dependencies(args):
-    validate_dependencies()
+    validate_dependency(SystemDependency.KUBECTL)
+    validate_dependency(SystemDependency.GCLOUD)
+    validate_dependency(SystemDependency.DOCKER)
```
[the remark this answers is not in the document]
scaliby: Done.

src/xpk/utils/validation_test.py
jamOne-: nit: IMO we should not delete the tests for `validate_dependency` and the list test should have at least one test with multiple values

scaliby: why? :D

PR #704, opened 2025-10-20
Kind kueue

FIoannides:
## Fixes / Features
- Update kind to use Kueue manager instead of Kueue.py
- Remove last few usages of Kueue.py and remove the file

## Testing / Documentation
Testing details.

- [y] Tests pass

### DIFF
--- a/src/xpk/parser/cluster.py
+++ b/src/xpk/parser/cluster.py
@@ -31,7 +31,7 @@ from ..core.config import CFG_BUCKET_KEY
 from ..core.vertex import DEFAULT_VERTEX_TENSORBOARD_NAME
 from .common import add_shared_arguments, ParserOrArgumentGroup
 from .validators import name_type
-from ..utils.feature_flags import SUB_SLICING_ENABLED
+from ..utils.feature_flags import FeatureFlags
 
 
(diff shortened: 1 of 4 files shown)


src/xpk/commands/info.py
```diff
@@ -46,7 +46,8 @@ def info(args: Namespace) -> None:
   add_zone_and_project(args)
   get_cluster_credentials(args)
 
-  verify_kueuectl()
+  kueue_manager = KueueManager()
+  kueue_manager.verify_kueuectl()
```
scaliby: kueue installed is already verified at line 41. Please remove.

scaliby (changes requested):

src/xpk/commands/cluster.py
```diff
@@ -82,6 +79,40 @@
...
+              - key: {nodeSelectorKey}
+                operator: Exists
+      tolerations:
+      - operator: "Exists"
+      containers:
+      - image: {image_name}
+        name: {cachekey}
+        command: [ "sleep", "inf" ]
+"""
+
```
jamOne-: nit: Could we already use j2 template for this?

PR #705, opened 2025-10-20
Use alpha resource-policies command

scaliby:
## Fixes / Features
This command is only available in alpha version of gcloud, so using alpha.

## Testing / Documentation

- [X] Tests pass


scaliby:
Discussed with team. This won't be necessary.

PR #710, opened 2025-10-21
feat: Add and use sub-slicing topology during Kueue configuration

jamOne-:
## Fixes / Features
- Add and use sub-slicing topology during Kueue configuration (http://b/452626879).
- I omitted the `NoSchedule` TPU toleration, because I'm not convinced it's required. We'll revisit this later if it turns out Sub-slicing won't work without it.
- Changed the way the yamls are concatenated in kueue_manager
- Also, made some kueue_manager_test improvements.

## Testing / Documentation
- [ y ] Tests pass


jamOne-:
OK

PR #719, opened 2025-10-22
feat: Add and use sub-slicing topology during Kueue configuration

jamOne-:
## Fixes / Features
- Add and use sub-slicing topology during Kueue configuration (http://b/452626879).
- I omitted the NoSchedule TPU toleration, because I'm not convinced it's required. We'll revisit this later if it turns out Sub-slicing won't work without it.
- Changed the way the yamls are concatenated in kueue_manager.

## Testing / Documentation
- [ y ] Tests pass

### DIFF
--- a/goldens/Workload_create.txt
+++ b/goldens/Workload_create.txt
@@ -25,7 +25,7 @@ docker tag dry-run-runner gcr.io/golden-project/dry-run-runner:prefix-current
 [XPK] Task: `Upload Docker Image` is implemented by the following command not running since it is a dry run. 
 docker push gcr.io/golden-project/dry-run-runner:prefix-current
 [XPK] Task: `Creating Workload` is implemented by the following command not running since it is a dry run. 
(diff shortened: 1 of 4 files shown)


src/xpk/commands/cluster_gcluster.py
```diff
@@ -159,6 +160,8 @@ def __install_kueue(args) -> int:
           cpu_limit=args.cpu_limit,
           is_pathways_cluster=args.enable_pathways,
           flex=args.flex,
+          configure_sub_slicing=FeatureFlags.SUB_SLICING_ENABLED
```
scaliby: nit: WDYT about wrapping this in brackets?
```
configure_sub_slicing=(
    FeatureFlags.SUB_SLICING_ENABLED
    and args.sub_slicing
),
```
IMO this is more readable.

FIoannides (approved):

PR #725, opened 2025-10-23
[slice] Error handling

pajakd:
Previously, when the slices transitions into an error state, the workload would get stuck as the AdmissionCheck transitioned into Rejected state. 

With this PR I'm changing the logic to transition into Retry state in case of Errors. Also, I'm cleaning up the Errored slices in the reconcile. This makes sure that we recreate a new slice and retry.

## Testing / Documentation
Testing details.

- [ y/n ] Tests pass

### DIFF
--- a/slice/internal/controller/workload_controller.go
+++ b/slice/internal/controller/workload_controller.go
@@ -310,7 +310,7 @@ func (r *WorkloadReconciler) findWorkloadSlices(ctx context.Context, wl *kueue.W
 // Returns:
 //   - A slice containing deleted Slice objects (with non-zero DeletionTimestamp).
 //   - A slice containing deformed Slice objects (being torn down).
-//   - A slice containing errored Slice objects
+//   - A slice containing errored Slice objects.
 //   - A slice containing other Slice objects (active/valid slices).
 func (r *WorkloadReconciler) groupSlices(slices []v1alpha1.Slice) ([]v1alpha1.Slice, []v1alpha1.Slice, []v1alpha1.Slice, []v1alpha1.Slice) {
 	var deleted, deformed, errored, other []v1alpha1.Slice


slice/test/e2e/jobset_test.go
```diff
@@ -569,6 +569,266 @@ var _ = ginkgo.Describe("JobSet", func() {
...
+
+			ginkgo.By("Checking that Slice is deleted", func() {
+				utils.ExpectObjectToBeDeleted(ctx, k8sClient, createdSlice, false)
+			})
+
+			ginkgo.By("Checking that Workload is deleted", func() {
+				utils.ExpectObjectToBeDeleted(ctx, k8sClient, createdWorkload, false)
+			})
+		})
+
```
PBundyra: This part feels like a duplicate of the test above, I think there's no need to check the same logic twice

PBundyra (commented):
Controller logic lgtm, left some nits regarding tests

slice/internal/controller/workload_controller.go
```diff
@@ -297,19 +311,21 @@ func (r *WorkloadReconciler) findWorkloadSlices(ctx context.Context, wl *kueue.W
 //   - A slice containing deleted Slice objects (with non-zero DeletionTimestamp).
 //   - A slice containing deformed Slice objects (being torn down).
 //   - A slice containing other Slice objects (active/valid slices).
```
mbobrovskyi: Could you please update docs?

slice/internal/controller/workload_controller.go
```diff
@@ -176,7 +176,21 @@ func (r *WorkloadReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
...
+	if len(errored) > 0 {
+		log.V(3).Info(
+			"Deleting errored Slices",
+			"erroredSlices", klog.KObjSlice(errored),
+		)
+		err = r.deleteSlices(ctx, errored)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+	}
```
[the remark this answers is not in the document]
pajakd: But cleanupSlices happens only when we finalize the workload and cleanup all the slices.  

But we also want to cover the case where a for example one of two slices changes state to `Error` while the workload is running. Then we would like to evict the workload, delete the one in `Error`, recreate it and hopefully re-admit the workload.

slice/internal/controller/workload_controller.go
```diff
@@ -616,8 +632,10 @@ func prepareAdmissionCheckStatus(ac *kueue.AdmissionCheckState, slices []v1alpha
 	slicesByState, noState := groupSlicesByState(slices)
 
 	switch {
-	case len(slicesByState[v1alpha1.Error]) > 0 || len(slicesByState[v1alpha1.Deformed]) > 0:
+	case len(slicesByState[v1alpha1.Deformed]) > 0:
 		ac.State = kueue.CheckStateRejected
+	case len(slicesByState[v1alpha1.Error]) > 0:
+		ac.State = kueue.CheckStateRetry
```
mbobrovskyi: Is it possible both statuses?

slice/internal/controller/workload_controller.go
```diff
@@ -299,9 +299,9 @@ func (r *WorkloadReconciler) findWorkloadSlices(ctx context.Context, wl *kueue.W
 	return slices.Items, nil
 }
 
-// groupSlices categorizes a list of Slice objects into three groups based on their state.
+// groupSlices categorizes a list of Slice objects into four groups based on their state.
 // It separates slices into deleted (marked for deletion), deformed (being torn down),
-// and other (active) slices.
+// errored and other (active) slices.
```
mbobrovskyi:
```suggestion
// errored (has encountered an error) and other (active) slices.
```
For consistency

PR #726, opened 2025-10-23
Add packages and access templates instead of absolute path

FIoannides:
https://chat.google.com/room/AAAA5hZCgEg/VMK90uPYyz8

- [y] Tests pass


jamOne- (approved):

jamOne- (changes requested):
goldens broken pls fix

scaliby (changes requested):
Will this work on a built whl package? Could you try building a whl package, installing it and seeing if it works?

jamOne- (approved):

PR #729, opened 2025-10-23
Fix semver compare in kueue_manager

scaliby:
## Fixes / Features
Current semver comparison is broken. Fixed it.

## Testing / Documentation
Unit testing.

- [X] Tests pass

### DIFF
--- a/src/xpk/core/kueue_manager.py
+++ b/src/xpk/core/kueue_manager.py
@@ -112,9 +112,9 @@ class KueueManager:
         )
         return 0
       else:
-        xpk_print(f"Upgrading Kueue to version {self.kueue_version}...")
+        xpk_print(f"Upgrading Kueue to version v{self.kueue_version}...")
     else:
-      xpk_print(f"Installing Kueue version {self.kueue_version}...")
+      xpk_print(f"Installing Kueue version v{self.kueue_version}...")
 
     install_return_code = self.__install(tolerations)
     if install_return_code != 0:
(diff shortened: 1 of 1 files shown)


src/xpk/core/kueue_manager.py
```diff
@@ -104,7 +105,10 @@ def install_or_upgrade(
     return_code, installed_version = self.get_installed_kueue_version()
 
     if return_code == 0:
-      if installed_version and installed_version > self.kueue_version:
+      if (
+          installed_version
+          and semver_compare(installed_version, self.kueue_version) >= 0
```
jamOne-: It used to be strictly greater on purpose. The idea is that if someone does a re-run we'll recreate all the resources.

src/xpk/utils/semver.py
```diff
@@ -0,0 +1,43 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+
+
+def semver_compare(a: str, b: str) -> int:
```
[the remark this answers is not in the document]
scaliby: I'm not sure if it makes sense to add another dependency for such a simple use case.

src/xpk/core/kueue_manager.py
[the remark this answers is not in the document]
scaliby: Done.

jamOne-: WDYT: Could we please operate on `Version` inside the whole file, not only in this one place?

jamOne-: nit: Add the `v` prefix here as well

PR #739, opened 2025-10-27
Add backoff_retry script and use it in create-pathways job

scaliby:
## Fixes / Features
This change will re-run the most flaky step in cluster / create-private job.

## Testing / Documentation
Tested the script on simple hello world with different exit statuses.

- [X] Tests pass

### DIFF
--- a/.github/workflows/build_tests.yaml
+++ b/.github/workflows/build_tests.yaml
@@ -42,8 +42,6 @@ jobs:
           RUN_ID="dispatch"
         elif [ "${{ github.ref }}" == "refs/heads/main" ]; then
           RUN_ID="main"
-        elif [ "${{ github.ref }}" == "refs/heads/develop" ]; then
-          RUN_ID="develop"
         else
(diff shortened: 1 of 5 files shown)


backoff_retry.sh
```diff
@@ -0,0 +1,113 @@
...
+# ARGUMENTS:
+#   $1 (Required): The command string to execute. This string *must* be
+#                  quoted if it contains spaces, pipes, or other special
+#                  shell characters.
+#
+# ENVIRONMENT VARIABLES:
+#   ATTEMPTS:     The total number of times to try the command.
+#                 (Default: 3)
+#
+#   BASE_DELAY:   The initial wait time (in seconds) used before the *first*
```
jamOne-: 1. Why not arguments?
2. "ATTEMPTS" and "BASE_DELAY" doesn't sound like they are specific to the backoff_retry and can potentially colide with other env variables. Should we add some prefix?

scaliby: Changed.

PR #753, opened 2025-10-29
Fix resource policy lifecycle management in cluster operations

### LINKED ISSUE #752: `xpk cluster delete` does not clean up workload-policy, causing errors on recreate with different topology
### Problem:

When using `xpk cluster create` for TPU types that require a workload-policy (e.g., tpu7x with specific topologies), XPK correctly creates a resource policy named `{cluster-name}-placement-policy` with the specified `acceleratorTopology`.

However, `xpk cluster delete` does not delete this associated resource policy. If a user then runs `xpk cluster create` again *with the same cluster name but a different TPU topology* (e.g., deleting a `tpu7x-64` cluster and creating a `tpu7x-128` cluster), the `cluster create` command fails.

This failure occurs because the `gcloud beta container node-pools create` command (generated internally by XPK) attempts to use the *existing* `{cluster-name}-placement-policy`, which still contains the `acceleratorTopology` from the *previous* cluster configuration (e.g., `2x4x4` instead of the newly requested `4x4x4`). This mismatch leads to an error during node pool creation, similar to the one encountered when manually providing an incorrect `--placement-policy` with a different topology.

### Steps to Reproduce:

1.  Create a cluster requiring a workload policy with a specific topology:
    ```bash
    xpk cluster create --cluster my-tpu-cluster --tpu-type=tpu7x-64 --reservation=RESERVATION_ID ...
    ```
    *(A workload policy named `my-tpu-cluster-placement-policy` with `acceleratorTopology: 2x4x4` is created)*
2.  Delete the cluster:
    ```bash
    xpk cluster delete --cluster my-tpu-cluster ...
    ```
    *(The workload policy `my-tpu-cluster-placement-policy` remains)*
3.  Attempt to create the cluster again with the same name but a different topology:
    ```bash
    xpk cluster create --cluster my-tpu-cluster --tpu-type=tpu7x-128 --reservation=RESERVATION_ID ...
    ```
4.  Observe the node pool creation failure due to the topology mismatch between the requested node pool (`4x4x4`) and the existing placement policy (`2x4x4`).

```
ERROR: (gcloud.beta.container.node-pools.create) Operation [<Operation
 clusterConditions: [<StatusCondition
 canonicalCode: CanonicalCodeValueValuesEnum(INVALID_ARGUMENT, 3)
 message: "Google Compute Engine: Invalid value for field 'resource.resizeBy': '16'. Requested invalid target size '16' for a Managed Instance Group in the gang mode of size '8'. Should be equal to the gang size.">]
 detail: "Google Compute Engine: Invalid value for field 'resource.resizeBy': '16'. Requested invalid target size '16' for a Managed Instance Group in the gang mode of size '8'. Should be equal to the gang size."
 endTime: '2025-10-28T14:39:44.912501721Z'
 error: <Status
 code: 3
 details: []
 message: "Google Compute Engine: Invalid value for field 'resource.resizeBy': '16'. Requested invalid target size '16' for a Managed Instance Group in the gang mode of size '8'. Should be equal to the gang size.">
 name: 'operation-1761662360188-eb0a551b-11ee-4ffd-8c31-a778e534cff8'
 nodepoolConditions: []
 operationType: OperationTypeValueValuesEnum(CREATE_NODE_POOL, 7)
 selfLink: 'https://container.googleapis.com/v1beta1/projects/579803409133/locations/us-central1/operations/operation-1761662360188-eb0a551b-11ee-4ffd-8c31-a778e534cff8'
 startTime: '2025-10-28T14:39:20.188771275Z'
 status: StatusValueValuesEnum(DONE, 3)
 statusMessage: "Google Compute Engine: Invalid value for field 'resource.resizeBy': '16'. Requested invalid target size '16' for a Managed Instance Group in the gang mode of size '8'. Should be equal to the gang size."
 targetLink: 'https://container.googleapis.com/v1beta1/projects/579803409133/locations/us-central1/clusters/tpu-v7/nodePools/tpu-v7-np-0'
 zone: 'us-central1'>] finished with error: Google Compute Engine: Invalid value for field 'resource.resizeBy': '16'. Requested invalid target size '16' for a Managed Instance Group in the gang mode of size '8'. Should be equal to the gang size.
```

### Expected Behavior:

XPK should handle the lifecycle of the associated workload policy more robustly.

### Proposed Solution:

1.  **Enhance `xpk cluster delete`:**
      * Check if a resource policy named `{cluster-name}-placement-policy` exists.
      * If it exists, attempt to delete it using `gcloud compute resource-policies delete {cluster-name}-placement-policy ...`. Handle potential errors if the policy is still in use (though ideally, node pools using it should already be deleted).
2.  **Enhance `xpk cluster create`:**
      * Before creating a workload policy named `{cluster-name}-placement-policy`, check if it already exists.
      * If it **exists**:
          * **Delete** the existing policy using `gcloud compute resource-policies delete ...` (attempt deletion even if it might fail if in use, letting subsequent create fail clearly).
      * **Always create** a new policy with the correct topology required for the *current* `tpu-type` being requested using `gcloud compute resource-policies create workload-policy ...`.

bzantium:
# Description

This change fixes resource policy lifecycle management in XPK cluster operations by implementing proper cleanup and creation workflows.

**What it introduces:**
- Resource policy deletion during `xpk cluster delete` to prevent resource leaks
- Enhanced `xpk cluster create` that always ensures fresh resource policies with correct accelerator topology
- New `delete_resource_policy_if_exists()` helper function for safe policy cleanup
- Updated `ensure_resource_policy_exists()` to use a delete-then-create approach

**Why we need it:**
- Resource policies with pattern `{cluster-name}-placement-policy` were persisting after cluster deletion, causing resource accumulation
- When reusing cluster names, existing policies might have incorrect accelerator topologies for new cluster configurations
- This led to potential topology mismatches and resource management issues

**Documentation:**
No documentation updates needed as this is an internal resource management fix that doesn't change user-facing behavior.

# Issue

Fixes resource management issues where:
1. Resource policies created during cluster creation were not cleaned up during cluster deletion
2. Existing resource policies with mismatched topologies could cause issues when reusing cluster names
3. No mechanism existed to ensure resource policies always matched the required accelerator topology

# Testing

**Manual Testing:**
- Verified no linting errors introduced
- Confirmed backward compatibility maintained
- Tested edge cases (policy doesn't exist, deletion failures)

**Golden Verification:**
- Reviewed existing golden files to ensure no breaking changes to expected behavior
- The change only affects internal resource policy management and doesn't alter user-visible command outputs
- Golden files show the expected `gcloud compute resource-policies describe` calls which are now enhanced with proper cleanup

**Error Handling:**
- Gracefully handles cases where resource policy doesn't exist during deletion
- Provides clear error messages when policy operations fail
- Maintains existing error handling patterns in the codebase

Fixes #752

### DIFF
--- a/goldens/Basic_cluster_create.txt
+++ b/goldens/Basic_cluster_create.txt
@@ -50,8 +50,13 @@ gcloud beta container node-pools describe 0 --cluster golden-cluster --project=g
 [XPK] Task: `GKE Cluster Get ConfigMap` is implemented by the following command not running since it is a dry run. 
 kubectl get configmap golden-cluster-resources-configmap -o=custom-columns="ConfigData:data" --no-headers=true
 [XPK] Existing node pool names  ['0']
-[XPK] Task: `Retrieve resource policy` is implemented by the following command not running since it is a dry run. 
+[XPK] Task: `Check resource policy exists` is implemented by the following command not running since it is a dry run. 
 gcloud compute resource-policies describe golden-cluster-placement-policy --project=golden-project --region=us-central1
+[XPK] Task: `Delete resource policy` is implemented by the following command not running since it is a dry run. 
+gcloud compute resource-policies delete golden-cluster-placement-policy --project=golden-project --region=us-central1 --quiet
+[XPK] Deleted resource policy golden-cluster-placement-policy
+[XPK] Task: `Create resource policy` is implemented by the following command not running since it is a dry run. 
+gcloud compute resource-policies create workload-policy golden-cluster-placement-policy --project=golden-project --region=us-central1 --type=HIGH_THROUGHPUT --accelerator-topology=2x2x1
 [XPK] To complete NodepoolCreate-golden-cluster-np-0 we are executing gcloud beta container node-pools create golden-cluster-np-0 --location=us-central1 --cluster=golden-cluster --project=golden-project --node-locations=us-central1-a --machine-type=tpu7x-standard-4t --host-maintenance-interval=AS_NEEDED --spot --placement-policy=golden-cluster-placement-policy --enable-gvnic --node-version=0 --num-nodes=1 --scopes=storage-full,gke-default,"https://www.googleapis.com/auth/cloud-platform" --max-pods-per-node 15  
 [XPK] Breaking up a total of 1 commands into 1 batches
 [XPK] Pretending all the jobs succeeded
--- a/goldens/Cluster_create_with_gb200-4.txt
+++ b/goldens/Cluster_create_with_gb200-4.txt
@@ -52,8 +52,13 @@ gcloud beta container node-pools describe 0 --cluster golden-cluster --project=g
 [XPK] Task: `GKE Cluster Get ConfigMap` is implemented by the following command not running since it is a dry run. 
 kubectl get configmap golden-cluster-resources-configmap -o=custom-columns="ConfigData:data" --no-headers=true
 [XPK] Existing node pool names  ['0']
-[XPK] Task: `Retrieve resource policy` is implemented by the following command not running since it is a dry run. 
+[XPK] Task: `Check resource policy exists` is implemented by the following command not running since it is a dry run. 
 gcloud compute resource-policies describe golden-cluster-placement-policy --project=golden-project --region=us-central1
+[XPK] Task: `Delete resource policy` is implemented by the following command not running since it is a dry run. 
+gcloud compute resource-policies delete golden-cluster-placement-policy --project=golden-project --region=us-central1 --quiet
+[XPK] Deleted resource policy golden-cluster-placement-policy
+[XPK] Task: `Create resource policy` is implemented by the following command not running since it is a dry run. 
+gcloud compute resource-policies create workload-policy golden-cluster-placement-policy --project=golden-project --region=us-central1 --type=HIGH_THROUGHPUT --accelerator-topology=1x72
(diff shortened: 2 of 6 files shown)


src/xpk/commands/cluster.py
```diff
@@ -1086,6 +1087,12 @@ def run_gke_cluster_delete_command(args) -> int:
   if return_code != 0:
     return return_code
 
+  # Delete resource policy if it exists
```
jamOne-: nit: please remove the comment, it seems redundant

src/xpk/core/nodepool.py
```diff
@@ -597,27 +597,66 @@ def get_desired_node_pool_names(
...
+      (
+          'gcloud compute resource-policies create workload-policy'
+          f' {resource_policy_name} --project={args.project} --region={zone_to_region(args.zone)} --type=HIGH_THROUGHPUT'
+          f' --accelerator-topology={topology}'
+      ),
+      'Create resource policy',
+  )
+
+  if return_code != 0:
+    raise RuntimeError('Unable to create resource policy')
```
jamOne-: Our code doesn't use exceptions. Please print the error and run `xpk_exit(1)`

FIoannides:
First of all, thank you so much for the detailed issue description and the time you spent on this PR.
After discussing this issue internally, we have decided to go with a slightly different approach to solve this issue, as can be seen in https://github.com/AI-Hypercomputer/xpk/pull/760. This way will not only fix the original problem, but also reduce the amount of WPs on the project and allow for re-use.

bzantium:
> First of all, thank you so much for the detailed issue description and the time you spent on this PR. After discussing this issue internally, we have decided to go with a slightly different approach to solve this issue, as can be seen in #760. This way will not only fix the original problem, but also reduce the amount of WPs on the project and allow for re-use.

Thank you for taking xthis issue and PR into consideration fast! As long as the issue can be resolved, better solution is always welcomed! I checked #760 and look like much simpler. It would be great for me to add me as a contributor for the PR if you don't mind :)

PR #754, opened 2025-10-29
Add more retries to tests

scaliby:
# Description
These failed tonight. Added retries so likelihood of success will increase. Most likely we should increase timeout for nodepool creation long term, but this is the simplest fix ATM.

# Issue
b/445869635

# Testing
WIll be tested in integration testing workflow at night.

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -70,6 +70,8 @@ Before opening a pull request make sure your change passes the following checkli
 * Goldens are up-to-date - regenerate them using `make goldens` command.
 * Change is production ready, if not make sure it is covered with a feature flag. See sample flags [here](https://github.com/AI-Hypercomputer/xpk/blob/main/src/xpk/utils/feature_flags.py).
 
+You can read more about our testing guidance [here](./testing.md).
+
 **Code merged to the main branch is expected to be released at any given point in time, hence it needs to be treated as a production code.**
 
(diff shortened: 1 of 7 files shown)


.github/workflows/nightly_tests.yaml
```diff
@@ -87,7 +87,7 @@ jobs:
     - name: Run xpk inspector with the workload created above
       run: python3 xpk.py inspector --cluster $TPU_CLUSTER_NAME --zone=us-central2-b  --workload $WORKLOAD_NAME
     - name: Wait for workload completion and confirm it succeeded
-      run: python3 xpk.py workload list --cluster $TPU_CLUSTER_NAME --zone=us-central2-b --wait-for-job-completion $WORKLOAD_NAME --timeout 300
+      run: ./backoff_retry.sh -c "python3 xpk.py workload list --cluster $TPU_CLUSTER_NAME --zone=us-central2-b --wait-for-job-completion $WORKLOAD_NAME --timeout 300"
```
jamOne-: Should we just double the timeout here?

jamOne- (changes requested):

.github/workflows/nightly_tests.yaml
scaliby: Done.

PR #760, opened 2025-10-29
Update WP naming convention

FIoannides:
# Description

Change the WP naming convention so that for clusters using the same device type and the same topology, the WP is reused. This will also break the 1-1 association of WP to cluster names, which caused the issue described below and will also help reduce the total number of WPs on the project.

# Issue

https://github.com/AI-Hypercomputer/xpk/issues/752

# Testing

Manual testing was performed


jamOne- (approved):

bzantium:
look much simpler and gracefully fix the problem!

PR #762, opened 2025-10-29
Remove duplicated workload tests

scaliby:
# Description
Workloads are already tested [here](https://github.com/AI-Hypercomputer/xpk/blob/60aab7229c47ee4aeb7e8ffb19ecdc3dc1ace24d/.github/workflows/nightly_tests.yaml#L81) and [here](https://github.com/AI-Hypercomputer/xpk/blob/60aab7229c47ee4aeb7e8ffb19ecdc3dc1ace24d/.github/workflows/nightly_tests.yaml#L205) for pathways, so there is no need to test this twice.

# Issue
b/445869176

# Testing
N/A

### DIFF
--- a/.github/workflows/build_tests.yaml
+++ b/.github/workflows/build_tests.yaml
@@ -15,39 +15,13 @@
 name: Build Tests
 
 on:
-  workflow_dispatch:
-    inputs:
-      tpu-type:
-        description: 'TPU Type'
-        required: true
-        default: 'v4-8'
-        type: choice
-        options:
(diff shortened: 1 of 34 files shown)


scaliby:
> These two seem to be missing, http://screen/6uVbpEc8Y4qZoVf from the set that we are keeping, can we add them?

Added. Should be good now.

scaliby:
@jamOne- could you PTAL? This is blocked as you requested changes.

PR #763, opened 2025-10-29
feat: Make Kueue installation required to consider cluster creation successful

jamOne-:
# Description

Before this change, even though Kueue installation failed, the user could see the successful exit of XPK at the end.

Also style change for the `kubectl wait` command. `-nkueue-system` also worked, but the space makes it more clear.

# Issue

# Testing


SikaGrr (approved):
Would it be easy to add a test in cluster_test?

jamOne-:
@SikaGrr it's hard because I'd have to mock so many dependencies. Although it's possible to test the `install_kueue` function, that it returns the installation code.

PR #765, opened 2025-10-30
Integration tests on xpk artifact

scaliby:
# Description
Refactoring a first batch of tests to run on xpk final build wheel artifact rather than source code.

# Issue
b/445869874

# Testing
I will schedule nightly after merge. In case it fails, immediately rollback. I'm pretty confident this will work as I've done testing here: https://github.com/AI-Hypercomputer/xpk/pull/764


scaliby:
> Should we fix the "Workflow does not contain permissions" security findings?

Fixed already

PR #774, opened 2025-10-30
Fix remaining type checking exceptions.

SikaGrr:
# Description
Fix remaining type checking exceptions.

# Testing
No logic changes, just type annotations.

### DIFF
--- a/.github/workflows/build_tests.yaml
+++ b/.github/workflows/build_tests.yaml
@@ -15,39 +15,13 @@
 name: Build Tests
 
 on:
-  workflow_dispatch:
-    inputs:
-      tpu-type:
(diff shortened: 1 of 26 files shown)


src/xpk/core/blueprint/blueprint_definitions.py
jamOne-: snake_case pls

SikaGrr: Done! Interesting that the linter didn't complain, just about "== None" => "is None"

jamOne-: yeh, it's pretty bad :(

PR #814, opened 2025-11-06
Server telemetry reporting

scaliby:
# Description
Sends telemetry data to server using a separate subprocess.

*Dear reviewers,*

*We should reflect on whether the additional complexity introduced by this proposal is justified by the ability to provide the best possible user experience. Specifically, this approach ensures immediate XPK exit upon request, regardless of potential network delays while flushing data.*

*For comparison, alternative approaches would involve:*

- *Sending telemetry data synchronously, which risks blocking the XPK process for seconds or even minutes on slow networks.*

- *Sending telemetry with a strict deadline (e.g., 1 second) and dropping the data if transmission cannot be completed within that window.*

*I look forward to your thoughts on these trade-offs.*

# Issue
b/452257096

# Testing
Unit testing, manual testing.

### DIFF
--- a/src/xpk/telemetry_uploader.py
+++ b/src/xpk/telemetry_uploader.py
@@ -1,0 +1,29 @@
+"""
+Copyright 2025 Google LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+     https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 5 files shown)


src/xpk/telemetry_uploader.py
```diff
@@ -0,0 +1,29 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+
+import sys
+import os
+import requests
+import json
+
+file_path = sys.argv[1]
```
SikaGrr: do we need an `if __name__ == '__main__':` guard here?

PR #827, opened 2025-11-13
[slice] Cluster scoped

pajakd:
# Description

The Slice CR is changed from namespaced to cluster-scoped. This PR is to align with this change.

# Issue

# Testing

### DIFF
--- a/slice/internal/util/testing/wrappers.go
+++ b/slice/internal/util/testing/wrappers.go
@@ -292,8 +292,8 @@ func (s *SliceWrapper) OwnerWorkloadAnnotations(ns, name string) *SliceWrapper {
 	if s.Annotations == nil {
 		s.Annotations = make(map[string]string)
 	}
-	s.Annotations["slice.accelerator.gke.io/owner-workload-name"] = name
-	s.Annotations["slice.accelerator.gke.io/owner-workload-namespace"] = ns
+	s.Annotations[core.OwnerWorkloadNameAnnotation] = name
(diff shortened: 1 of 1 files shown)


slice/internal/core/slice.go
```diff
@@ -33,8 +33,11 @@ func SliceKeyFromWorkload(wl *kueue.Workload, podSetName kueue.PodSetReference)
 func SliceWithMetadata(wl *kueue.Workload, podSetName kueue.PodSetReference) *v1alpha1.Slice {
 	return &v1alpha1.Slice{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      SliceName(wl.Name, podSetName),
-			Namespace: wl.Namespace,
+			Name: SliceName(wl.Name, podSetName),
```
mbobrovskyi: What if we have workloads with the same names and PodSet names in different namespaces? Should we include the namespace in the name?

mbobrovskyi (commented):
Previously, we didn’t handle removing slices when a workload was deleted due to ownerReference. I think we should add this logic to the reconciler to handle workload deletion properly.

slice/internal/core/slice.go
mbobrovskyi: Yes, what if the user creates a prebuilt workload?

mbobrovskyi:
> I can't think of a scenario that requires additional cleanup (also, none of our e2e tests is failing with this change). What am I missing?

Before this, we try to get the workload, and if it doesn’t exist, we skip the reconciliation. But probably it shouldn't be a problem because we have finalizer, right?

PR #828, opened 2025-11-13
add placement policy for NAP workload

zxhe-sean:
# Description

Support creating new workload policies in xpk workload create. This enables scheduling of different topologies within the same GKE cluster through NAP which is very important for sharing capacity easily across multiple test users.

# Issue 
https://b.corp.google.com/issues/455642310
NAP previously did not support creating workload policies during workload create and only during cluster create so only just one topology would be supported for that cluster limited NAP to just one topology per cluster AKA making NAP not useful! In that scenario we would just use static clusters.

# Testing
Created a tpu7x nap cluster and created workloads of different sizes and saw nodepools created and workloads running.

### DIFF
--- a/src/xpk/core/scheduling.py
+++ b/src/xpk/core/scheduling.py
@@ -19,6 +19,9 @@ from ..utils.topology import is_topology_valid
 from ..utils.execution_context import is_dry_run
 from .capacity import AUTOPROVISIONING_CONFIG_MAXIMUM_KEY, AUTOPROVISIONING_CONFIG_VALUE
 from .resources import CLUSTER_RESOURCES_CONFIGMAP, get_cluster_configmap
+from .commands import run_command_for_value
+from .gcloud_context import zone_to_region
+
 from .system_characteristics import (
     AcceleratorType,
     AcceleratorTypeToAcceleratorCharacteristics,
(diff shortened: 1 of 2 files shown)


src/xpk/core/scheduling.py
```diff
@@ -306,8 +309,36 @@ def create_sub_slicing_annotations(sub_slicing_topology: str) -> list[str]:
...
+      'Create resource policy',
+  )
+
+  if return_code != 0:
+    raise RuntimeError('Unable to create resource policy')
+
+
+def create_placement_policy_label(system: SystemCharacteristics, args) -> str:
   name = get_placement_policy_name(system)
+  ensure_resource_policy_exists(name, args, system.topology)
```
FIoannides: can you explain a little bit why we need this during the `xpk workload create` command? Is it for the cases where cluster topology and workload topology are different? If yes, I am surprised that this works as it would create incompatible labels on the node and the job. Do you have an example where you tested this? I would like to inspect the cluster and workload there

scaliby:
@Obliviour we're just behind the corner with subsliicing support, so let's hold on on merging this. This will be supported through sub-slicing and should be possible to be used by you tomorrow.

PR #829, opened 2025-11-14
Flush to clearcut synchronously if uploader script cannot be found

scaliby:
# Description
Flush to clearcut in a blocking way (without running background script).
This will address the inability to report data for `python3 xpk.py` users.

# Issue
b/460017018

# Testing
Tested locally.

### DIFF
--- a/src/xpk/core/telemetry.py
+++ b/src/xpk/core/telemetry.py
@@ -61,7 +61,16 @@ def _store_payload_in_temp_file(data: str) -> str:
 
 def _schedule_clearcut_background_flush(
     file_path: str, wait_to_complete: bool
-) -> None:
+) -> bool:
+  """Schedules clearcut background flush.
+
+  Args:
+    file_path: path to the temporary file where the events are stored.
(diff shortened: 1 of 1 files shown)


src/xpk/core/telemetry.py
jamOne-: Is this covered by unit tests?

jamOne- (changes requested):

src/xpk/core/telemetry.py
```diff
@@ -80,6 +95,14 @@ def _flush_temp_file_to_clearcut(
     )
     if wait_to_complete:
       process.wait()
+    return True
+
+
+def _clearcut_flush(file_path: str) -> None:
+  with open(file_path, mode="r", encoding="utf-8") as file:
```
SikaGrr: These three lines are duplicated in telemetry_uploader.py, maybe you can call this function from there.

SikaGrr (approved):

jamOne- (approved):

PR #839, opened 2025-11-17
Fix GPU versioning and version checks

FIoannides:
# Description

This PR ensures:
- We use the rapid channel if no gke-version flag is specified otherwise we use REGULAR
- Valid and default versions are checked against the appropriate channel
- We stop hardcoding version 1.32 for GPU provisioning in CT
- We specify release channel for GPU provisioning in CT to be consistent with the TPU behavior

# Issue

# Testing

Unit tests added, also manually tested

### DIFF
--- a/.github/workflows/integration_basic_cluster_create.yaml
+++ b/.github/workflows/integration_basic_cluster_create.yaml
@@ -84,7 +84,7 @@ jobs:
               run: xpk job ls --cluster $TPU_CLUSTER_NAME --zone=us-central2-b | grep 'xpk-def-app-profile-slurm-'
             - name: Get created job name
               run: |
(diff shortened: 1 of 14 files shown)


src/integration/gcluster_a3mega_test.py
```diff
@@ -28,6 +29,8 @@
 zone = os.getenv("ZONE")
 auth_cidr = os.getenv("AUTH_CIDR")
 cluster_name = os.getenv("A3_MEGA_TEST_CLUSTER_NAME")
+release_channel = os.getenv("RELEASE_CHANNEL")
```
scaliby: please run nightly on your branch as I think this will fail because the value is not provided in nightly env.

scaliby (approved):

PR #843, opened 2025-11-18
Add Autoprovisioning scopes

SujeethJinesh:
# Description
Auto provisioned nodepools do not have the right scopes attached to them when scaling up, resulting in our cluster's nodes being kept in the pending state. This fix ensures that the auto provisioned nodepools have the cloud-platform scope piped correctly. We explicitly do not set `gke-default` or `storage-full` as they're already encompassed by cloud-platform and if we do include them, we see control plane errors related to those scopes not working.

# Issue
b/460847426

# Testing
Manually tested here b/460847426#comment16


scaliby (commented):
We're already specifying scopes [here](https://github.com/AI-Hypercomputer/xpk/blob/94710cd8737f5ab6cbede5f5273ca014c104cd41/src/xpk/core/nap.py#L47). Can we just add this new scope to the list instead of specifying it through the additional flag?

PR #846, opened 2025-11-19
fix: Improve clarity of command execution logs

stony-tark:
# Description
Removes the potentially confusing log message 'hiding output unless there is an error.' This change improves log clarity and prevents false positives when searching for the word 'error'. The new behavior will only print command output upon failure.
Changed the sleep value from 1 to 10 to decrease log pollution.

# Issue
b/462055036

# Testing
I've run xpk workload create to check if this change works

### DIFF
--- a/src/xpk/core/commands.py
+++ b/src/xpk/core/commands.py
@@ -201,6 +201,7 @@ def run_command_with_updates(command, task, verbose=True) -> int:
           xpk_print(f'Task: `{task}` terminated with code `{return_code}`')
           return return_code
   else:
+    xpk_print(f'Task: `{task}` is implemented by `{command}`)
     try:
       subprocess.check_output(command, shell=True, stderr=subprocess.STDOUT)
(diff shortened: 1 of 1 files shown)


src/xpk/core/commands.py
```diff
@@ -195,16 +195,12 @@ def run_command_with_updates(command, task, verbose=True) -> int:
...
-          time.sleep(1)
-          i += 1
+          time.sleep(10)
+          i += 10
         else:
           xpk_print(f'Task: `{task}` terminated with code `{return_code}`')
           return return_code
   else:
-    xpk_print(
-        f'Task: `{task}` is implemented by `{command}`, hiding output unless'
```
SikaGrr: I would only remove `hiding output unless there is an error`. I've talked with some users who troubleshoot problems by running commands manually after they fail, so they need to know what was executed.

SikaGrr (changes requested):

SikaGrr (approved):

PR #873, opened 2025-11-25
Restore SEND_TELEMETRY value after test

scaliby:
# Description
Currently, running unit tests sets send-telemetry config value to false. Because of this we have no data collected after 18th of Nov. This change restores original telemetry reporting preference after test run.

# Issue
n/a

# Testing
unit testing.


jamOne-:
wow, I still cannot get used to the fact that our UTs can interact with the real world

+1 to Ivan's comment

PR #878, opened 2025-11-26
Pathways machine type validation

scaliby:
# Description
Validate pathways machine type, so it is not smaller than 100GB of memory and less than 49CPUs.

# Issue
b/463714129

# Testing
Unit testing.

### DIFF
--- a/src/xpk/core/pathways.py
+++ b/src/xpk/core/pathways.py
@@ -345,8 +345,8 @@ def get_pathways_machine_types(
   min_memory_mb = 233 * 1024
   command = (
       'gcloud compute machine-types list --filter "guestCpus >= 49 AND memoryMb'
-      f' >= {min_memory_mb} AND zone = \'{zone}\'" --format="value(name)" '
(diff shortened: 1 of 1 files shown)


src/xpk/core/pathways.py
```diff
@@ -333,3 +333,19 @@ def try_to_delete_pathwaysjob_first(args, workloads) -> bool:
     xpk_print(f'Delete Workload request returned ERROR {return_code}')
     return False
   return True
+
+
+def get_pathways_machine_types(zone: str) -> tuple[int, list[str]]:
+  min_memory_mb = 233 * 1024
```
FIoannides: can you add a comment please explaining what these numbers mean and how they are derived? also for the guestCpus >= 49 condition below

FIoannides (approved):

jamOne- (approved):

SikaGrr (changes requested):

SikaGrr (approved):

PR #885, opened 2025-12-01
Log event latency

scaliby:
# Description
We need this field to be able to understand the execution time of the command.

# Issue
n/a

# Testing
Unit testing.


SikaGrr (approved):
Can we rename it from LATENCY to DURATION?

PR #899, opened 2025-12-02
feat: Add WorkloadScheduling.SUPER_SLICING_AVAILABLE

jamOne-:
# Description

Add `WorkloadScheduling.SUPER_SLICING_AVAILABLE` so it can be later used during `workload create` to apply a special behavior.

# Issue

http://b/465041125

# Testing

unit tests

### DIFF
--- a/src/xpk/core/scheduling.py
+++ b/src/xpk/core/scheduling.py
@@ -96,10 +96,9 @@ def check_if_workload_can_schedule(
     else:
       return WorkloadScheduling.UNAVAILABLE
 
-  if _check_sub_slicing_availability(
+  if cluster_system and _check_sub_slicing_availability(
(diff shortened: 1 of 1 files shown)


src/xpk/core/scheduling.py
```diff
@@ -107,6 +109,19 @@ def check_if_workload_can_schedule(
     else:
       return WorkloadScheduling.UNAVAILABLE
 
+  if _check_super_slicing_availability(
+      workload_system=workload_system, cluster_system=cluster_system
+  ):
+    assert cluster_system
```
scaliby: when `_check_super_slicing_availability` would not accept `None` then this would not be needed.

jamOne-: Done

PR #906, opened 2025-12-04
Periodic releases script

scaliby:
# Description
- Periodic releases triggered by crontab at 10:30 on Monday.
- Manual approval step before branch pushing (environment it pushes to requires an approval).
- Additionally, ability to trigger manually through workflow_dispatch.

*Note: This assumes some release branch already exists and computes new version on top of the existing release branch, so we need to create first one manually.*

# Issue
b/465710125

# Testing
Tested on my fork https://github.com/scaliby/xpk
- No changes run: https://github.com/scaliby/xpk/actions/runs/19922664228
- Changes run: https://github.com/scaliby/xpk/actions/runs/19924285660
  * which triggered: https://github.com/scaliby/xpk/actions/runs/19924291436
  * which triggered: https://github.com/scaliby/xpk/actions/runs/19924299238

### DIFF
--- a/.github/workflows/periodic_release.yaml
+++ b/.github/workflows/periodic_release.yaml
@@ -1,0 +1,89 @@
+# Copyright 2025 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


scaliby:
> Will we later add the "nightly tests check" to it?

I didn't want to explicitly add dependency on passing nightly, as they are often failing for unrelated reasons. There's an approval step so you can use your best judgment whenever to approve this or not based on the nightly status. Do you think it's better to directly add nightly as a dependency?

jamOne-:
I'm fine with the manual approval for now. We can add the automatic nightly check if we decide to fully automate the releases

PR #911, opened 2025-12-04
Super-slicing cluster create arguments validation

SikaGrr:
# Description
In case of super-slicing cluster creation, validate that:
- reservation is gSC
- reservation is block or sub-block
- the TPU type is tpu7x-4x4x4 or tpu7x-128

# Issue
b/465041782

# Testing
Unit tests.

### DIFF
--- a/src/xpk/commands/cluster.py
+++ b/src/xpk/commands/cluster.py
@@ -244,8 +244,8 @@ def _validate_super_slicing_reservation(args):
   reservation = parse_reservation(args.reservation, args.project)
   if reservation.block_name is None:
     xpk_print(
-        f'Error: Validation failed: Super-slicing cluster creation'
-        f' requires a block or sub-block reservation.'
+        'Error: Validation failed: Super-slicing cluster creation'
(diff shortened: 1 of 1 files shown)


src/xpk/core/capacity.py
```diff
@@ -284,27 +285,59 @@ def get_capacity_node_selectors_from_capacity_type(
...
+      block_name=None,
+      sub_block_name=None,
+  )
+  parts = reservation_path.split('/')
+  if min(map(len, parts)) == 0:  # all parts must be non-empty
+    return None
+  if len(parts) == 1:
+    return reservation  # trivial case
+
+  if parts[0] == 'projects':
```
jamOne-: Could we create a helper that can parse any oneplatformlink to a dict, it could be useful later, e.g. `foo/foo-1/bar/bar-1` would result in `{'foo': 'foo-1', 'bar': 'bar-1'}`. Although, it's a bit less strict, as it doesn't enforce the order.

scaliby (approved):

PR #912, opened 2025-12-05
Fix XPK freezing

SikaGrr:
# Description
Reported via chat, some users just see version printed out, then XPK hangs.
It will continue after pressing ENTER, but not everybody notices this.
Fixed by adding `--no-input`.

# Issue
b/466129455

# Testing
Manual testing.


SikaGrr:
rebased PR correctly, only version check is removed now

PR #914, opened 2025-12-05
Add --num-cubes for superslicing cluster create

SikaGrr:
# Description
Add a --num-cubes flag as an alias for --num-slices in case of super-slicing cluster create.
If the feature is enabled, then move num_slices default value setting from parser to validation.

Alternatives considered: add an alias in the parser directly. Doesn't work because we allow it only if --super-slicing. Also, it would hide the alias in the usage example.

# Issue
b/465042527

# Testing
Unit tests and goldens.


scaliby (approved):

jamOne- (approved):
nice

PR #938, opened 2025-12-12
Fix dry run

scaliby:
# Description
The current logic makes us write file in a directory that does not exist. Fixed that.

# Issue
n/a

# Testing
golden

### DIFF
--- a/src/xpk/core/config.py
+++ b/src/xpk/core/config.py
@@ -19,6 +19,7 @@ import os
 import ruamel.yaml
 from abc import ABC, abstractmethod
 from ..utils import file
+from ..utils.execution_context import is_dry_run
 from ..utils.console import xpk_print
 from setuptools_scm import get_version as setuptools_get_version
 from importlib.metadata import version, PackageNotFoundError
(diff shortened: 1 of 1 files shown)


FIoannides:
> > After your change we'll create a dir in dry_run, correct? Why? I believe dry_run shouldn't have any side-effects/modify anything.
> 
> This is right - we will create a dir. This is because in dry-run mode `file.ensure_directory_exists()` is a no-op which causes `config.py` to always throw error when property is set:
> 
> ```
>   def _open_configs(self) -> dict | None:
>     file.ensure_directory_exists(os.path.dirname(self._config)) # no-op
> 
>     if not os.path.exists(self._config):
>       return None # returning here as file does not exist
> 
>     with open(self._config, encoding='utf-8', mode='r') as stream:
>       config_yaml: dict = yaml.load(stream)
>       return config_yaml
> 
>   def _save_configs(self, config_yaml: dict) -> None:
>     with open(self._config, encoding='utf-8', mode='w') as stream: # throw here as parent doesn't exist
>       yaml.dump(config_yaml, stream)
> 
>   def set(self, key: str, value: str | None) -> None:
>     if key not in self._allowed_keys:
>       xpk_print(f'Key {key} is not an allowed xpk config key.')
>       return
> 
>     config_yaml = self._open_configs()
>     if config_yaml is None:
>       config_yaml = {'version': 'v1', CONFIGS_KEY: {}}
> 
>     config_yaml[CONFIGS_KEY][key] = value
>     self._save_configs(config_yaml) # throw here
> ```
> 
> What are your other ideas to handle this case @jamOne-? I think it would be a good idea to have config support in dry-run.

I agree with @jamOne- on this one, dry-run shouldn't create anything permanent, even if here it seems harmless, it sets a dangerous precedence. Can we just have an if dry_run: early exit from the _open_configs function?

PR #953, opened 2025-12-19
fix DNS credentials retry

FIoannides:
# Description

Retry get credentials without dns-endpoint flag

# Issue

https://github.com/AI-Hypercomputer/xpk/actions/runs/20361450149/job/58507441651

# Testing


scaliby (approved):

jamOne- (changes requested):
Please add a unit test for this, this function is already covered with the tests

PR #955, opened 2025-12-19
Add unit test for get credentials

FIoannides:
# Description

Handling comment on https://github.com/AI-Hypercomputer/xpk/pull/953

# Issue

# Testing


src/xpk/core/cluster_test.py
```diff
@@ -41,11 +41,15 @@ def command_args(mocker: MockerFixture):
   return mocker.Mock(cluster="cluster", project="project", zone="zone")
 
 
-def test_get_cluster_credentials_returns_1_when_retrieval_command_fails(
+def test_get_cluster_credentials_retries_when_first_call_fails(
```
jamOne-: nit: it doesn't actually test the re-try mechanism, but returning the 1 code. I'd stay with the old name, WDYT?

FIoannides: reverted to a slightly different old name

PR #963, opened 2026-01-08
Stamp telemetry trash execution

scaliby:
# Description
This change modifies calculation of the `XPK_TESTER` flag in telemetry to set it to `true` when the `TELEMETRY_TRASH_EXECUTION` environment variable is set to `true`.

I will globally set this flag for all XPK's github action executions on repository level. It will allow us to easily filter out GH automated runs from telemetry data.

**Why not rely on `os.getenv("GITHUB_ACTIONS") == "true"`?**

While this will work for our use case, it will also exclude all data, in case XPK is run by a 3rd party within github actions context. Such executions are not trash and should not be treated as a test executions but actual production use cases.

# Issue
n/a

# Testing
Unit testing

### DIFF
--- a/src/xpk/core/telemetry_test.py
+++ b/src/xpk/core/telemetry_test.py
@@ -241,17 +241,18 @@ def test_metrics_collectors_logs_correct_tester_value_for_is_tester_variable(
 
 
 @pytest.mark.parametrize(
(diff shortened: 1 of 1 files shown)


jamOne- (approved):
"trash execution" -- brutal...

PR #968, opened 2026-01-14
[slice] Make slice activation timeout configurable

j-skiba:
# Description
Allow to configure slice activation timeout using `activation-timeout` flag. The value defaults to 3m.

# Issue
---

# Testing
Performed manual tests


gabesaba:
Thank you!

lgtm

PR #992, opened 2026-01-20
[slice] More topology validation

pajakd:
# Description

Add more validation to the topology string. In particular, we were not checking for a topology being zero (which resulted in panics). 

# Issue

# Testing


PBundyra:
LGTM

PR #994, opened 2026-01-20
Migrate to recipes

scaliby:
# Description
Drop-in replacement of goldens with recipes.

*#antigravity*

# Issue
b/476991735 + b/476968048

# Testing
CI/CD + manual testing on my local.


SikaGrr (approved):
We could also remove --dry_run from all commands now, right?

PR #1001, opened 2026-01-20
Fix Super-slicing nume aware conflict

jamOne-:
# Description

Numa aware workloads don't work for super-slicing. Currently they are enabled whenever a system supports it.

# Issue

http://b/477121903

# Testing

Goldens


jamOne-:
I tested it, but it seems not sufficient enough:
- This numa aware conflict is occurring only when using a real workload, requiring TPUs. I was just testing `fake_training.py` with different slice sizes and that was successful.
- Block validation removal (that was in one of the previous PRs) was something completely unexpected for me, breaking my assumptions.

PR #1006, opened 2026-01-22
[slice] Fix injection of an empty TPU topology node selector

j-skiba:
# Description
Fixed injection of an empty TPU topology node selector for replicated jobs that don't use TPUs.

# Testing
tested manually and added a separate test case


j-skiba:
It seems to be unrelated. There might be some rare problem with workload admission check status update. I will try to find the cause and open a separate PR.

PR #1019, opened 2026-02-04
Record used command flags in telemetry

scaliby:
# Description
Record used command flags in telemetry.

# Issue
b/480866802

# Testing
unit testing

### DIFF
--- a/src/xpk/parser/common.py
+++ b/src/xpk/parser/common.py
@@ -265,8 +265,8 @@ def enable_flags_usage_tracking(
           setattr(namespace, dest_attr, set())
 
         if self.option_strings:
-          option_string = max(self.option_strings, key=len)
-          getattr(namespace, dest_attr).add(option_string)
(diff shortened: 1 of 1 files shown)


src/xpk/parser/common.py
```diff
@@ -244,3 +244,56 @@ def _get_path_segments(current_parser):
     return [chosen_command]
 
   return ' '.join(_get_path_segments(parser))
+
+
+def enable_flags_usage_tracking(
+    parser: argparse.ArgumentParser, dest_attr: str = '_supplied_flags'
```
jamOne-: nit: I'd prefer to remove dest_attr parameter, as we're never going to change it and just declare some CONST: `_SUPPLIED_FLAGS_ATTR = '_supplied_flags'`

scaliby: I moved it to a const, but still prefer to leave the `dest_attr` attrubute as it provides greater flexibility.

SikaGrr (approved):

PR #1035, opened 2026-02-06
Fix pylint

jamOne-:
# Description

#### Gemini:

Resolved the `inconsistent-quotes` linting errors by disabling the `check-quote-consistency` check in `.pylintrc`. This avoids conflicts between `pylint` and `pyink` (which enforces double quotes) without cluttering the code with extra variables or escaping.

# Issue

# Testing

`make verify`


scaliby (commented):
Why? I like quotes to be consistent. Having warnings now seems to be a temporal inconvenience that will profit in the future by having consistent quotes.

scaliby (approved):

PR #1042, opened 2026-02-10
fix a crash when running just `xpk`

SikaGrr:
# Description
fix a crash when running just `xpk`

# Testing
Manual testing.


SikaGrr:
> It is not the first time this is broken. Can we cover this with golden?

Not trivial, since golden run appends `--dry-run`, which is not what we want to test.
We'd have this output:
```
<!--
$ xpk
usage: xpk [-h] {workload,storage,cluster,inspector,info,version,config} ...
xpk: error: unrecognized arguments: --dry-run
-->
```

PR #1082, opened 2026-02-27
Fix b/485965692: Support custom topologies in superslicing

jamOne-:
# Description

The requirements changed and we need to lift the hardcoded maximum dimensions (16x24x24) for super-slicing topology validation and instead use a max cube count limit (144).

# Issue

http://b/485965692

# Testing

unit tests


scaliby (approved):
I was actually vibe-coding this now :D Reassigning ticket to you.

PR #1091, opened 2026-03-03
Add script to compute dependencies checksums

scaliby:
# Description
Computes checksums for specified system dependencies.

# Issue
b/489308867

# Testing
Tested manually.

### DIFF
--- a/src/xpk/utils/dependencies/binary_dependencies.py
+++ b/src/xpk/utils/dependencies/binary_dependencies.py
@@ -21,6 +21,8 @@ import enum
 
 @dataclasses.dataclass(frozen=True)
 class BinaryDependency:
+  """Represents a binary dependency with its metadata and checksums."""
+
   archive_type: str
   binary_name: str
   checksums: dict[str, str]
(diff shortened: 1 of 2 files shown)


tools/update_dependencies.py
```diff
@@ -0,0 +1,139 @@
...
+    SCRIPT_DIR.parent
+    / "src"
+    / "xpk"
+    / "utils"
+    / "dependencies"
+    / "checksums.py"
+)
+
+
+def get_hash(url: str) -> str | None:
```
jamOne-: nit: I don't think it ever returns None?

src/xpk/utils/dependencies/binary_dependencies.py
```diff
@@ -0,0 +1,97 @@
...
+import enum
+
+
+@dataclasses.dataclass(frozen=True)
+class BinaryDependency:
+  """Represents a binary dependency with its metadata and checksums."""
+
+  archive_type: str
+  binary_name: str
+  checksums: dict[str, str]
```
jamOne-: nit: This too could be some kind of enum / dict[Env, str].

PR #1096, opened 2026-03-05
Ignore nodepool creation errors

scaliby:
# Description
Ignore nodepool creation errors when `--ignore-nodepool-creation-errors` flag is supplied.

# Issue
b/485645043

# Testing
Unit + manual.

### DIFF
--- a/src/xpk/core/testing/commands_tester_test.py
+++ b/src/xpk/core/testing/commands_tester_test.py
@@ -56,7 +56,7 @@ def test_run_command_batch_default_result(
       output_logs=["log1", "log2"],
   )
 
-  assert result == []
+  assert not result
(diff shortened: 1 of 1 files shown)


src/xpk/parser/cluster.py
```diff
@@ -731,6 +731,14 @@ def add_shared_cluster_create_optional_arguments(
       action='store_true',
       help='Enable Workload Identity Federation on the cluster and node-pools.',
   )
+  parser_or_group.add_argument(
```
jamOne-: Q: Will this work for `cluster adapt`?

scaliby: Registered this in `cluster adapt` as well.

PR #1097, opened 2026-03-05
feat: add flag to disable DAD

scaliby:
# Description
Adds flag to disable DAD. We will make DAD behavior a default, but there should be a possibility to download it.

# Issue
n/a

# Testing
Unit testing.

### DIFF
--- a/src/xpk/parser/common_test.py
+++ b/src/xpk/parser/common_test.py
@@ -15,7 +15,6 @@ limitations under the License.
 """
 
 import argparse
-from unittest.mock import patch
 from .common import extract_command_path, enable_flags_usage_tracking, retrieve_flags, add_shared_arguments, FeatureFlags
 from .core import set_parser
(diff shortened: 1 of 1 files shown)


src/xpk/utils/validation.py
```diff
@@ -52,11 +52,16 @@ def should_validate_dependencies(args):
   return not skip_validation and not dry_run
 
 
-def validate_dependencies_list(dependencies: list[SystemDependency]):
+def validate_dependencies_list(args, dependencies: list[SystemDependency]):
   """Validates a list of system dependencies and returns none or exits with error."""
   for dependency in dependencies:
+    auto_download = (
+        not hasattr(args, 'dependency_auto_download')
```
jamOne-: Do we need to be so cautious? Why not just `args.dependency_auto_download`? It should be included in every args?

PR #1121, opened 2026-03-13
[slice] Use retry on failure mechanism

pajakd:
# Description

Adding a featuregate UseRetryMechanismForSliceCreation that Annotates each workload to enable the mechanism.

Additionally, we tread Stale slices (slices that fail to become Ready within the timout) as Failed and we don't check for partition id conflicts upon slice creation (we will rely on the retry mechanism handling that).

# Issue

# Testing


j-skiba:
lgtm

PR #1146, opened 2026-03-19
Implement NativeCommandRunner

scaliby:
# Description
Implements native cluster toolkit execution. No docker needed!

# Issue
b/493833231

# Testing
manual testing on my local


src/xpk/core/cluster_toolkit/native_manager.py
jamOne-: Could you cover the public methods with basic unit tests please?

PR #1160, opened 2026-03-25
[slice] Handle long-named slices after slice-controller upgrade

j-skiba:
# Description
Add support for handling slices with long names that might exist in the cluster after upgrade of slice-controller when `ShorterSliceNameLength` feature gate is enabled.

# Issue

# Testing


j-skiba:
cc @mwielgus @pajakd

slice/internal/controller/workload_controller_test.go
```diff
@@ -2232,12 +2233,65 @@ func TestWorkloadReconciler(t *testing.T) {
 					Obj(),
 			},
 		},
+		"should create Slices with shorter name when feature gate enabled": {
```
mwielgus: What about a test where there is an old slice with the long name, but now someone enabled short names?

PR #1165, opened 2026-03-30
[slice] Default log level

pajakd:
# Description

We don't have any default log level. This means that without manual modification no logs are exported. In [Kueue we set this to 2](https://github.com/kubernetes-sigs/kueue/blob/66c2cf1a93ee3f19d8b29309d27832a4595d4106/config/components/manager/manager.yaml#L26) but here (I think) 3 would be a good default.

# Issue

# Testing


mwielgus (approved):
/lgtm

PR #1169, opened 2026-04-03
[slice] LWS support

pajakd:
# Description

Since for LWS, the Annotations are not propagated to PodSets, we introduce an alternative mechanism of fetching the topology from NodeSelector (or NodeAffinity). This is only if the Annotation is missing. Also its behind a featuregate "TopologyFromNodeSelectorOrAffinity", which is **true by default**. This means that currently LWS will work only with subslicing (maximum topology 4x4x4). 

# Issue

# Testing


slice/go.mod
```diff
@@ -5,17 +5,18 @@ go 1.25.0
 require (
 	github.com/go-logr/logr v1.4.3
 	github.com/google/go-cmp v0.7.0
-	github.com/onsi/ginkgo/v2 v2.27.2
-	github.com/onsi/gomega v1.38.2
+	github.com/onsi/ginkgo/v2 v2.27.5
+	github.com/onsi/gomega v1.39.0
```
mwielgus: Is it strictly required in this PR?

slice/internal/util/workload/workload.go
```diff
@@ -64,3 +64,23 @@ func IsJobOwner(wl *kueue.Workload) bool {
...
+func IsLeaderWorkerSetOwner(wl *kueue.Workload) bool {
+	if owner := GetOwner(wl); owner != nil {
+		return owner.APIVersion == leaderworkersetv1.SchemeGroupVersion.String() && owner.Kind == "LeaderWorkerSet"
+	}
+	return false
+}
+
+func GetOwner(wl *kueue.Workload) *metav1.OwnerReference {
+	if owner := metav1.GetControllerOf(wl); owner != nil {
+		return owner
```
mwielgus, line 77: It seems inconsistent with the below lines which validate the owner kind and return nil otherwise. Why so?

slice/internal/webhooks/leaderworkerset_webhook.go
```diff
@@ -0,0 +1,99 @@
...
+		log.V(5).Info("Skipping due to missing Kueue Label")
+		return nil
+	}
+
+	if !core.IsRelevantPodTemplateSpec(lws.Spec.LeaderWorkerTemplate.WorkerTemplate) {
+		log.V(5).Info("Skipping annotating WorkerTemplate due to TPU Annotation or Node Selector misconfigured at worker template")
+		return nil
+	}
+	if lws.Spec.LeaderWorkerTemplate.LeaderTemplate != nil && !core.IsRelevantPodTemplateSpec(*lws.Spec.LeaderWorkerTemplate.LeaderTemplate) {
+		log.V(5).Info("Skipping annotating WorkerTemplate due to TPU Annotation or Node Selector misconfigured at leader template")
```
mwielgus: This sounds weird. When can it happen and why is it not an error?

slice/internal/webhooks/leaderworkerset_webhook.go
```diff
@@ -0,0 +1,99 @@
...
+	if lws.Spec.LeaderWorkerTemplate.WorkerTemplate.Annotations == nil {
+		lws.Spec.LeaderWorkerTemplate.WorkerTemplate.Annotations = make(map[string]string)
+	}
+	if lws.Spec.LeaderWorkerTemplate.LeaderTemplate != nil {
+		// if the leader is defined, annotate both templates with the same group name
+		lws.Spec.LeaderWorkerTemplate.WorkerTemplate.Annotations[podsetGroupName] = podsetGroupValue
+	}
+	lws.Spec.LeaderWorkerTemplate.WorkerTemplate.Annotations[kueue.PodSetRequiredTopologyAnnotation] = parsed.RequiredSliceLevel()
+
+	if lws.Spec.LeaderWorkerTemplate.LeaderTemplate != nil {
```
mwielgus: Duplicated lws.Spec.LeaderWorkerTemplate.LeaderTemplate != nil  if statement. If this is intended, combine two ifs.

slice/internal/webhooks/leaderworkerset_webhook.go
```diff
@@ -0,0 +1,99 @@
...
+		return err
+	}
+	annotatePodTemplateSpecWithSliceHealth(&lws.Spec.LeaderWorkerTemplate.WorkerTemplate, parsed, r.DefaultSliceHealthValues)
+	if lws.Spec.LeaderWorkerTemplate.WorkerTemplate.Annotations == nil {
+		lws.Spec.LeaderWorkerTemplate.WorkerTemplate.Annotations = make(map[string]string)
+	}
+	if lws.Spec.LeaderWorkerTemplate.LeaderTemplate != nil {
+		// if the leader is defined, annotate both templates with the same group name
+		lws.Spec.LeaderWorkerTemplate.WorkerTemplate.Annotations[podsetGroupName] = podsetGroupValue
+	}
```
PBundyra: Is this if needed? Here we set the annotation for the WorkerTemplate, and below there is the same if with annotation for Leader

slice/internal/controller/workload_controller.go
```diff
@@ -577,6 +624,9 @@ func (r *WorkloadReconciler) syncSlices(
 }
 
 func shouldCreateSlicesForPodSetAssignment(wl *kueue.Workload, psa kueue.PodSetAssignment, nodes map[string]corev1.Node) bool {
+	if utilworkload.IsLeaderWorkerSetOwner(wl) && psa.Name == "leader" {
```
mwielgus: Can we use some constant, preferably from LWS API, instead of hardcoded string?

slice/internal/topology/validation.go
```diff
@@ -41,6 +42,10 @@ func AllAssignmentsValid(wl *kueue.Workload, nodes map[string]corev1.Node) bool
 		if psa.TopologyAssignment == nil {
 			continue
 		}
+		// we assume that leader will be co-located with workers and will use the same slice
+		if utilworkload.IsLeaderWorkerSetOwner(wl) && psa.Name == "leader" {
```
mwielgus: Same, can we use constant here?

slice/internal/topology/validation.go
```diff
@@ -41,6 +42,10 @@ func AllAssignmentsValid(wl *kueue.Workload, nodes map[string]corev1.Node) bool
...
+			continue
```
mwielgus, line 47: Validation is skipped for the "leader" PodSet under the assumption it will be co-located with the workers on the same slice. However, if the leader's PodTemplate mistakenly requests core.TPUResourceName limits/requests, those TPUs will not be validated or properly managed by the slice controller. Should there be a sanity check here ensuring the leader template doesn't request TPUs before skipping validation?

slice/internal/core/core.go
```diff
@@ -50,7 +52,50 @@ func IsRelevantPodTemplateSpec(spec corev1.PodTemplateSpec) bool {
 }
 
 func GetTPUTopology(spec corev1.PodTemplateSpec) string {
-	return spec.Annotations[TPUSliceTopologyAnnotation]
+	if val, ok := spec.Annotations[TPUSliceTopologyAnnotation]; ok && val != "" {
+		return val
+	}
+	if features.Enabled(features.TopologyFromNodeSelectorOrAffinity) {
+		if val := GetTPUTopologyFromSelectorOrAffinity(spec); val != "" {
+			return val
```
mwielgus, line 60: Can we log the values extracted and actual conditions used?

slice/go.mod
pajakd: Removed.

slice/internal/webhooks/leaderworkerset_webhook.go
pajakd: Grouping is needed only if both leaders and workers are specified. 

Second if is not needed -- merged both under one if.

slice/internal/webhooks/leaderworkerset_webhook.go
pajakd: Fixed.

slice/internal/controller/workload_controller.go
pajakd: Done.

slice/internal/topology/validation.go
pajakd: Done.

slice/internal/util/workload/workload.go
pajakd: Kueue creates a separate Workload for each replica of a LeaderWorkerSet. Because of this Kueue adds the LeaderWorkerSet to the Workload's OwnerReferences list, but does not set Controller: true for it

slice/internal/webhooks/leaderworkerset_webhook.go
pajakd: I did it based on the High Level Design, where the annotations were both on the Worker and Leader templates.

slice/internal/topology/validation.go
pajakd: We can add such validation in the followup. But for sure we cannot create a separate slice for leader pod, right? We have to rely on the fact that Kueue will schedule it on the same nodes as the workers.

slice/internal/core/core.go
pajakd: Added logging of the extracted topology.

slice/internal/topology/validation.go
pajakd: It should work even if the leader requests TPUs. The way it works is: 
1. the leaders and workers are co-located within the same `required` domain  
2. we create a Slice only for the worker podset -- the leader pods will use the same Slice
Added a few e2e test for this.

PR #1172, opened 2026-04-08
[slice] Refactor OwnerPodsFinished

pajakd:
# Description

Instead of having 3 functions jobSetPodsFinished, jobPodsFinished and lwsPodsFinished lets have a single one that receives the owner kind and reference as an argument.

# Issue

# Testing

### DIFF
--- a/slice/internal/controller/workload_controller.go
+++ b/slice/internal/controller/workload_controller.go
@@ -373,121 +373,66 @@ func (r *WorkloadReconciler) deleteSlicesForEvictedWorkload(ctx context.Context,
 }
 
 func (r *WorkloadReconciler) ownerPodsFinished(ctx context.Context, wl *kueue.Workload) (bool, error) {
+	var ownerKind, logKey, podLabelKey string
+	var ownerObj client.Object
+	var owner *metav1.OwnerReference
+
 	if utilworkload.IsJobSetOwner(wl) {
-		return r.jobSetPodsFinished(ctx, wl)
-	}
-	if utilworkload.IsJobOwner(wl) {
-		return r.jobPodsFinished(ctx, wl)
-	}
-	if utilworkload.IsLeaderWorkerSetOwner(wl) {
-		return r.lwsPodsFinished(ctx, wl)
-	}
-	// Finalize Workloads that have no owner or have unsupported owner types.
-	return true, nil
-}
-
-func (r *WorkloadReconciler) jobSetPodsFinished(ctx context.Context, wl *kueue.Workload) (bool, error) {
-	owner := metav1.GetControllerOf(wl)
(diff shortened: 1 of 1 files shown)


slice/internal/controller/workload_controller.go
```diff
@@ -373,121 +373,66 @@ func (r *WorkloadReconciler) deleteSlicesForEvictedWorkload(ctx context.Context,
...
-	}
-	if err := r.client.List(ctx, pods, opts...); err != nil {
-		log.Error(err, "Failed to get Pods")
-		return false, err
+	if utilworkload.IsJobSetOwner(wl) {
+		owner = metav1.GetControllerOf(wl)
+		ownerObj = &jobset.JobSet{}
+		podLabelKey = jobset.JobSetNameKey
+		logKey = "jobSet"
+		ownerKind = "JobSet"
```
PBundyra: Can we use a constants from the JobSet API?

slice/internal/controller/workload_controller.go
```diff
@@ -373,121 +373,66 @@ func (r *WorkloadReconciler) deleteSlicesForEvictedWorkload(ctx context.Context,
...
+	} else if utilworkload.IsJobOwner(wl) {
+		owner = metav1.GetControllerOf(wl)
+		ownerObj = &batchv1.Job{}
+		podLabelKey = "batch.kubernetes.io/job-name"
```
PBundyra: Can we use a constant here?

slice/internal/controller/workload_controller.go
```diff
@@ -373,121 +373,66 @@ func (r *WorkloadReconciler) deleteSlicesForEvictedWorkload(ctx context.Context,
...
 
-	pods := &corev1.PodList{}
-	opts := []client.ListOption{
-		client.InNamespace(wl.Namespace),
-		client.MatchingLabels{jobset.JobSetNameKey: owner.Name},
-	}
-	if err := r.client.List(ctx, pods, opts...); err != nil {
-		log.Error(err, "Failed to get Pods")
-		return false, err
+	if utilworkload.IsJobSetOwner(wl) {
```
PBundyra: It just a nit but I believe in Go it's a good practice to use switch-case statements in such cases

PBundyra:
Besides just a few nits, LGTM

slice/internal/controller/workload_controller.go
pajakd: Right, fixed.

slice/internal/controller/workload_controller.go
pajakd: Done

slice/internal/controller/workload_controller.go
pajakd: I was able to get rid of both of these constants since, the OwnerReference has field Kind that holds exactly this.

PR #1173, opened 2026-04-10
[slice] Skip TPU warden check for LWS

pajakd:
# Description

In LWS there is no suspend so (gated) pods are created immediately, before Kueue admits the workload so before the slice is created. 

If these pods don’t have the cloud.google.com/gke-tpu-topology node selector, they get rejected by the GKE warden and if they do have, they cannot get scheduled (because since the slice is not there yet, the nodes don’t have the topology labels). 

This PR adds an Annotation to skip the GKE warden TPU webhook check to allow these pods to be created.

# Issue

# Testing

### DIFF
--- a/slice/internal/webhooks/leaderworkerset_webhook.go
+++ b/slice/internal/webhooks/leaderworkerset_webhook.go
@@ -78,15 +78,17 @@ func (r *LeaderWorkerSetWebhook) Default(ctx context.Context, lws *leaderworkers
 	if lws.Spec.LeaderWorkerTemplate.WorkerTemplate.Annotations == nil {
 		lws.Spec.LeaderWorkerTemplate.WorkerTemplate.Annotations = make(map[string]string)
 	}
+	lws.Spec.LeaderWorkerTemplate.WorkerTemplate.Annotations[core.TPUSkipWardenAnnotation] = "true"
 	if lws.Spec.LeaderWorkerTemplate.LeaderTemplate != nil {
-		// if the leader is defined, annotate both templates with the same group name
-		lws.Spec.LeaderWorkerTemplate.WorkerTemplate.Annotations[podsetGroupName] = podsetGroupValue
-		lws.Spec.LeaderWorkerTemplate.LeaderTemplate.Annotations[podsetGroupName] = podsetGroupValue
-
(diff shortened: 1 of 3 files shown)


slice/internal/webhooks/leaderworkerset_webhook.go
```diff
@@ -78,15 +78,17 @@ func (r *LeaderWorkerSetWebhook) Default(ctx context.Context, lws *leaderworkers
...
 		lws.Spec.LeaderWorkerTemplate.WorkerTemplate.Annotations[podsetGroupName] = podsetGroupValue
 		lws.Spec.LeaderWorkerTemplate.LeaderTemplate.Annotations[podsetGroupName] = podsetGroupValue
+		lws.Spec.LeaderWorkerTemplate.LeaderTemplate.Annotations[core.TPUSkipWardenAnnotation] = "true"
 
 		annotatePodTemplateSpecWithSliceHealth(lws.Spec.LeaderWorkerTemplate.LeaderTemplate, parsed, r.DefaultSliceHealthValues)
-		if lws.Spec.LeaderWorkerTemplate.LeaderTemplate.Annotations == nil {
-			lws.Spec.LeaderWorkerTemplate.LeaderTemplate.Annotations = make(map[string]string)
-		}
 		lws.Spec.LeaderWorkerTemplate.LeaderTemplate.Annotations[kueue.PodSetRequiredTopologyAnnotation] = parsed.RequiredSliceLevel()
 	}
```
gabesaba: I'm happy to leave this as a follow-up, but what do you think about breaking code into functions, perhaps the following?

```
annotateWorkerTemplate(...)
annotateLeaderTemplate(...)
conditionallySetGroupName(...)
```

gabesaba (commented):
lgtm

slice/internal/webhooks/leaderworkerset_webhook.go
gabesaba: (the last function name is just to capture functionality, there can probably be a much better name for it :) )

pajakd: Refactored. Actually `annotateWorkerTemplate` and `annotateLeaderTemplate` could be captured with a single function. PTAL

gabesaba: nice!

gabesaba:
lgtm

PR #1174, opened 2026-04-15
[slice] Change default node affinity for workloads to HEALTHY or DEGRADED health status

j-skiba:
change default node affinity to HEALTHY or DEGRADED when node selectors or node affinity is not defined for a workload.

### DIFF
--- a/slice/cmd/main.go
+++ b/slice/cmd/main.go
@@ -101,7 +101,7 @@ func main() {
 	flag.StringVar(&metricsCertKey, "metrics-cert-key", "tls.key", "The name of the metrics server key file.")
 	flag.BoolVar(&enableHTTP2, "enable-http2", false,
 		"If set, HTTP/2 will be enabled for the metrics and webhook servers")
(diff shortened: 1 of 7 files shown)


slice/cmd/main.go
```diff
@@ -101,7 +101,7 @@ func main() {
 	flag.StringVar(&metricsCertKey, "metrics-cert-key", "tls.key", "The name of the metrics server key file.")
 	flag.BoolVar(&enableHTTP2, "enable-http2", false,
 		"If set, HTTP/2 will be enabled for the metrics and webhook servers")
-	flag.StringVar(&sliceHealthNodeAffinityMode, "default-slice-health-node-affinity", "HEALTHY",
+	flag.StringVar(&sliceHealthNodeAffinityMode, "default-slice-health-node-affinity", "HEALTHY_AND_DEGRADED",
```
mwielgus, line 104: Please also set it explicitly in the manifest, so it is more visible.

j-skiba: config manifest updated

PR #1176, opened 2026-04-15
[slice] NodeSelector in STS webhook

pajakd:
# Description

Add a StatefulSet webhook, that (for relevant tpuv7 pods) makes sure that the topology NodeSelector is present. 

If the selector is not there, it adds it with the value taken from the "cloud.google.com/gke-tpu-topology" annotation.

This is to make sure that LWS workloads are not blocked by the GKE warden.

# Issue

# Testing

### DIFF
--- a/slice/internal/util/testingjobs/leaderworkerset/wrappers.go
+++ b/slice/internal/util/testingjobs/leaderworkerset/wrappers.go
@@ -182,3 +182,72 @@ func (w *Wrapper) LeaderNodeAffinity(key string, values []string) *Wrapper {
 	)
 	return w
 }
+
+func (w *Wrapper) StartupPolicy(policy leaderworkersetv1.StartupPolicyType) *Wrapper {
+	w.Spec.StartupPolicy = policy
+	return w
+}
+
+func (w *Wrapper) WorkerName(name string) *Wrapper {
+	if len(w.Spec.LeaderWorkerTemplate.WorkerTemplate.Spec.Containers) == 0 {
+		w.Spec.LeaderWorkerTemplate.WorkerTemplate.Spec.Containers = []corev1.Container{{}}
+	}
+	w.Spec.LeaderWorkerTemplate.WorkerTemplate.Spec.Containers[0].Name = name
+	return w
+}
+
+func (w *Wrapper) LeaderName(name string) *Wrapper {
+	if w.Spec.LeaderWorkerTemplate.LeaderTemplate == nil {
+		w.Spec.LeaderWorkerTemplate.LeaderTemplate = &corev1.PodTemplateSpec{}
+	}
+	if len(w.Spec.LeaderWorkerTemplate.LeaderTemplate.Spec.Containers) == 0 {
+		w.Spec.LeaderWorkerTemplate.LeaderTemplate.Spec.Containers = []corev1.Container{{}}
+	}
+	w.Spec.LeaderWorkerTemplate.LeaderTemplate.Spec.Containers[0].Name = name
+	return w
+}
+
+func (w *Wrapper) LeaderImage(img string) *Wrapper {
+	if w.Spec.LeaderWorkerTemplate.LeaderTemplate == nil {
+		w.Spec.LeaderWorkerTemplate.LeaderTemplate = &corev1.PodTemplateSpec{}
+	}
+	if len(w.Spec.LeaderWorkerTemplate.LeaderTemplate.Spec.Containers) == 0 {
+		w.Spec.LeaderWorkerTemplate.LeaderTemplate.Spec.Containers = []corev1.Container{{}}
+	}
+	w.Spec.LeaderWorkerTemplate.LeaderTemplate.Spec.Containers[0].Image = img
+	return w
+}
+
+func (w *Wrapper) LeaderArgs(args ...string) *Wrapper {
+	if w.Spec.LeaderWorkerTemplate.LeaderTemplate == nil {
+		w.Spec.LeaderWorkerTemplate.LeaderTemplate = &corev1.PodTemplateSpec{}
+	}
+	if len(w.Spec.LeaderWorkerTemplate.LeaderTemplate.Spec.Containers) == 0 {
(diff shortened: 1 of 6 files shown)


slice/internal/webhooks/statefulset_webhook.go
```diff
@@ -0,0 +1,59 @@
...
+		log.V(5).Info("Skipping non-TPUv7 StatefulSet")
+		return nil
+	}
+
+	tpuTopology := core.GetTPUTopology(sts.Spec.Template)
+	if sts.Spec.Template.Spec.NodeSelector == nil {
+		sts.Spec.Template.Spec.NodeSelector = make(map[string]string)
+	}
+	if _, ok := sts.Spec.Template.Spec.NodeSelector[core.TPUTopologyAnnotation]; !ok {
+		sts.Spec.Template.Spec.NodeSelector[core.TPUTopologyAnnotation] = tpuTopology
```
mwielgus, line 55: What about health selector?

slice/internal/webhooks/statefulset_webhook.go
```diff
@@ -0,0 +1,59 @@
...
+		Complete()
+}
+
+// +kubebuilder:webhook:path=/mutate-apps-v1-statefulset,mutating=true,failurePolicy=fail,sideEffects=None,groups=apps,resources=statefulsets,verbs=create,versions=v1,name=mstatefulset.slice-controller.kb.io,admissionReviewVersions=v1
+
+var _ admission.Defaulter[*appsv1.StatefulSet] = &StatefulSetWebhook{}
+
+func (r *StatefulSetWebhook) Default(ctx context.Context, sts *appsv1.StatefulSet) error {
+	log := ctrl.LoggerFrom(ctx).WithName("statefulset-webhook")
+	log.V(5).Info("Defaulting StatefulSet")
```
mwielgus: Maybe move it below the IF.

slice/internal/webhooks/statefulset_webhook.go
```diff
@@ -0,0 +1,59 @@
...
+
+	if !core.IsRelevantPodTemplateSpec(sts.Spec.Template) {
+		log.V(5).Info("Skipping non-TPUv7 StatefulSet")
```
mwielgus: This log is not future proof. We may have TPUv8, v9 etc soon.

slice/internal/util/testingjobs/leaderworkerset/wrappers.go
```diff
@@ -182,3 +182,72 @@ func (w *Wrapper) LeaderNodeAffinity(key string, values []string) *Wrapper {
...
+func (w *Wrapper) StartupPolicy(policy leaderworkersetv1.StartupPolicyType) *Wrapper {
+	w.Spec.StartupPolicy = policy
+	return w
+}
+
+func (w *Wrapper) WorkerName(name string) *Wrapper {
+	if len(w.Spec.LeaderWorkerTemplate.WorkerTemplate.Spec.Containers) == 0 {
+		w.Spec.LeaderWorkerTemplate.WorkerTemplate.Spec.Containers = []corev1.Container{{}}
+	}
+	w.Spec.LeaderWorkerTemplate.WorkerTemplate.Spec.Containers[0].Name = name
```
mwielgus, line 195: What if there is more than 1 container?

slice/internal/webhooks/statefulset_webhook_test.go
```diff
@@ -0,0 +1,180 @@
...
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	slice "tpu-slice-controller/api/v1beta1"
+	"tpu-slice-controller/internal/core"
+	utiltesting "tpu-slice-controller/internal/util/testing"
+)
+
+func TestStatefulSetDefault(t *testing.T) {
```
mwielgus, line 32: Add a test case where a StatefulSet already has unrelated NodeSelector entries (like kubernetes.io/os: linux) to ensure they are handled ok during injection.

mwielgus (commented):
+ Fix the typo in the commit.

slice/internal/webhooks/statefulset_webhook.go
pajakd: The health labels are always at the nodes (unlike the topology labels that are added only after the slice is Active), so it is not interfering with the scheduling to just add them in the LWS/Job/Jobset webhook.

mwielgus: I was asking whether we should add node selector for healthy/degraded nodes or not.

slice/internal/util/testingjobs/leaderworkerset/wrappers.go
pajakd: Hmm but this is only for e2e tests. Making it work for any number of containers would introduce additional complexity and in kueue also we have 1 container hardcoded in such wrappers (https://github.com/kubernetes-sigs/kueue/blob/90375d4a3e8c363f7187512c8f5ad4145d041610/pkg/util/testingjobs/leaderworkerset/wrappers.go#L189-L211)

slice/internal/webhooks/statefulset_webhook.go
pajakd: Right, fixed.

slice/internal/webhooks/statefulset_webhook_test.go
pajakd: Added.

slice/internal/webhooks/statefulset_webhook.go
pajakd: Good point, fixed.

slice/internal/webhooks/statefulset_webhook.go
pajakd: We add this healthy/degraded selector in LWS webhook and it is propagated to kueue workload, so TAS knows which nodes to exclude. I think it should be propagated from LWS to STS but even if its not then is it a problem?

PR #1179, opened 2026-04-27
[slice] New workload condition with reason for eviction

pajakd:
# Description

When evicting the workload (by setting the admission check to Retry) due to some slice issue, we would like to specify the reason for eviction in a new dedicated workload condition. I identified 4 possible issues:
1. Slice runtime failure (slice in state FAILED).
2. Slice formation timeout. If slice does not get ACTIVE within the specified time despite the retry mechanism being on.
3. Slice deletion (eg. with `kubectl delete ...`)
4. Slice configuration issue (eg. number of cubes not matching the topology)

# Issue

# Testing

### DIFF
--- a/slice/internal/controller/workload_controller.go
+++ b/slice/internal/controller/workload_controller.go
@@ -644,7 +644,7 @@ func (r *WorkloadReconciler) syncSlicesForAssignment(ctx context.Context, wl *ku
 			log.V(2).Info(fmt.Sprintf("Admission check %q updated state from %q to %q", ac.Name, ac.State, kueue.CheckStatePending), "reason", msg)
 			ac.State = kueue.CheckStatePending
 			ac.Message = api.TruncateConditionMessage(msg)
-			patchErr := r.updateWorkloadAdmissionCheckStatus(ctx, wl, ac, "")
+			patchErr := r.updateWorkloadStatus(ctx, wl, ac, "")
 			if patchErr != nil {
 				return nil, nil, errors.Join(err, patchErr)
 			}
@@ -724,10 +724,10 @@ func (r *WorkloadReconciler) evictWorkload(ctx context.Context, wl *kueue.Worklo
 	ac.State = kueue.CheckStateRetry
 	ac.RequeueAfterSeconds = ptr.To(int32(r.retryDelayOnSliceFailure.Round(time.Second).Seconds()))
 	ac.Message = api.TruncateConditionMessage(message)
-	return r.updateWorkloadAdmissionCheckStatus(ctx, wl, ac, reason)
+	return r.updateWorkloadStatus(ctx, wl, ac, reason)
 }
 
-func (r *WorkloadReconciler) updateWorkloadAdmissionCheckStatus(ctx context.Context, wl *kueue.Workload, ac *kueue.AdmissionCheckState, evictedReason string) error {
+func (r *WorkloadReconciler) updateWorkloadStatus(ctx context.Context, wl *kueue.Workload, ac *kueue.AdmissionCheckState, evictedReason string) error {
 	wlPatch := workload.BaseSSAWorkload(wl, true)
 	workload.SetAdmissionCheckState(&wlPatch.Status.AdmissionChecks, *ac, r.clock)
 
@@ -771,7 +771,7 @@ func (r *WorkloadReconciler) syncAdmissionCheckStatus(ctx context.Context, wl *k
 		return nil
 	}
 
-	if err := r.updateWorkloadAdmissionCheckStatus(ctx, wl, ac, evictedReason); err != nil {
+	if err := r.updateWorkloadStatus(ctx, wl, ac, evictedReason); err != nil {
 		return err
 	}
 
--- a/slice/internal/controller/workload_controller.go
+++ b/slice/internal/controller/workload_controller.go
@@ -741,7 +741,8 @@ func (r *WorkloadReconciler) updateWorkloadAdmissionCheckStatus(ctx context.Cont
 		}
 		apimeta.SetStatusCondition(&wlPatch.Status.Conditions, evictionCond)
 	} else if ac.State != kueue.CheckStateRetry {
-		if oldCond := apimeta.FindStatusCondition(wl.Status.Conditions, core.WorkloadSliceFailureConditionType); oldCond != nil {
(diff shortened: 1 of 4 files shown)


slice/internal/core/constants.go
```diff
@@ -75,6 +74,12 @@ const (
 	OwnerPodSetNameAnnotation        = "accelerator.gke.io/owner-podset-name"
 
 	LWSLeaderPodSetName = "leader"
+
+	WorkloadSliceFailureConditionType = "SliceFailure"
```
mwielgus: Maybe TPUSliceFailure, so it is clear what type of slice we are talking about.

slice/internal/controller/workload_controller.go
```diff
@@ -720,16 +720,37 @@ func (r *WorkloadReconciler) validatePartitionCount(
...
+func (r *WorkloadReconciler) evictWorkload(ctx context.Context, wl *kueue.Workload, ac *kueue.AdmissionCheckState, reason, message string) error {
 	ac.State = kueue.CheckStateRetry
 	ac.RequeueAfterSeconds = ptr.To(int32(r.retryDelayOnSliceFailure.Round(time.Second).Seconds()))
 	ac.Message = api.TruncateConditionMessage(message)
-	return r.updateWorkloadAdmissionCheckStatus(ctx, wl, ac)
+	return r.updateWorkloadAdmissionCheckStatus(ctx, wl, ac, reason)
 }
 
-func (r *WorkloadReconciler) updateWorkloadAdmissionCheckStatus(ctx context.Context, wl *kueue.Workload, ac *kueue.AdmissionCheckState) error {
+func (r *WorkloadReconciler) updateWorkloadAdmissionCheckStatus(ctx context.Context, wl *kueue.Workload, ac *kueue.AdmissionCheckState, evictedReason string) error {
```
mwielgus: The function updateWorkloadAdmissionCheckStatus now also manages the SliceFailure condition on the Workload. Consider renaming it to something like updateWorkloadStatus to better reflect its expanded responsibility.

slice/internal/controller/workload_controller.go
```diff
@@ -644,7 +644,7 @@ func (r *WorkloadReconciler) syncSlicesForAssignment(ctx context.Context, wl *ku
 			log.V(2).Info(fmt.Sprintf("Admission check %q updated state from %q to %q", ac.Name, ac.State, kueue.CheckStatePending), "reason", msg)
 			ac.State = kueue.CheckStatePending
 			ac.Message = api.TruncateConditionMessage(msg)
-			patchErr := r.updateWorkloadAdmissionCheckStatus(ctx, wl, ac)
+			patchErr := r.updateWorkloadAdmissionCheckStatus(ctx, wl, ac, "")
```
mwielgus: In the event of a slice creation failure, ac.State is set to Pending and updateWorkloadAdmissionCheckStatus is called with an empty evictedReason. This will clear any existing SliceFailure condition. Consider if we should instead report this creation failure via the SliceFailure condition as well

slice/internal/controller/workload_controller.go
pajakd: OK, done.

slice/internal/core/constants.go
pajakd: Good point.

slice/internal/controller/workload_controller.go
pajakd: This is slightly different because we are not evicting the workload in the case of this failure. We are returning with an error from the Reconcile and we will retry. The goal here is to have some way to differentiate the reasons for eviction so for now (in my opinion) we don't need to set the condition here.

Its a good point and in the future it would be reasonable to set some limits to these retries and eventually evict if we still cannot create after a few retries. But, for now we don't see issues with slice creation.

slice/internal/core/constants.go
```diff
@@ -75,6 +74,12 @@ const (
 	OwnerPodSetNameAnnotation        = "accelerator.gke.io/owner-podset-name"
 
 	LWSLeaderPodSetName = "leader"
+
+	WorkloadSliceFailureConditionType = "TPUSliceFailure"
+	WorkloadSliceDeletion             = "Deletion"
+	WorkloadSliceConfigurationFailure = "ConfigurationFailure"
+	WorkloadSliceRuntimeFailure       = "RuntimeFailure"
+	WorkloadSliceFormationTimeout     = "FormationTimeout"
```
mimowo: Let's add comments to all of these reasons.

slice/internal/controller/workload_controller.go
```diff
@@ -190,7 +190,7 @@ func (r *WorkloadReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 	desiredSlicesCount := totalDesiredSlices(wl, nodes)
 	if ac.State == kueue.CheckStateReady && (len(grouped.deleted) > 0 || len(slices) != desiredSlicesCount) {
 		log.V(3).Info("Slice has been deleted, evicting workload")
-		if err := r.evictWorkload(ctx, wl, ac, "Slice has been deleted"); err != nil {
+		if err := r.evictWorkload(ctx, wl, ac, core.WorkloadSliceDeletion, "Slice has been deleted"); err != nil {
```
mimowo, line 193: Is it immediately clear what is the name of the deleted slice?  I think this could make debugging easier for someone who doesn't know the slice naming convention. Also, it may clarify the case if there are multiple slices, and only one is failed. 

The same question / remark applies to other failure scenarios.

mimowo (commented):
LGTM, overall, but I think it would be great to include the name(s) of the failied slices. This should not be many of them, but if we are worried about bloating the message too much then we could truncate, similarly as we do in Kueue.

slice/internal/core/constants.go
pajakd: Added comments.

slice/internal/controller/workload_controller.go
pajakd: Good point. For the configuration failure the slice names were displayed.

I added the same for runtime failures and the timeout.

This deletion case is more tricky as the slice name is not available. But here I think it is not so much needed because this case only happens if someone deletes a slices while the workload is running (eg `kubectl delete`). I think in this case the name of that slice should be fairly easy to find.

pajakd: Plus, checked that the message is truncated

PR #1180, opened 2026-04-28
xpk: add team-based PoC quota routing with ConfigMap discovery

ultrons:
# Add team-based queue routing to `xpk workload create`

## Summary

This adds three flags to `xpk workload create` and one new sub-command
`xpk workload status`, enabling xpk to route a workload to a team-specific
Kueue ClusterQueue / LocalQueue / PriorityClass on clusters that have been set
up with multi-team quota sharing:

- `--team=<name>` — routes the JobSet to a per-team namespace + LocalQueue +
  PriorityClass, overriding `--priority`.
- `--value-class=<name>` — adds a job-class label for downstream audit and
  ordering.
- `--declared-duration-minutes=<int>` — added to the pod-template metadata so a
  cluster-side time-limit controller can stop overrunning jobs.
- `xpk workload status --team=<t> [--workload=<name>]` — diagnoses whether a
  workload is QUEUED / STUCK / RUNNING / FINISHED, and on STUCK cases parses
  recent Warning events to surface common admission failures (with a
  copy-pasteable shorter-name suggestion when the super-slice 49-char limit is
  hit).

When `--team` is **not** set, behavior is bit-identical to upstream — no
cluster calls, no new error paths, no new dependencies.

When `--team` **is** set, xpk reads a `kueue-system/poc-team-config` ConfigMap
on the target cluster to discover the available teams, namespaces, value
classes, and the super-slice name-length budget. This means **adding,
removing, or retuning a team is a cluster-side config change with no xpk
release needed**.

## Why a ConfigMap, not a hardcoded dict

An earlier iteration of this work (visible in the commit history of the fork)
shipped the team list as a Python dict in `core/kueue_manager.py`. That made
every cluster-side change (adding a team, retuning a quota, fixing a typo in a
PriorityClass name) require an xpk release + every user reinstalling. The
current shape moves the source of truth to the cluster: a small ConfigMap
rendered by the cluster admin's tooling. xpk's view of "what teams exist" is
always whatever `kubectl get cm -n kueue-system poc-team-config -o json` says
right now.

### Did-you-mean

```
$ xpk workload create --cluster=foo --team=mlpef ...
ERROR: --team='mlpef' not found on this cluster. Did you mean: ml-perf?
       Available teams: dev, gsc, ml-perf, nightly-regression, scale-test
```

### Tab completion (via existing `argcomplete` integration)

```
$ xpk workload create --cluster=bodab<TAB>
  bodaborg-super-alpha-cluster  bodaborg-tpu7x-8
  bodaborg-tpu7x-auto-nap2      bodaborg-tpu7x-nap-auto

$ xpk workload create --cluster=bodaborg-super-alpha-cluster --team=<TAB>
  dev  gsc  ml-perf  nightly-regression  scale-test
```

`--cluster` completes from `kubectl config get-contexts` (GKE prefix).
`--team` and `--value-class` complete from a small per-cluster cache at
`~/.xpk/poc-cache/<context>.json` that is refreshed every time a `--team`
invocation succeeds — so completion never blocks on a cluster call.

### Workload-name decoupling

Users can now pass any-length `--workload` name. xpk derives a short K8s-safe
JobSet name (`{ldap_prefix}-{hex4}` deterministic from
`sha256(display_name)`) that fits the super-slice admission controller's
49-char ceiling, and stashes the original on the JobSet via the
`xpk.google.com/workload` label. `xpk workload status / delete / list` then
use the label as the lookup key, so the user never needs to know the
truncated form.

## Commits

1. **xpk: add team-based routing flags for PoC quota system** — the three new
   flags, plumbed into the JobSet template. Initial implementation has the
   team table hardcoded.
2. **xpk: add `workload status` PoC queue diagnostic command** — the new
   sub-command and its parser.
3. **xpk: ConfigMap-driven team discovery + local cache for completion** —
   replaces the hardcoded table with runtime discovery, decouples the
   user-facing workload name from the K8s JobSet name, and adds the local
   cache that powers tab completion + did-you-mean. The end state of this
   commit is the actual feature the user-facing surface should be reviewed
   against; the earlier two are kept separate for reviewability of the layered
   design rather than because they ship a useful intermediate state.

## Backward compatibility

- No new required flags anywhere.
- `xpk workload create` without `--team` is unchanged: same default namespace,
  same `multislice-queue`, same priority handling, no extra cluster calls.
- `xpk workload create --help` no longer enumerates a fixed `choices=[...]`
  for `--team` / `--value-class` (because the set is cluster-dependent now);
  the help text points to runtime discovery. If maintainers prefer
  argparse-time validation, an alternative is to keep a non-strict default
  list in help text only.

## Note on naming

These flags and the `poc-team-config` ConfigMap are PoC-named throughout
because they were developed for a specific TPU quota PoC. The mechanism
itself is general-purpose ("team-aware queue routing for Kueue-enabled
clusters") and could be renamed if upstream prefers — happy to follow up
with a rename once we agree on a final name. Wanted to land the
implementation first to get the design discussion grounded.

## Test plan

- [x] `xpk workload create` (no `--team`): unchanged behavior, no cluster
  call for team discovery.
- [x] `xpk workload create --team=<valid>`: routes correctly to
  `<namespace>` / `lq` / `<priorityClass>`; pod-template carries the three
  labels.
- [x] `xpk workload create --team=<typo>`: errors with did-you-mean +
  available list, no JobSet created.
- [x] `xpk workload create --team=<valid>` against a cluster with no
  `poc-team-config` ConfigMap: errors with deployment hint, no JobSet
  created.
- [x] `xpk workload status --team=<t>`: lists all workloads in the team's
  namespace with QUEUED/STUCK/RUNNING/FINISHED diagnoses and quota
  percentages.
- [x] `xpk workload status --team=<t> --workload=<long-display-name>`:
  resolves to the K8s JobSet via the `xpk.google.com/workload` label.
- [x] `register-python-argcomplete xpk` then `<TAB>` on `--cluster`,
  `--team`, `--value-class`: returns expected completions.
- [x] Multi-cluster: cache at `~/.xpk/poc-cache/<context>.json` is
  per-cluster; completing on cluster A doesn't suggest teams from
  cluster B.

## Out of scope

- The cluster-side resources that produce the `poc-team-config` ConfigMap
  (Kueue ClusterQueues, namespaces, time-limit controller, the chart that
  renders them) live outside this PR. Reviewers don't need to deploy them
  to assess the xpk side; the discovery code is a no-op for any cluster
  that doesn't have the ConfigMap and `--team` unset.


src/xpk/commands/workload.py
```diff
@@ -104,6 +114,84 @@
 _PATHWAYS_WORKLOAD_TEMPLATE = 'pathways_workload_create.yaml.j2'
 
 _SUPER_SLICING_WORKLOAD_NAME_LIMIT = 28
+
+
+def _load_poc_cfg(args) -> dict | None:
```
jamOne-: What is PoC and why do we name our code after it?

src/xpk/commands/workload.py
```diff
@@ -104,6 +114,84 @@
...
+      xpk_print(
+          f'ERROR: --value-class={args.value_class!r} not valid on this cluster.'
+          f'{hint_line} Available: {", ".join(vcs)}'
+      )
+      xpk_exit(1)
+  args._poc_cfg = cfg
+  return cfg
+
+
+def _resolve_poc_team(args) -> tuple[str, str, str]:
```
jamOne-: Please create a dataclass instead for this ambiguous tuple.

src/xpk/commands/workload.py
```diff
@@ -472,6 +567,12 @@ def workload_create(args) -> None:
 
   parse_env_config(args, tensorboard_config)
 
+  # For PoC teams, inject the user's full display name as XPK_WORKLOAD_NAME so
+  # training scripts can reference it for GCS artifact paths regardless of the
+  # short K8s name xpk derives for the JobSet.
+  if getattr(args, 'team', None):
+    args.env.setdefault('XPK_WORKLOAD_NAME', args.workload)
```
jamOne-: This looks like a hack. Do we need this name shortening?

src/xpk/commands/workload.py
```diff
@@ -746,8 +848,21 @@ def workload_create(args) -> None:
         args, workload_system, parallel_containers
     )
 
+    poc_namespace, poc_local_queue, poc_priority = _resolve_poc_team(args)
+    # Derive a short K8s-safe JobSet name for PoC teams; use display name as-is otherwise.
```
jamOne-: But why? This can be unexpected from a user perspective. We have also other commands that take `workload` as an argument, e.g. `xpk inspector`.

So it actually required for the --teams changes?

src/xpk/commands/workload.py
```diff
@@ -1020,3 +1142,294 @@ def workload_list(args) -> None:
   xpk_print(f'See your workloads in Cloud Console: {workload_list_gcp_link}')
 
   xpk_exit(0)
+
+
+# ---------------------------------------------------------------------------
+# workload status — PoC queue health diagnostic
+# ---------------------------------------------------------------------------
```
jamOne-: nit: Please remove. I don't think it adds any value.

src/xpk/commands/workload.py
```diff
@@ -1020,3 +1142,294 @@ def workload_list(args) -> None:
...
+
+  get_cluster_credentials(args)
+
+  team = args.team
+  namespace, _lq, _pc = _resolve_poc_team(args)
+  cq_name = namespace  # ClusterQueue name matches namespace name
+
+  # ------------------------------------------------------------------
+  # Internal helpers (no --context needed; xpk sets up kubeconfig)
+  # ------------------------------------------------------------------
```
jamOne-: nit: Please remove?

src/xpk/commands/workload.py
```diff
@@ -1020,3 +1142,294 @@ def workload_list(args) -> None:
   xpk_print(f'See your workloads in Cloud Console: {workload_list_gcp_link}')
 
   xpk_exit(0)
+
+
+# ---------------------------------------------------------------------------
+# workload status — PoC queue health diagnostic
+# ---------------------------------------------------------------------------
+
+def workload_status(args) -> None:
```
jamOne-: What is the difference between `xpk workload status` and `xpk inspector workload=...`?

src/xpk/commands/workload.py
```diff
@@ -1020,3 +1142,294 @@ def workload_list(args) -> None:
...
+
+  team = args.team
+  namespace, _lq, _pc = _resolve_poc_team(args)
+  cq_name = namespace  # ClusterQueue name matches namespace name
+
+  # ------------------------------------------------------------------
+  # Internal helpers (no --context needed; xpk sets up kubeconfig)
+  # ------------------------------------------------------------------
+
+  def _kube_json(*kubectl_args):
```
jamOne-: All of these seem to be "status"-related, right? Please move it to a separate file.

src/xpk/core/kueue_manager.py
```diff
@@ -47,6 +48,33 @@
 WAIT_FOR_KUEUE_TIMEOUT = "10m"
 CLUSTER_QUEUE_NAME = "cluster-queue"
 LOCAL_QUEUE_NAME = "multislice-queue"
+
+def derive_k8s_workload_name(display_name: str, max_len: int) -> str:
```
jamOne-, line 65: Why did you put it between the CONSTS? Please move it below.

jamOne- (changes requested):
Thanks for the PR!

1. Please also resolve the failing presubmits.
2. Please add unit tests to your changes.

src/xpk/commands/workload.py
ultrons: Fair point — "PoC" was an internal project name and shouldn't leak into upstream code. Renamed throughout: poc_discovery.py → quota_discovery.py, _resolve_poc_team → _resolve_quota_team, _load_poc_cfg → _load_quota_cfg, _poc_cfg → _quota_cfg, _poc_labels → _team_labels, etc. All user-facing strings (parser help text, error messages, log lines) now use neutral wording like "team-quota" / "team-routing". The cluster-side ConfigMap name poc-team-config is kept as-is for now since renaming it is a chart-side change that needs separate coordination — happy to do that as a follow-up.

src/xpk/commands/workload.py
ultrons: Done. resolve_team() now returns a @dataclass(frozen=True) TeamRouting{namespace, local_queue, priority_class} (in core/quota_discovery.py). Callers use attribute access instead of positional unpacking

src/xpk/commands/workload.py
ultrons: The shortening exists because the super-slice admission controller has a 49-char hard limit on the slice name it creates from each JobSet (format: <namespace>-jobset-<name>-<kueue-hash>-slice-job-<replica-index>). Subtracting the fixed framing (~26 chars) and the namespace length leaves a small budget for the user-facing workload name — e.g. for poc-ml-perf (11 chars), only 12 chars are available. User-friendly names like mlperf-conv-13 (14 chars) blow past this, the JobSet/Slice fails admission, and the user sees a confusing low-level error. Without shortening, this is a real foot-gun.
  
 Updated this commit to:
 - Add an inline comment explaining the constraint and pointing at quota_discovery.max_k8s_workload_name_len for the budget math.
 - Add --no-shorten-jobset-name so users can opt out when their workload name is short enough already.
 - Only print the "workload → JobSet name" mapping when shortening actually triggered (no-op cases stay silent).

src/xpk/commands/workload.py
ultrons: Yes — same root cause as the previous comment. With --team, the workload routes to a per-team namespace, which both forces the JobSet to be created in that namespace AND eats into the slice-name char budget. Both behaviors only kick in when --team is set; the upstream non-team path is unchanged. Added a docstring + inline comment explaining the conditional. Plus the --no-shorten-jobset-name opt-out flag for users who'd rather see the create fail loudly than have their name silently shortened

src/xpk/commands/workload.py
ultrons: fixed by moving workload_status and its helpers into a dedicated commands/workload_status.py (also addresses the "separate file" comment below). The decorative banner comments are gone.

src/xpk/commands/workload.py
ultrons: See the previous comment.

src/xpk/commands/workload.py
ultrons: They're solving different problems for different audiences:
  
 - **xpk inspector** is a kitchen-sink debug dump for SREs investigating cluster bugs — it runs several kubectl/gcloud commands to infer and write raw cluster-wide state (configmaps, all node pools, kueue/jobset controller deployments, last 100 lines of controller logs, etc.) to a file. The --workload arg is largely a label hint; the dump is cluster-wide regardless.
 - **xpk workload status** answers "why is my job stuck?" with a 3-line plain-English diagnosis (RUNNING / QUEUED / STUCK / FINISHED, plus team-quota numbers and a specific kubectl describe ... fix when an AdmissionCheck error or queue position is detected).
 
Open to renaming this xpk workload diagnose if the difference would be clearer that way. Or, if you'd prefer one surface, I could fold the focused-diagnosis logic into xpk inspector --workload=X --diagnose-only. Which direction would you like?

src/xpk/commands/workload.py
ultrons: Done. workload_status() and its inner helpers now live in commands/workload_status.py. The parser imports workload_status from there. workload.py is back to non-status concerns

src/xpk/core/kueue_manager.py
ultrons: Moved. derive_k8s_workload_name() is now below the full module-level constants block (after MIN_CPU_LIMIT_SIZE). Also tightened its docstring to point at quota_discovery.max_k8s_workload_name_len instead of the old poc_discovery name

ultrons:
> Thanks for the PR!
> 
> 1. Please also resolve the failing presubmits.
> 2. Please add unit tests to your changes.

Added 23 new tests:
  - core/quota_discovery_test.py (new, 17 tests): resolve_team for known/unknown/empty-map cases, available_teams / available_value_classes happy + missing-key paths, max_k8s_workload_name_len with cfg / defaults / partial overrides, suggest did-you-mean ranking, and fetch_quota_config with kubectl output mocked (success / failure / malformed JSON / cache-write side effect).
  - core/kueue_manager_test.py (extended, 6 tests): derive_k8s_workload_name determinism, max_len cap, ldap-prefix shape, distinct inputs producing distinct hashes, short-input still-hashed.
 
  All 502 tests pass locally. Goldens regenerated for the renamed YAML template variables.

ultrons:
## Update — review feedback addressed (commit 4e284ca)

Thanks for the thorough review @jamOne-! Here's the round-trip on each comment.

### Renamings (response to "What is PoC and why...")

`PoC` was an internal project name and shouldn't have leaked into upstream code. Fully de-PoC-ified throughout:

| Before | After |
|---|---|
| `core/poc_discovery.py` | `core/quota_discovery.py` |
| `_resolve_poc_team` | `_resolve_quota_team` |
| `_load_poc_cfg` / `_poc_cfg` | `_load_quota_cfg` / `_quota_cfg` |
| `fetch_poc_config` | `fetch_quota_config` |
| `_build_poc_labels` (+ pod-template variant) | `_build_team_labels` (+ pod-template variant) |
| YAML template `{poc_labels}` / `{poc_pod_template_labels}` | `{team_labels}` / `{team_pod_template_labels}` |
| `~/.xpk/poc-cache/` | `~/.xpk/quota-cache/` |
| All "PoC" references in parser help text, docstrings, log messages | neutral team-quota wording |

Cluster-side ConfigMap name `poc-team-config` is kept as-is for now (renaming it is a chart-side change that needs separate coordination — happy to do that as a follow-up PR).

### Structure / code quality

- **Tuple → dataclass** (your `:159` comment): `resolve_team()` now returns a `@dataclass(frozen=True) TeamRouting{namespace, local_queue, priority_class}`. Callers use attribute access.
- **`workload_status` → its own file** (`:1204` comment): moved to `commands/workload_status.py`. The parser imports from there. `workload.py` is back to non-status concerns. The decorative banner nits at `:1149` and `:1202` are gone with it.
- **`derive_k8s_workload_name` placement** (kueue_manager `:52`): moved below the module-level constants block.

### Justifications added (`:574`, `:852`)

The K8s name shortening exists because the **super-slice admission controller has a 49-char hard limit** on slice names of the form `<namespace>-jobset-<name>-<kueue-hash>-slice-job-<replica-index>`. After the framing overhead and namespace, only ~12 chars are left for the user-facing workload name on a typical team — so `mlperf-conv-13` (14 chars) silently fails admission with an opaque K8s error. The shortening is a sharp-edge fix.

To make the behavior less surprising:
- Inline comment now points at `quota_discovery.max_k8s_workload_name_len` for the budget math.
- New `--no-shorten-jobset-name` opt-out flag.
- The "workload → JobSet name" log line only prints when shortening actually triggered (no-op cases stay silent).

### `workload status` vs `inspector` (`:1151`)

They serve different audiences:

- **`xpk inspector`** = SRE debug dump. Runs ~15 commands collecting cluster-wide state (configmaps, all node pools, kueue/jobset controller deployments + logs, etc.). The `--workload` arg is largely a label hint; the dump is cluster-wide regardless.
- **`xpk workload status`** = focused user-facing diagnosis. Answers "why is *my* job stuck?" in a 3-line plain-English answer (RUNNING / QUEUED / STUCK / FINISHED, plus team-quota numbers and a specific fix when an AdmissionCheck error or queue position is detected).

If you'd like I can rename to `xpk workload diagnose` to make the focused-diagnosis intent obvious, or fold the logic into `xpk inspector --workload=X --diagnose-only` to keep one surface. Let me know which direction you prefer.

### Tests (response to global "write tests for your changes")

**+23 new tests**, all 502 pass locally + goldens regenerated:

- `core/quota_discovery_test.py` (new, 17 tests): `resolve_team` for known/unknown/empty-map cases, `available_teams` / `available_value_classes` happy + missing-key paths, `max_k8s_workload_name_len` with cfg / defaults / partial overrides, `suggest` did-you-mean ranking, and `fetch_quota_config` with kubectl output mocked (success / failure / malformed JSON / cache-write side effect).
- `core/kueue_manager_test.py` (extended, 6 tests): `derive_k8s_workload_name` determinism, `max_len` cap, ldap-prefix shape, distinct inputs producing distinct hashes, short-input still-hashed.

### Files touched in `4e284ca`

| File | Change |
|---|---|
| `core/poc_discovery.py` → `core/quota_discovery.py` | renamed (with `git mv`) + dataclass + cleaned docstring |
| `core/quota_discovery_test.py` | new, 17 tests |
| `core/local_cache.py` | docstring + cache dir name |
| `core/kueue_manager.py` | moved `derive_k8s_workload_name` below CONSTS |
| `core/kueue_manager_test.py` | +6 tests for `derive_k8s_workload_name` |
| `commands/workload.py` | rename + dataclass call sites + opt-out flag handling |
| `commands/workload_status.py` | new (split from `workload.py`) |
| `parser/workload.py` | import from new module + neutral help text + new flag |
| `recipes/Workload_create*.md` | regenerated goldens for renamed YAML template variables |

src/xpk/commands/workload.py
jamOne-: Nice, thanks!

The poc-team-config name, let's change that as well please. We can make the name configurable in the config like: `xpk config set team-configmap-name poc-team-config` if you do not want to update your cluster.

You can read the name like here: https://github.com/AI-Hypercomputer/xpk/blob/20e3da5ad2881a912273b3aea64f3536f4470b7d/src/xpk/core/telemetry.py#L39.
And prepare the config key like here: https://github.com/AI-Hypercomputer/xpk/blob/20e3da5ad2881a912273b3aea64f3536f4470b7d/src/xpk/core/config.py#L59-L67.

src/xpk/parser/workload.py
```diff
@@ -426,7 +472,7 @@ def set_workload_delete_parser(workload_delete_parser: ArgumentParser):
       default=None,
       help='The name of the cluster to delete the job on.',
       required=True,
-  )
+  ).completer = _cluster_completer  # type: ignore[attr-defined]
```
jamOne-: Please do not by-pass the type-checker like this.

1. If it's needed add a coment above why.
2. But please prefer to resolve the issue so it does not introduce unexpected errors in the future.

src/xpk/commands/workload_status.py
jamOne-, line 1: Add argument types in those helper functions, otherwise they are hard to follow

src/xpk/core/kueue_manager_test.py
```diff
@@ -716,3 +716,44 @@ def _first(generator: Generator[T, None, None]) -> T:
   result = next(generator, None)
   assert result is not None
   return result
+
+
+# ---------------------------------------------------------------------------
+# derive_k8s_workload_name
+# ---------------------------------------------------------------------------
```
jamOne-: Please remove we do not use this kind of comments

src/xpk/parser/workload.py
jamOne-, line 1: I think we should also update our docs about these features.

src/xpk/commands/workload_status.py
```diff
@@ -0,0 +1,376 @@
...
+    args: user provided arguments (--cluster, --team, --workload).
+  """
+  if should_validate_dependencies(args):
+    validate_dependencies_list(
+        args, [SystemDependency.KUBECTL, SystemDependency.GCLOUD]
+    )
+
+  # Fill project from gcloud config if not provided.
+  if not getattr(args, 'project', None):
+    r = _subprocess.run(
```
jamOne-: Why not using `run_command_for_value`?

src/xpk/commands/workload_status.py
```diff
@@ -0,0 +1,376 @@
...
+          f' {args.project}.'
+      )
+      xpk_exit(1)
+    args.zone = loc.strip()
+
+  get_cluster_credentials(args)
+
+  # Imported lazily to avoid a circular import: commands/workload.py imports
+  # this module's parser, and we'd otherwise close the loop.
+  from .workload import _resolve_quota_team  # pylint: disable=import-outside-toplevel
```
jamOne-: Move the quota function to another file then. Do not disable linter checks...

src/xpk/commands/workload_status.py
```diff
@@ -0,0 +1,376 @@
...
+      return '?'
+    ts = datetime.fromisoformat(ts_str.replace('Z', '+00:00'))
+    s = int((datetime.now(timezone.utc) - ts).total_seconds())
+    if s < 60:
+      return f'{s}s'
+    if s < 3600:
+      return f'{s // 60}m'
+    return f'{s // 3600}h{(s % 3600) // 60}m'
+
+  def _cond(wl, ctype):
```
jamOne-: Maybe we could use https://github.com/AI-Hypercomputer/xpk/blob/20e3da5ad2881a912273b3aea64f3536f4470b7d/src/xpk/core/kubectl_common.py#L38?

src/xpk/commands/workload.py
```diff
@@ -746,8 +870,35 @@ def workload_create(args) -> None:
...
 
+    routing = _resolve_quota_team(args)
+    team_namespace = routing.namespace
+    # Derive a short K8s-safe JobSet name when the workload is routed to a
+    # team namespace AND the user-facing workload name exceeds the
+    # super-slice admission controller's per-name budget. See
+    # core.kueue_manager.derive_k8s_workload_name and
+    # core.quota_discovery.max_k8s_workload_name_len for the budget math.
+    # The user can disable shortening with --no-shorten-jobset-name.
+    if team_namespace:
```
jamOne-: Why only when `team_namespace` is used? Seems inconsistent to me.

src/xpk/commands/workload.py
jamOne-: Got it, thanks. I think `xpk workload status` SG in the end. I'm not sure though if it should be a part of this PR as it seems unrelated to the Team changes, making it all harder to review.

src/xpk/commands/workload.py
jamOne-: @FIoannides I need your review on that one. IMO this decision can be breaking for some users.

jamOne- (changes requested):

src/xpk/commands/workload.py
ultrons: Intentional. This PR is meant to be purely additive: workloads submitted without --team keep their original K8s name and don't go through any of the new code paths, so existing xpk users see zero behavior change. The shortening solves a problem specific to the team-routing path (because namespace prefix eats most of the super-slice charLimit budget), so it's gated on that path. Happy to extend it to non-team workloads in a follow-up PR if you'd prefer that — but doing so here would change naming behavior for every existing xpk user, which felt riskier than warranted for this PR.

src/xpk/commands/workload_status.py
ultrons: Done in 7d77cc5 — every inner helper in workload_status.py is now fully annotated. 

Note: this file is moving to its own PR (see your :1151 comment) so the annotated version lives on the workload-status-pr branch.

src/xpk/commands/workload_status.py
ultrons: Makes sense!  fix applied. This whole file is removed from PR #1180 per the :1151 split, so it's no longer in the diff here. The change is on the workload-status-pr branch (lines 57-62) and will land with the follow-up PR:

  >> was:  subprocess.run(['gcloud', 'config', 'get', 'project'], ...)
  rc, out = run_command_for_value(
      'gcloud config get project',
      task='gcloud config get project',
      quiet=True,
  )
  args.project = out.strip().splitlines()[-1] if rc == 0 and out else ''
--

src/xpk/commands/workload_status.py
ultrons: Done in 7d77cc5. The two quota helpers _load_quota_cfg and _resolve_quota_team (now load_quota_cfg and resolve_team_for_args) moved from commands/workload.py to core/quota_discovery.py — their natural home. Both call sites import from there at module top: no import-outside-toplevel, no pylint disable. Side benefit: args._quota_cfg → args.quota_cfg drops the
  matching protected-access disable too.

src/xpk/commands/workload_status.py
ultrons: Done in 7d77cc5. workload_status._cond / _is_true now use kubectl_common.parse_kubernetes_status for typed condition access. Added a reason field to KubernetesCondition (with one new test in kubectl_common_test.py) so the typed access covers all four fields workload_status reads. (Also moving to the separate PR.)

src/xpk/core/kueue_manager_test.py
ultrons: Removed in 7d77cc5.

src/xpk/parser/workload.py
ultrons: Added in e4e19a3 — new section in docs/usage/workloads.md covering --team, --value-class, --declared-duration-minutes, --no-shorten-jobset-name, plus the team-configmap-name config override and the auto-shortening rationale.

src/xpk/parser/workload.py
ultrons: Fair callout,  point 2 isn't fully resolved by my change, Here is where things stand:

  What's done in 7d77cc5 + 383b3a3:
  - Every # type: ignore[attr-defined] removed at the call sites.
  - New _set_completer(action, completer) helper centralises the dynamic-attribute assignment.
  - Helper docstring (expanded in 383b3a3) explicitly acknowledges that setattr is itself a bypass that mypy can't introspect, and explains why no purely-static fix is currently available: argcomplete monkey-patches argparse.Action.completer at runtime, and there is no upstream type stub that declares it.

  Two options for actually resolving the bypass:
  1. Wait for argcomplete to ship updated stubs declaring Action.completer. Out of our hands.
  2. Shadow Action with a Protocol that declares completer: Any and cast at the helper boundary. This is a few extra lines but makes the "trust me, this attribute exists" contract explicit to mypy. I haven't done this because it'd diverge from the existing pattern at parser/common.py:71, which uses a different escape hatch (untyped add_argument returning Any), wanted to check before introducing a new pattern just for these new flags.

  Happy to switch to (2) if you'd prefer it here, or to tackle it as a cross-cutting cleanup (consistent across both new code and parser/common.py) in a follow-up. Let me know which

src/xpk/commands/workload.py
ultrons: Done in 3a9ac68. Default is now team-quota-config (no poc- literal anywhere). Operators with the legacy ConfigMap can override without touching the cluster: xpk config set team-configmap-name poc-team-config. Wired exactly per your references — added TEAM_CONFIGMAP_NAME_KEY to core/config.py's DEFAULT_KEYS, and quota_discovery._configmap_name() reads it via
  get_config().get(...) with the default as fallback. +2 tests covering both the default and the override path.

src/xpk/commands/workload.py
ultrons: Agreed — done in dea20f7. workload status and its parser hookup are removed from this PR. The implementation (with the round-2 cleanup applied: typed helpers, run_command_for_value, kubectl_common, no lazy imports) is preserved at https://github.com/ultrons/xpk/tree/workload-status-pr and will be opened as a separate PR after #1180 merges (it depends on the new resolve_team_for_args helper, so it's stacked)

ultrons:
Thanks for the round-2 review @jamOne-! Pushed dea20f7, 3a9ac68, e4e19a3, 383b3a3 (on top of 7d77cc5 from the previous round). Per-comment responses inline. All four CI checks green.

docs/usage/workloads.md
```diff
@@ -154,6 +154,60 @@ xpk workload create \
...
+  --tpu-type=tpu7x-128 \
+  --num-slices=2 \
+  --workload=$WORKLOAD_NAME \
+  --command="python3 train.py" \
+  --team=ml-perf \
+  --value-class=benchmark \
+  --declared-duration-minutes=90
+```
+
+The new flags (each optional individually, but typically used together):
```
jamOne-: Please remove this line

recipes/Workload_create.md
jamOne-, line 1: Could you create a recipe file for your use-case, please? 

It'll require providing a dry-run ConfigMap somehow. I'd suggest doing that via env var, just like we do here: [DRY_RUN_RESERVATION_SUB_BLOCKS](https://github.com/AI-Hypercomputer/xpk/blob/20e3da5ad2881a912273b3aea64f3536f4470b7d/recipes/Cluster_create_super-slicing.md?plain=1#L6), we then read it like https://github.com/AI-Hypercomputer/xpk/blob/20e3da5ad2881a912273b3aea64f3536f4470b7d/src/xpk/core/reservation.py#L481. Recipes give us: insights into how the final YAML looks like and a better understanding of the blast radius of a change.

src/xpk/commands/workload.py
```diff
@@ -395,6 +441,10 @@ def workload_create(args) -> None:
     k8s_api_client = setup_k8s_env(args)
     setup_k8s_service_accounts()
 
+  # Default team-routing values; the TPU non-pathways path below may override.
```
jamOne-: Please remove this comment. It does not bring any value.

src/xpk/commands/workload.py
```diff
@@ -472,6 +522,13 @@ def workload_create(args) -> None:
 
   parse_env_config(args, tensorboard_config)
 
+  # When --team is set, inject the user's full display name as XPK_WORKLOAD_NAME
+  # so training scripts can reference it for GCS artifact paths regardless of
+  # the (potentially shortened) K8s name xpk derives for the JobSet.
+  team = getattr(args, 'team', None)
+  if isinstance(team, str) and team.strip():
+    args.env.setdefault('XPK_WORKLOAD_NAME', args.workload)
+
```
jamOne-: Can we move this under the `if team_routing`?

src/xpk/core/local_cache.py
```diff
@@ -0,0 +1,173 @@
...
+
+
+def _path_for(context: str) -> Path:
+  return CACHE_DIR / f"{_safe_key(context)}.json"
+
+
+def current_context() -> str | None:
+  """Return the current kubectl context, or None."""
+  try:
+    r = subprocess.run(
```
jamOne-: Please use `run_command_for_value`

src/xpk/parser/workload.py
```diff
@@ -27,6 +30,65 @@
...
+    if not candidates:
+      for c in local_cache.all_contexts():
+        payload = local_cache.read(c) or {}
+        candidates.update(payload.get(field) or [])
+    return sorted(v for v in candidates if v.startswith(prefix or ''))
+
+  return _complete
+
+
+def _set_completer(action: argparse.Action, completer: Any) -> None:
```
jamOne-: This still looks like a hack, also why is the `completer` of `Any` type?

We already have some `completer` logic in https://github.com/AI-Hypercomputer/xpk/blob/20e3da5ad2881a912273b3aea64f3536f4470b7d/src/xpk/parser/common.py#L71. And we do not need any hacks or disabling mypy. Why is your code different?

src/xpk/parser/workload.py
```diff
@@ -27,6 +30,65 @@
...
+  call site.
+
+  A purely-static fix would require either:
+    (a) updated argcomplete type stubs upstream that declare
+        `Action.completer`, or
+    (b) shadowing `Action` with a `Protocol` that declares
+        `completer: Any` and casting at this boundary.
+
+  Happy to switch to (b) on request — leaving as setattr keeps this in
+  line with the existing xpk pattern at parser/common.py:71.
```
jamOne-: Looks like an AI output that should not be commited?

src/xpk/parser/workload.py
```diff
@@ -27,6 +30,65 @@
...
+  """
+  setattr(action, 'completer', completer)
+
+
+def _cluster_completer(prefix, parsed_args, **_kwargs):
```
jamOne-: Please add argument types

jamOne- (changes requested):
Gemini review below:

The `reviewer` agent has completed the review of the pull request. Here is the verdict:

### Code Review Verdict

**Status:** 🔴 Changes Requested

**Overall Impression:** The introduction of team-based routing and a runtime configurable quota discovery mechanism is a fantastic architectural improvement, successfully decoupling xpk releases from cluster administration. The code is well-structured, well-tested, and provides robust caching for responsive autocompletion. However, there are a few edge cases related to input validation, logic constraints, and shell injection hygiene that must be addressed before merging.

### Finding Summary

| ID | Severity | Category | File/Location | Finding (Concise) |
| :--- | :--- | :--- | :--- | :--- |
| **F1** | 🔴 Critical | Security | `src/xpk/core/quota_discovery.py` (fetch_quota_config) | The customizable `team-configmap-name` is directly interpolated into a shell command, creating a shell injection vulnerability. |
| **F2** | 🟡 Important | Logic | `src/xpk/core/kueue_manager.py` (derive_k8s_workload_name) | `derive_k8s_workload_name` incorrectly uses negative slicing if `max_len < 5`, violating the max length constraint. |
| **F3** | 🟡 Important | Validation | `src/xpk/core/quota_discovery.py` (load_quota_cfg) | The `--declared-duration-minutes` flag is not validated as required when `--team` is set, despite documentation stating otherwise. |

### Detailed Findings

#### **F1** - Prevent shell injection in `kubectl` command execution

- **Location:** `src/xpk/core/quota_discovery.py` around line 68 (`fetch_quota_config`)
- **Why it matters:** The `cm_name` variable is retrieved from the user's config via `get_config().get(TEAM_CONFIGMAP_NAME_KEY)` and directly interpolated into the `cmd` string, which is executed via `run_command_for_value`. While the user controls their own config, this still permits arbitrary local code execution if a malformed or malicious ConfigMap name (e.g. `test; rm -rf /`) is set in the XPK config.
- **How to fix:** Use `shlex.quote` to properly escape the ConfigMap name before string interpolation.

```python
import shlex

def fetch_quota_config() -> dict | None:
  # ...
  cm_name = shlex.quote(_configmap_name())
  cmd = f'kubectl get configmap -n {CONFIG_CM_NAMESPACE} {cm_name} -o json'
```

#### **F2** - Fix negative slicing bug in `derive_k8s_workload_name`

- **Location:** `src/xpk/core/kueue_manager.py` around line 84
- **Why it matters:** The function computes `max_prefix = max_len - 5` and then slices the string using `ldap[:max_prefix]`. If `max_len` is less than 5 (which is possible if a cluster admin misconfigures the `sliceName` character limits or uses an exceptionally long namespace), `max_prefix` becomes negative. In Python, a negative slice index counts from the end of the string, which causes the function to return a string much longer than `max_len`, thus silently violating the super-slice Kubernetes character limits and causing downstream admission check failures.
- **How to fix:** Ensure `max_len` is at least 5 before attempting to generate the prefix and hash, raising an explicit error if it's too small.

```python
def derive_k8s_workload_name(display_name: str, max_len: int) -> str:
  # ...
  if max_len < 5:
    raise ValueError(f"Calculated max_len ({max_len}) is too small to accommodate the required short name format. Please check the 'sliceName' character limits in the team-quota ConfigMap.")
  # Reserve 5 chars for '-' + 4 hex digits
  max_prefix = max_len - 5
  ldap = display_name.split("-")[0]
  prefix = ldap[:max_prefix]
  hex4 = hashlib.sha256(display_name.encode()).hexdigest()[:4]
  return f"{prefix}-{hex4}"
```

#### **F3** - Enforce `--declared-duration-minutes` when `--team` is specified

- **Location:** `src/xpk/core/quota_discovery.py` inside `load_quota_cfg`
- **Why it matters:** The `argparse` help text explicitly states that `--declared-duration-minutes` is "Required when --team is set." However, `load_quota_cfg` does not enforce this. If a user omits the flag, the workload might be submitted without the duration label, preventing the time-limit controller from stopping overrunning jobs or functioning correctly.
- **How to fix:** Add explicit validation to ensure the duration value is provided whenever the team flag is utilized.

```python
def load_quota_cfg(args) -> dict | None:
  # ...
  if args.team not in (cfg.get('teams') or {}):
    # ... existing team not found error handling ...
    xpk_exit(1)
    
  if getattr(args, 'declared_duration_minutes', None) is None:
    xpk_print(
        'ERROR: --declared-duration-minutes is required when --team is set.'
    )
    xpk_exit(1)
```

docs/usage/workloads.md
ultrons: Done.

recipes/Workload_create.md
ultrons: Done. Added XPK_TEAM_QUOTA_DRY_RUN_CONFIG env var (modeled on DRY_RUN_RESERVATION_SUB_BLOCKS) and recipes/Workload_create_with_team_routing.md showing the rendered  JobSet with team-routing fields highlighted. Also passing the golden tests via tools/recipes.py

src/xpk/commands/workload.py
ultrons: Done.

src/xpk/commands/workload.py
ultrons: Done.

src/xpk/parser/workload.py
ultrons: Indeed. I failed to spot the .completer chain example you pointed out. Removed _set_completer. Inlined .completer = X at all 5 call sites as in your example.
 I did relax the parameter type on the two enclosing parser-setup functions (set_workload_delete_parser, set_workload_list_parser) to match the existing convention in add_shared_workload_create_required_arguments in the same file - untyped parser param so Any.add_argument().completer = X works for mypy.

src/xpk/parser/workload.py
ultrons: It is indeed. I instructed the agent to write that comment since I was not satisfied with the fix. Gone now along with _set_completer per the previous comment.

src/xpk/parser/workload.py
ultrons: Done. _cluster_completer and the inner _complete in _cached_field_completer are fully typed.

ultrons:
Pushed three commits addressing round-3 on top of 383b3a3:

  - 9bd7cfe — nits: #1, #3, #4, #5, #8
  - 393a33e — #6/#7: dropped _set_completer, inlined .completer = X matching parser/common.py:71
  - 05f4d46 — #2: XPK_TEAM_QUOTA_DRY_RUN_CONFIG env var + recipes/Workload_create_with_team_routing.md

  Local CI: mypy, pyink, pylint, pytest (504), and tools/recipes.py golden all green. Ready for re-review.

ultrons:
Pushed 65c4ec4 addressing the three Gemini findings:

  - F1 (shell injection): shlex.quote on cm_name before kubectl interpolation in fetch_quota_config.
  - F2 (negative slicing): derive_k8s_workload_name now raises ValueError when max_len < 5, surfacing the ConfigMap misconfiguration up-front instead of silently emitting a too-long name.
  - F3 (required-flag enforcement): load_quota_cfg now validates --declared-duration-minutes when --team is set, matching the help text.

  +4 unit tests covering each case. mypy, pyink, pylint, pytest (508), and tools/recipes.py golden all ok locally.

src/xpk/parser/workload.py
```diff
@@ -405,7 +450,7 @@ def set_workload_create_pathways_parser(
   workload_create_pathways_parser.set_defaults(func=workload_create_pathways)
 
 
-def set_workload_delete_parser(workload_delete_parser: ArgumentParser):
```
jamOne-, line 408: Oh I see it now why it worked for `ParserOrArgumentGroup`, it just types the object to `Any` after `add_argument`...

Sorry for this confusion, could you please revert removing the types and go back to the original solution (add mypy bypassing comments)? That was the cleanest approach of all that we have available.

src/xpk/parser/workload.py
```diff
@@ -660,6 +705,52 @@ def add_shared_workload_create_optional_arguments(args_parsers):
...
+        '--team',
+        type=str,
+        default=None,
+        help=(
+            'Team name. When set, automatically routes the job to the'
+            " team's namespace, LocalQueue, and PriorityClass. Overrides"
+            ' --priority. Valid teams are discovered at runtime from'
+            " the cluster's team-quota ConfigMap."
+        ),
+    ).completer = _cached_field_completer('teams')
```
scaliby, line 722: Please move field caching here and below to a separate PR and make it consistent with other arguments. Otherwise drop this.
```
.completer = _cached_field_completer('teams')
```

src/xpk/parser/workload.py
jamOne-: Ah got it.

src/xpk/parser/workload.py
```diff
@@ -27,6 +30,48 @@
 from .validators import directory_path_type, name_type
 
 
+def _cached_field_completer(field: str) -> Callable[..., list[str]]:
```
scaliby, line 33: please move this to a separate PR. This is not related to team-based routing

src/xpk/core/quota_discovery_test.py
```diff
@@ -0,0 +1,298 @@
...
+
+
+def test_load_quota_cfg_requires_declared_duration_minutes_with_team(mocker):
+  """The --declared-duration-minutes help text says it's required when
+  --team is set. load_quota_cfg enforces that — submission without it
+  must fail loud, not silently emit a workload missing the duration label."""
+  cfg = _sample_cfg()
+  mocker.patch('xpk.core.quota_discovery.fetch_quota_config', return_value=cfg)
+  mocker.patch('xpk.core.quota_discovery.xpk_print')
+  mocker.patch('xpk.core.quota_discovery.xpk_exit', side_effect=SystemExit(1))
```
scaliby, line 276: `xpk_exit` by default raises `SystemExit`. Is this mock necessary?

src/xpk/core/quota_discovery_test.py
```diff
@@ -0,0 +1,298 @@
...
+# ---------- load_quota_cfg validation ----------
+
+
+def test_load_quota_cfg_requires_declared_duration_minutes_with_team(mocker):
+  """The --declared-duration-minutes help text says it's required when
+  --team is set. load_quota_cfg enforces that — submission without it
+  must fail loud, not silently emit a workload missing the duration label."""
+  cfg = _sample_cfg()
+  mocker.patch('xpk.core.quota_discovery.fetch_quota_config', return_value=cfg)
+  mocker.patch('xpk.core.quota_discovery.xpk_print')
```
scaliby, line 275: Why? I think it's fine to just print some stuff as a part of the test.

jamOne- (changes requested):
Gemini review below. Please resolve `F1` and `F2`.

Here is the review of the pull request https://github.com/AI-Hypercomputer/xpk/pull/1180:

### Code Review Verdict

**Status:** 🔴 Changes Requested

**Overall Impression:** The PR introduces a robust and flexible team-based Kueue routing mechanism via ConfigMaps and implements an elegant local caching strategy for sub-second CLI tab completions. The codebase is well-structured and follows clean abstraction principles. However, there are critical logic errors in the JobSet name derivation algorithm that will cause the Kubernetes API to reject workloads under specific constraints, as well as a shell injection vulnerability when executing commands with user-provided config values.

### Finding Summary

| ID | Severity | Category | File/Location | Finding (Concise) |
| :--- | :--- | :--- | :--- | :--- |
| **F1** | 🔴 Critical | Logic | `src/xpk/core/kueue_manager.py` | Negative string slicing in `derive_k8s_workload_name` produces invalid or excessively long names when `max_len` < 6. |
| **F2** | 🟡 Important | Security | `src/xpk/core/quota_discovery.py` | Missing shell escaping (`shlex.quote`) for the ConfigMap name risks arbitrary command execution. |
| **F3** | 🔵 Suggestion | Error Handling | `src/xpk/core/local_cache.py` | Overly broad exception suppression (`except Exception`) masks legitimate filesystem or permission errors. |

### Detailed Findings

#### **F1** - Logic Error in Workload Name Derivation

*   **Location:** `src/xpk/core/kueue_manager.py:86` (in `derive_k8s_workload_name`)
*   **Why it matters:** The slice length math relies on `max_prefix = max_len - 5`. If a team has a long namespace (e.g., > 18 characters), `max_len` can drop below 5, making `max_prefix` negative.
    1.  **Negative Slicing Bug:** In Python, `ldap[:negative_number]` removes characters from the *end* of the string rather than capping the front (e.g., `"amandaliang"[:-1]` becomes `"amandalian"`). This creates a name that is *longer* than `max_len`, violating the strict 49-character super-slice controller limit and causing loud admission failures.
    2.  **RFC 1123 Violation:** If `max_len == 5`, `max_prefix == 0`. `ldap[:0]` evaluates to `""`, resulting in a derived name like `"-abcd"`. Kubernetes API enforces RFC 1123, which strictly requires object names to start with an alphanumeric character, not a hyphen.
*   **How to fix:** Ensure `max_len` is sufficiently large (at least 6) and safely handle edge cases before attempting to slice the string.

```python
def derive_k8s_workload_name(display_name: str, max_len: int) -> str:
  if max_len < 6:
    raise ValueError(
        f"max_len ({max_len}) is too small to accommodate a valid "
        "Kubernetes name with the deterministic hash suffix. Check the "
        "namespace length and sliceName limits."
    )
  # Reserve 5 chars for '-' + 4 hex digits
  max_prefix = max_len - 5
  ldap = display_name.split("-")[0]
  prefix = ldap[:max_prefix]
  hex4 = hashlib.sha256(display_name.encode()).hexdigest()[:4]
  return f"{prefix}-{hex4}"
```

#### **F2** - Command Injection Vulnerability via Missing Escaping

*   **Location:** `src/xpk/core/quota_discovery.py:75` (in `fetch_quota_config`)
*   **Why it matters:** The ConfigMap name is resolved from local configuration (`get_config().get(TEAM_CONFIGMAP_NAME_KEY)`) and immediately interpolated into a `kubectl` command executed with `shell=True` (via `run_command_for_value`). If a user configures their `team-configmap-name` with shell metacharacters (e.g., `xpk config set team-configmap-name "foo; rm -rf /"`), it will execute arbitrary commands on their machine. While `xpk` relies heavily on `shell=True`, configuration-driven injection is an avoidable risk.
*   **How to fix:** Use `shlex.quote()` to safely escape the dynamically retrieved ConfigMap name.

```python
import shlex

def fetch_quota_config() -> dict | None:
  # ...
  cm_name = shlex.quote(_configmap_name())
  cmd = f'kubectl get configmap -n {CONFIG_CM_NAMESPACE} {cm_name} -o json'
  rc, out = run_command_for_value(cmd, task=cmd, quiet=True)
  # ...
```

#### **F3** - Overly Broad Exception Suppression in Local Cache

*   **Location:** `src/xpk/core/local_cache.py` (in `write`, `read`, `all_contexts`)
*   **Why it matters:** The cache operations wrap their execution in a broad `except Exception: pass` (or returning `None`). While making the cache "best effort" is a good architectural choice, silently catching all `Exception`s suppresses logical bugs (like `NameError`, `TypeError`) alongside legitimate `OSError`s (like full disks or permission issues).
*   **How to fix:** Narrow the exception scope to catch only expected runtime failure conditions, making debugging significantly easier if the code breaks in the future.

```python
def read(context: str) -> dict | None:
  # ...
  try:
    p = _path_for(context)
    if not p.exists():
      return None
    payload = json.loads(p.read_text())
    return payload if isinstance(payload, dict) else None
  except (OSError, json.JSONDecodeError):
    return None
```

scaliby (changes requested):
This change is too big to be reviewed and carries a significant risk. Please work on splitting it into small, incremental PR's that are easy to reason about.

PR #1182, opened 2026-05-01
feat: move colocated-python-sidecar to worker job in pathways template

ChromeHearts:
Reported in b/b/507543259.  

This change moves the colocated-python-sidecar from the pathways-head job to the worker job as an initContainer. It also updates the sidecar configuration with necessary environment variables for gRPC, unbuffered logging, and high-verbosity debug logs.

# Description
Change the Pathways workload creation template.

# Issue
when --colocated-python-sidecar-image is set, the created jobset is incorrect.

# Testing
In this PR, I am able to run colocated-python job successfully.


scaliby:
@ChromeHearts please update goldens.

scaliby (approved):

PR #1184, opened 2026-05-14
FIx scheduling of nap + pathways workloads

scaliby:
# Description
`autoprovisioning_args` can contain following values depending on selected capacity:
- On-Demand: An empty string, meaning no special node selectors
- Spot: `cloud.google.com/gke-spot: "true"`
- Flex-Start: `cloud.google.com/gke-queued: "true"`
- Reservation: `cloud.google.com/reservation-name: <reservation_name>`

This causes scheduling conflict with already declared `cloud.google.com/gke-nodepool: cpu-np` `nodeSelector` and causes workload to never be scheduled.

# Issue
b/510416959

# Testing
Unit tests.

### DIFF
--- a/src/xpk/commands/workload_test.py
+++ b/src/xpk/commands/workload_test.py
@@ -487,8 +487,8 @@ def test_pathways_yaml_excludes_autoprovisioning_args_from_head(mocker):
       jobset_annotations='',
   )
 
-  head_section = yaml_output[:yaml_output.find('name: worker')]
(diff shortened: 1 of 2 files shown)


SikaGrr (approved):
LGTM, but why was this included in the first place? I assume it was needed, so could removing it possibly break something else?

jamOne- (approved):
fix tests pls

PR #1188, opened 2026-06-19
Add a adapt_from_ct cluster create flag

SikaGrr:
# Description
Add a --adapt_from_ct flag to cluster create, skipping cluster and nodepool creation

# Issue
b/525457681

# Testing
Recipe, unit tests.

### DIFF
--- a/recipes/Cluster_create_adapt_from_ct.md
+++ b/recipes/Cluster_create_adapt_from_ct.md
@@ -1,0 +1,231 @@
+# Cluster create with adapt_from_ct
+Creates a basic GKE cluster, skipping hardware setup via --adapt-from-ct, and setting up only ConfigMaps and Kueue.
+
+# Running the command
+```shell #golden
+xpk cluster create --project=golden-project --zone=us-central1-a --cluster=golden-cluster --tpu-type=tpu7x-8 --spot --adapt-from-ct
+```
+<!--
+$ xpk cluster create --project=golden-project --zone=us-central1-a --cluster=golden-cluster --tpu-type=tpu7x-8 --spot --adapt-from-ct
+[XPK] Starting xpk v0.0.0
+[XPK] Starting cluster create for cluster golden-cluster:
+[XPK] Working on golden-project and us-central1-a
+[XPK] Cluster creation and Nodepool creation was skipped due to the --adapt-from-ct flag.
+[XPK] Task: `Find cluster region or zone` is implemented by the following command not running since it is a dry run. 
+gcloud container clusters list --project=golden-project --filter=name=golden-cluster --format="value(location)"
+[XPK] Task: `get-credentials-dns-endpoint to cluster golden-cluster` is implemented by the following command not running since it is a dry run. 
+gcloud container clusters get-credentials golden-cluster --location=us-central1 --dns-endpoint --project=golden-project && kubectl config view && kubectl config set-context --current --namespace=default
+[XPK] Task: `Test kubectl credentials` is implemented by the following command not running since it is a dry run. 
+kubectl get pods
+[XPK] Finished get-credentials and kubectl setup.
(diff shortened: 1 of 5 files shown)


src/xpk/commands/cluster.py
```diff
@@ -387,66 +387,72 @@ def cluster_create(args) -> None:
   _validate_cluster_create_args(args, system)
```
jamOne-: Could we skip that one too? For instance, this validation would be blocking for super-slicing clusters as it would require passing a gsc reservation

src/xpk/commands/cluster.py
```diff
@@ -543,13 +551,14 @@ def cluster_create(args) -> None:
       f' https://console.cloud.google.com/kubernetes/clusters/details/{get_cluster_location(args.project, args.cluster, args.zone)}/{args.cluster}/details?project={args.project}'
   )
 
-  if args.managed_mldiagnostics:
+  if not adapt_from_ct and args.managed_mldiagnostics:
```
jamOne-: I think this one does not need to be skipped, WDYT?

recipes/Cluster_create_adapt_from_ct.md
```diff
@@ -0,0 +1,230 @@
...
+<!--
+$ xpk cluster create --project=golden-project --zone=us-central1-a --cluster=golden-cluster --tpu-type=tpu7x-8 --spot --adapt-from-ct
+[XPK] Starting xpk v0.0.0
+[XPK] Starting cluster create for cluster golden-cluster:
+[XPK] Working on golden-project and us-central1-a
+[XPK] Task: `Find cluster region or zone` is implemented by the following command not running since it is a dry run. 
+gcloud container clusters list --project=golden-project --filter=name=golden-cluster --format="value(location)"
+[XPK] Task: `get-credentials-dns-endpoint to cluster golden-cluster` is implemented by the following command not running since it is a dry run. 
+gcloud container clusters get-credentials golden-cluster --location=us-central1 --dns-endpoint --project=golden-project && kubectl config view && kubectl config set-context --current --namespace=default
+[XPK] Task: `Test kubectl credentials` is implemented by the following command not running since it is a dry run.
```
jamOne-, line 18: Should we add a log line that the Cluster creation and Nodepool creation was skipped due to the `--adapt-from-ct` flag?

jamOne- (changes requested):

src/xpk/commands/cluster.py
SikaGrr: Initially I thought to skip as much as possible to avoid clashing with CT, but since this already has a flag we don't need to skip it.

SikaGrr: Actually we need GKE master version for this, and during cluster create we add --enable-managed-mldiagnostics flag, so fixing this for adapt_from_ct would require some work, let's do it in a followup.

jamOne-: SG to skip it.

src/xpk/commands/cluster.py
SikaGrr: Sure, but it turned out that `_validate_num_slices_and_set_defaul` (to set defaults) is needed to avoid a crash in configmaps, Is there some other side effect of validation that we might be missing?

recipes/Cluster_create_adapt_from_ct.md
SikaGrr: Done.

src/xpk/commands/cluster.py
jamOne-: Don't think so. I knew only about this one.

scaliby (approved):

jamOne- (approved):

FIoannides (approved):

